PR #380, opened 2016-12-02
Fix RTD URLS

cpascual:
Some links to URLs to sardana-controls.org are wrong (in the SEP15.md):
- they use https but RTD does not support that (and therefore, some browsers warn)
- the SEP index URL must end with a slash

@zreszela, could you review this?


reszelaz:
Agree. Thanks!

PR #382, opened 2016-12-07
Feature 360

rhomspuron:
Implement the code to save the OneD channels on spec file

### DIFF
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -308,14 +308,25 @@ class SPEC_FileRecorder(BaseFileRecorder):
         #store names for performance reason
         labels = []
         names = []
+        oned_labels = []
+        oned_names = []
+        oned_shape = 0
         for e in env['datadesc']:
             dims = len(e.shape)
+            if dims >= 2:
+                continue
+            sanitizedlabel = "".join(x for x in e.label.replace(' ', '_') if x.isalnum() or x == '_')  #substitute whitespaces by underscores and remove other non-alphanumeric characters
             if not dims or (dims == 1 and e.shape[0] == 1):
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi,
        I have tried this and I have a question about the information you put in the header.
You add there, in case of MCAs:

#@MCA 12C
#@CHANN 12 0 11 1
#@MCA_NB 2

In my case I put two MCAs in the MG and both with 12 channels, but I have checked that the code
generating those lines is:

        if len(oned_labels) > 0:
            header += '#@MCA %sC\n' % oned_shape
            header += '#@CHANN %s 0 %s 1\n' % (oned_shape, oned_shape-1)
            header += '#@MCA_NB %s\n' % len(oned_labels)
            for idx, oned_label in enumerate(oned_labels):
                header += '#@DET_%s %s\n' %(idx, oned_label)

And oned_shape comes from:

 oned_shape = e.shape[0]

where 'e' would be the last of the MCAs found in the datadesc, so what happens in case of two
MCAs with different length?, it is not forseen this case?. Perhaps I have missed something.

                             Regards,
                                               Teresa

rhomspuron:
Hi Teresa, 

This information on the channel is optional, we implemented with @vrey01 and we saw this problem for multi MCA channels. 

This information was included new on the specfile specification:
https://github.com/specpython/filespec/blob/master/specpython/spec_format.rst

We should agree a convention with him to include multi MCA with different shape and different ROIs. This is not implemented on the spec (program). 

In our case, we use the same shape for all 1D channes. At the end, none programs read this information, because it is new and optional.

teresanunez:
Thanks Roberto, then it is fine for me. At DESY we also do not  read the spectra from the SPEC
 files. I can then integrate the patch like it is and in the future it can always be changed if the information
 of the header is needed for something.

rhomspuron:
Thanks

PR #387, opened 2016-12-12
Add bumpversion support (and bump version to 2.1.2-dev)

cmft:
Add bumpversion configuration to update the version using the
bumpversion script from https://github.com/peritus/bumpversion

Deprecate the version_info and revision variables from the
sardana.release module.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 2.1.2-dev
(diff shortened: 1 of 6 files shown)


.bumpversion.cfg
```diff
@@ -0,0 +1,20 @@
...
+	{major}.{minor}.{patch}
+
+[bumpversion:file:src/sardana/release.py]
+search = version = '{current_version}'
+replace = version = '{new_version}'
+
+[bumpversion:part:release]
+optional_value = gamma
+values = 
+	dev
```
cmft: Ups! You are right!! I will fix it

PR #391, opened 2016-12-20
Add CHANGELOG in Sardana

sagiss:
Add CHANGELOG file in Sardana documenting the major changes
between versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,6 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the formats and conventions from [keepachangelog.com]
 
-Find Sardana issues in the following locations:
-[old SF sardana issues](https://sourceforge.net/p/sardana/tickets/)
-[migrated sardana issues](https://github.com/sardana-org/sardana/issues)
-
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,132 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+This file follows the formats and conventions from [keepachangelog.com]
+
+Find Sardana issues in the following locations:
+[old SF sardana issues](https://sourceforge.net/p/sardana/tickets/)
```
cpascual: I would remove the mention to the old SF. In the case of Sardana the new issues already contain links to the SF tickets

CHANGELOG.md
```diff
@@ -0,0 +1,132 @@
...
+Fix SF issues: #509, #507, #508, #511, #426
+
+
+## [2.1.0] - 2016-09-13
+For a full log of commits between versions run (in your git repo):
+`git log 2.1.0..2.0.0`
+Main improvements since sardana 2.0.0 (aka Jan16) are:
+
+### Added
+- Add itango dependency to make Sardana compatible with the latest
```
cpascual: IMHO, these entries need to be rewritten to avoid "added" or "add" in the "Added setcion. It is unnatural and redundant. Similarly, avouid "fix" or "solve"  in the case of the "Fixed" sections, etc.

cpascual: Also, some entries read a bit like commit message subjects... they should be rewritten according to the guidelines of http://keepachangelog.com/en/0.3.0/ (especially sections such as "what make unicorns cry" and "Why can’t people just use a git log diff"

reszelaz:
I agree with the Carlos' suggestions, and with the Marc's corrections. I will proceed with the merge.
Thanks to both!

Issue #403, opened 2017-01-01
spock upgrade fails with alpha releases


reszelaz:
The spock upgrade mechanism fails with `alpha` releases. After upgrading to `2.2.2-alpha`, the subsequent executions of the spock always offer an upgrade to `2.2.2-alpha`:
```
zreszela@pc255:~/workspace/taurus (develop)> spock --profile=test01
MainThread     INFO     2017-01-01 20:15:48,467 TaurusRootLogger: Using "PyQt4" for Qt
MainThread     INFO     2017-01-01 20:15:50,772 taurus.qt.qtgui.icon.icon: Setting Tango icon theme (from /home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/icon/)
Your current spock door extension profile has been created with spock 2.1.1.
Your current spock door extension version is 2.2.2-alpha, therefore a profile upgrade is needed.

Do you wish to upgrade now (warn: this will shutdown the current spock session) ([y]/n)? y
Storing ipython_config.py in /home/zreszela/.ipython/profile_test01... [DONE]

zreszela@pc255:~/workspace/taurus (develop)> spock --profile=test01
MainThread     INFO     2017-01-01 20:16:15,833 TaurusRootLogger: Using "PyQt4" for Qt
MainThread     INFO     2017-01-01 20:16:16,097 taurus.qt.qtgui.icon.icon: Setting Tango icon theme (from /home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/icon/)
Your current spock door extension profile has been created with spock 2.2.2-alpha.
Your current spock door extension version is 2.2.2-alpha, therefore a profile upgrade is needed.

Do you wish to upgrade now (warn: this will shutdown the current spock session) ([y]/n)? y
Storing ipython_config.py in /home/zreszela/.ipython/profile_test01... [DONE]

zreszela@pc255:~/workspace/taurus (develop)> spock --profile=test01
MainThread     INFO     2017-01-01 20:16:30,124 TaurusRootLogger: Using "PyQt4" for Qt
MainThread     INFO     2017-01-01 20:16:30,377 taurus.qt.qtgui.icon.icon: Setting Tango icon theme (from /home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/icon/)
Your current spock door extension profile has been created with spock 2.2.2-alpha.
Your current spock door extension version is 2.2.2-alpha, therefore a profile upgrade is needed.

[...]

PR #404, opened 2017-01-01
SEP6 - Continuous Scans

reszelaz:
[SEP6](https://github.com/reszelaz/sardana/blob/sep6/doc/source/sep/SEP6.md) is almost ready to be integrated in the develop branch. The SEP6 document was recently reviewed and updated. The documentation was extended with the changes introduced in SEP6 and is temporarily available [here](http://sardanatest.readthedocs.io/en/sep6). Especially the documentation has still a lot of room for improvement e.g. the “Springfield” controllers used in the “How to write a counter/timer controller” or “How to write a trigger/gate controller” are just skeletons and are not functional. 

In the following week, at Alba, we will try to upgrade one of our beamline to this branch and will commit changes (if necessary) and post updates on the migration and tests in this PR. We also ask the other institutes and users to review the SEP6 and evaluate this branch so soon we could vote the acceptance. What should be especially tested is the backwards compatibility of all the acquisition modes of Sardana available up to now and used in your particular scenarios.

The travis-ci will fail for the moment, cause the Docker image is not ready to perform all the available tests. Finally the CHANGELOG.md must be updated with the changes introduced by this PR.


teresanunez:
Hi Zibi,
                I have tried to use this with the current controllers at 
DESY (for counters and timers)
  and I can not run the scans. The Pool give me the warning 'LoadOne API 
has changed' and when
  I try to run a scan sometimes it crashes (complaining about the 
counters) and sometimes it
  runs but the output of spock shows only the timer but not any counter.
  Another thing, when I start spock I get always the message telling 
that an upgrade is needed, even
  after making the upgrade. Always when I start spock I get:

  Your current spock door extension profile has been created with spock 
2.2.2-alpha.
Your current spock door extension version is 2.2.2-alpha, therefore a 
profile upgrade is needed.
Do you wish to upgrade now (warn: this will shutdown the current spock 
session) ([y]/n)? n

  I have always to answer this question (with 'n' for getting spock 
running).
  I have even created a new spock profile and it happens the same.
  Well, enough for now. Thanks.

                           Regards,
                                           Teresa

teresanunez:
Hi again,
                   I see now that you force the message in spock because 
you have added
'and spock_profile_ver_str.find("-alpha") == -1' to the check of the 
version in the check_upgrade
  function. Do you want this question always to rise up for alpha versions?.

                     Regards,
                                       Teresa

teresanunez:
Hi again,
              I have seen now that the problem with the scans is due to the Nexus recorder from
 DESY, if I do not use it I get only the warnings but the scan runs. I will contact Jan for looking at
 that. Sorry. 
                            Regards,
                                         Teresa

reszelaz:
The issue with spock is present in develop as well: #403. Let's it there...

reszelaz:
We have deployed the SEP6 at one of our beamline and we have just received a positive feedback after one week of operation.

However there was one regression reported - some acquisitions in step scan, ct or ascanc causes deadlock in the MacroServer. In all of the cases the origin of the problem was an exception raised in one of the CounterTimer controllers and more precisely in one of the start sequence methods (`PreLoadAll`, `PreLoadOne`, `LoadOne`, `LoadAll`, `PreStartAll`, `PreStartOne`, `StartOne`, `StartAll`).

With SEP6 the whole start sequence have been moved to a working thread that starts  when the software synchronizer gives a signal to the start. Previously the start sequence was handled in the measurement groups's Start command, so any exception there was easily noticed by the MacroServer.

We need to protect the calls to the controllers API and modify the measurement group state accordingly (via state of the elements), so the MacroServer could realize of an abnormal situation.

reszelaz:
I’m thinking about removing the `read_when_acq` experimental channel controller property which in my opinion makes the acquisition actions much more complicated at the same time adding few benefits. The same feature could be easily achieved when programming the controller by basically verifying the controller’s state in the read methods. The state is evaluated before the first readout and the final read is ensured after the state had changed from MOVING to ON.

So in my opinion we are not losing this possibility and we make the acquisition actions less complex. This is crucial having in mind that they will require refactoring and enhancements - technical debt related with the measurement group configuration, SEP2, ...

If no negative comments are raised I will proceed with the change soon.

reszelaz:
I have already removed the `read_when_acq` from the code and SEP6.

Teresa had spotten some incosistencies with the `active_domain` and `passive_domain` configuration parameters of the TriggerGate elements. I will review then and post back soon.

reszelaz:
I start with a little bit of background. The `active_domain` and `passive_domain` are attributes of trigger/gate elements. The trigger/gate elements are configured with the synchronization description via the `Synch` methods of their controllers. Depending on the acquisition, either a continuous scan, a time scan or simply a count, this description is expressed in the position and time domains or simply in the time domain. In case of the continuous scans, when the synchronization is expressed in both domains, it is recommended to configure the hardware so that the acquisition starts i.e. trigger, or beginning of gate, based on the configuration in the position domain and stops i.e. end of gate, based on the time domain (this is just a recommendation to the controllers developers and not a mechanism implemented in the core). However we would like to give freedom to the user to alternate the default behavior. The idea behind the `active_domain` and `passive_domain` attributes is to be able to change the default behavior and force the use of a different domain. Furthermore it it is worth mentioning that not all hardware are compatible with both domains.

There are some points that I would like to discuss:
1. **Naming**. `active_domain` and `passive_domain` names originates from the `active` and `passive` events that are the core internal concepts related to the way in which the software synchronization works. Basically the `active` is the event which starts the acquisition and `passive` is the event that stops it. These names may be a little bit confusing, since in the synchronization description the corresponding fields are named: `initial` and `active`. I’m more in favor of using `initial_domain` and `active_domain` instead of `active_domain` and `passive_domain` respectively. What is your opinion?
2. **Priority**. These attributes are considered as part of the SEP6, but in my feeling they are quite advanced concepts. We could even think of excluding them from the SEP6 and add them in the future when we gain more feedback from the trigger/gate controller developers. The current implementation is not complete - the Tango attributes are still not implemented in the TriggerGate class. If we decide to exclude them from SEP6, I estimate that the effort is similar to remove them or to finish the implementation. Do you see that we could already take a profit of these attributes?

teresanunez:
As you already suggested, I would change the names active_domain and pasive_domain. I also agree with the names you suggested, I am not sure if they are clear enough, but since they are already used in the Synchronization attribute of the measurement group (and there I find them appropriate)  I think it is better to use them also here.
About the priority, well, we will need to use them for our PiLC trigger generator, in fact I was messing this and I added an extra attribute in the TriggerGateController that I have implemented for tests. But we
 can live with this extra attribute until the implementation is done. No problem.

reszelaz:
I’m not sure if these attributes should form part of the generic interface of any TriggerGate element so for the moment it is better to not include them. We can always add them in the future. So for the moment, please use them as extra axis attributes in the PiLC controller.
It would be nice if you use the names InitialDomain and ActiveDomain cause if we add them in the future I see these names as the most appropriate.

I have already changed the internal members of the software synchronizer to use the new names and also removed the SetAxisPar and GetAxisPar from the DummyTriggerGate controller.

I consider this issue as closed.

reszelaz:
The current state of the SEP6 is a complete proposal of implementation.

Please try it and comment because I would like to pass SEP6 to accepted if everybody is ok with the implementation.

**IMPORTANT**: remember that before trying it on an already existing instance of the Sardana system it is recommended to make a backup of the measurement groups’ configurations. Modifying these configurations with this version of Sardana makes them not compatible with the previous versions - rollbacks won’t work straightforward.

I also kindly ask Teresa to be the “formal” reviewer and integrator. Finally we will proceed to the formal voting on SEP6 acceptance.

teresanunez:
Hi Zibi,I have continued reviewing this implementation. There are some points, like for example the use of external channels as counters, that, I think, are important to have, but you have already listed them as 'Out of the scope' of SEP6. So, ok,  it is fine for me, they can be added later, but they should be taken into account. Another point is the error handling, it is difficult to know what is wrong if something is wrong, usually one get the error message:

 "error while moving to waypoints"

and from there one can not get much more information about what to do, but well, I think that this can also be improved during the use (and I also think that setting properly times/positions/velocities ... most of these errors can be avoided, so in real conditions it should mainly work). 
From my point of view it could be accepted. Many thanks for the great job.

reszelaz:
I completely agree that we must support the external attributes in continuous scan. But this is a bigger change. It will require to move the readout to the Pool, where the software synchronization takes place now. Apart of that we will need to think on how to transfer the data of the external attributes from Pool to the MacroServer later on. So let’s leave it for the future.

Regarding the verbosity of the error messages you are right. There is a big room for improvement. I have already changed the way the motion configuration errors are handled. I think these are the most common cases that could be confusing (calculated velocities, acceleration time out of range, etc.). So now the error messages should look more or less like this:

~~~
Error when configuring scan motion (setting velocity of mot14 to 100.0 failed)
~~~

So I would like to start the voting on the acceptance of SEP6 and merge of reszelaz/sep6 into the canonical develop branch.

teresanunez:
Hi Zibi, thanks for your comments and the changes in the error messages. It is perfectly fine for
 us to postpone the rest of the improvements. 
DESY says YES to the acceptance of the SEP6.

LJBD:
A YES from SOLARIS, too!

amilan:
Hi all, 

With a little bit of delay, but from MAX IV we also go for the YES.

 :thumbsup:

reszelaz:
Thank you all for your votes. Alba is also in favor of accepting this SEP.

So I think it means a green light to accept the PR. Before doing so, I will modify the SEP document to reflect its new state as well as modify some minor travis things (in this branch it works agains the SEP6 version of the the sardana-test docker image).

I will take care of the merge.

cpascual:
On Thursday, March 30, 2017 11:44:37 PM CEST reszelaz wrote:
> So I think it means a green light to accept the PR.

Hip hip hooraaaay!

this took LOTS of work. Congrats to all involved

-- 
+----------------------------------------------------+
 Carlos Pascual Izarra
 Scientific Software Coordinator
 Computing Division
 ALBA Synchrotron  [http://www.albasynchrotron.es]
 Carrer de la Llum 2-26
 E-08290 Cerdanyola del Valles (Barcelona), Spain
 E-mail: cpascual@cells.es
 Phone: +34 93 592 4428
+----------------------------------------------------+

droldan:
Great! :+1:

PR #405, opened 2017-01-02
New ParamRepeat interface in spock

teresanunez:
Hello,
               I have done a first implementation for extending the new ParamRepeat interface to spock.
The old spock input syntax is , of course, still valid and, as before, only admit ParamRepeat parameters in the last position. With the new interface the ParamRepeat parameters can be used at any position and even recursively. I have choosen the syntax using square brackets as group characters and space as separator e.g.

mv [[mot01 5] [mot02 10]]
pt10 [0 5 10] mot01
pt11 ct01 [10 5 0] mot01
pt12 [0 5 10] [mot01 mot02 mot03]
pt13 [[mot01 mot02 mot03] [mot04 mot05]]

Everything is, of course, open to discussion, this is only a first attempt. 

                                      Regards,
                                                        Teresa

### DIFF
--- a/doc/source/devel/howto_macros/macros_general.rst
+++ b/doc/source/devel/howto_macros/macros_general.rst
@@ -333,8 +333,8 @@ execution time.
 The repeat parameter definition allows to:
 
     - restrict the minimum and/or maximum number of repetitions
-    - nest repeat parameters inside of another repeat parameters [#f3]_
-    - define multiple repeat parameters in the same macro [#f3]_
+    - nest repeat parameters inside of another repeat parameters
+    - define multiple repeat parameters in the same macro
 
 Repeat parameter values are passed to the macro function in the form of a list.
 If the repeat parameter definition contains just one member it is a plain list
@@ -596,12 +596,16 @@ parameters with different *flavors*:
         self.execMacro('mv', [[motor.getName(), '0']])
         self.execMacro('mv', motor.getName(), '0') # backwards compatibility - see note
 
-    * parameters as space separated string (this is not compatible with multiple
-      or nested repeat parameters, furthermore the repeat parameter must be the last one)::
+    * parameters as space separated string::
 
         self.execMacro('ascan %s 0 100 10 0.2' % motor.getName())
-        self.execMacro('mv %s 0' % motor.getName())
-
+        self.execMacro('mv [[%s 0]]' % motor.getName()) 
+        self.execMacro('mv [%s 0]' % motor.getName()) # simplified interface
+        self.execMacro('mv %s 0' % motor.getName()) # backwards compatibility - see note
+        self.execMacro('mv [[%s 0][%s 20]]' % (motor.getName(), motor2.getName()))
+        self.execMacro('mv [%s 0 %s 20]' % (motor.getName(), motor2.getName())) # simplified interface
+        self.execMacro('mv %s 0 %s 20' % (motor.getName(), motor2.getName())) # backwards compatibility - see note
+      
     * parameters as concrete types::
 
         self.execMacro(['ascan', motor, 0, 100, 10, 0.2])
@@ -613,7 +617,9 @@ parameters with different *flavors*:
     use :ref:`repeat parameters <sardana-macro-repeat-parameters>`.
     From Sardana 2.0 the repeat parameter values must be passed as lists of
     items. An item of a repeat parameter containing more than one member is a
-    list. In case when a macro defines only one repeat parameter
+    list. In order to simplify the interface the use of the brackets closing each
+    repetition of a ParamRepeat can be skipped.
+    In case when a macro defines only one repeat parameter
     and it is the last parameter, for the backwards compatibility reasons, the
     plain list of items' members is allowed.
 
@@ -1249,9 +1255,6 @@ value but also the ranges so it is safe to change them if necessary.
 
 .. [#f2] To check which version of Python_ you are using type on the command
          line ``python -c "import sys; sys.stdout.write(sys.version)"``
-
-.. [#f3] Currently only GUI macro clients (*macroexecutor* or *sequencer*)
-       are compatible with this kind of repeat parameters usage.
 
 .. |input_integer| image:: ../../_static/macro_input_integer.png
     :align: middle
--- a/doc/source/devel/howto_macros/macros_general.rst
+++ b/doc/source/devel/howto_macros/macros_general.rst
@@ -600,10 +600,8 @@ parameters with different *flavors*:
 
         self.execMacro('ascan %s 0 100 10 0.2' % motor.getName())
         self.execMacro('mv [[%s 0]]' % motor.getName()) 
-        self.execMacro('mv [%s 0]' % motor.getName()) # simplified interface
         self.execMacro('mv %s 0' % motor.getName()) # backwards compatibility - see note
         self.execMacro('mv [[%s 0][%s 20]]' % (motor.getName(), motor2.getName()))
-        self.execMacro('mv [%s 0 %s 20]' % (motor.getName(), motor2.getName())) # simplified interface
         self.execMacro('mv %s 0 %s 20' % (motor.getName(), motor2.getName())) # backwards compatibility - see note
       
     * parameters as concrete types::
@@ -617,8 +615,7 @@ parameters with different *flavors*:
     use :ref:`repeat parameters <sardana-macro-repeat-parameters>`.
     From Sardana 2.0 the repeat parameter values must be passed as lists of
     items. An item of a repeat parameter containing more than one member is a
-    list. In order to simplify the interface the use of the brackets closing each
-    repetition of a ParamRepeat can be skipped.
+    list.
     In case when a macro defines only one repeat parameter
     and it is the last parameter, for the backwards compatibility reasons, the
     plain list of items' members is allowed.
--- a/src/sardana/macroserver/macros/examples/parameters.py
+++ b/src/sardana/macroserver/macros/examples/parameters.py
@@ -154,7 +154,7 @@ class pt6(Macro):
 class pt7(Macro):
     """Macro with a list of pair Motor,Float.
     Usages from Spock, ex.:
-    pt7 [[mot1 1][mot2 3]]
+    pt7 [[mot1 1] [mot2 3]]
     pt7 mot1 1 mot2 3
     """
 
@@ -170,10 +170,10 @@ class pt7(Macro):
 class pt7d1(Macro):
     """Macro with a list of pair Motor,Float. Default value for last ParamRepeat element.
     Usages from Spock, ex.:
-    pt7d1 [[mot1 1][mot2 3]]
+    pt7d1 [[mot1 1] [mot2 3]]
     pt7d1 mot1 1 mot2 3
     Using default value, ex.:
-    pt7d1 [[mot1][mot2 3]] # at any repetition
+    pt7d1 [[mot1] [mot2 3]] # at any repetition
     
     """
 
@@ -190,10 +190,10 @@ class pt7d1(Macro):
 class pt7d2(Macro):
     """Macro with a list of pair Motor,Float. Default value for both ParamRepeat elements.
     Usages from Spock, ex.:
-    pt7d2 [[mot1 1][mot2 3]]
+    pt7d2 [[mot1 1] [mot2 3]]
     pt7d2 mot1 1 mot2 3
     Using both default values, ex.:
-    pt7d2 [[][mot2 3][]] # at any repetition
+    pt7d2 [[] [mot2 3] []] # at any repetition
     """
 
     param_def = [
@@ -209,7 +209,7 @@ class pt8(Macro):
     """Macro with a list of pair Motor,Float. The min and max elements have been
     explicitly stated.
     Usages from Spock, ex.:
-    pt8 [[mot1 1][mot2 3]]
+    pt8 [[mot1 1] [mot2 3]]
     pt8 mot1 1 mot2 3    
     """
     
@@ -262,7 +262,6 @@ class pt11(Macro):
     """Macro with list of numbers followed by a motor parameter. The repeat
     parameter may be defined as first one.
     Usages from Spock, ex.:
-    pt11 ct1 [[1][3]] mot1
     pt11 ct1 [1 3] mot1
     """
 
@@ -296,7 +295,7 @@ class pt13(Macro):
     """Macro with list of motors groups, where each motor group is a list of 
     motors. Repeat parameters may be defined as nested.
     Usage from Spock, ex.:
-    pt13 [[mot1 mot2][mot3 mot4]]
+    pt13 [[mot1 mot2] [mot3 mot4]]
 """
 
     param_def = [
@@ -313,7 +312,7 @@ class pt14(Macro):
     """Macro with list of motors groups, where each motor group is a list of 
     motors and a float. Repeat parameters may be defined as nested.
     Usage from Spock, ex.:
-    pt14 [[[mot61 mot62] 3] [[mot63] 5]]
+    pt14 [[[mot1 mot2] 3] [[mot3] 5]]
     """
 
     param_def = [
@@ -331,14 +330,14 @@ class pt14d(Macro):
     motors and a float. Repeat parameters may be defined as nested.
     Default values can be used.
     Usages taken default values, ex.:
-    pt14d [[[mot61 mot62] 3] [[mot63] []]]
-    pt14d [[[mot61 []] 3] [[mot63] []]]
-    pt14d [[[[]] 3] [[mot03] []]]
+    pt14d [[[mot1 mot2] 3] [[mot3] []]]
+    pt14d [[[mot1 []] 3] [[mot3] []]]
+    pt14d [[[[]] 3] [[mot3] []]]
     """
 
     param_def = [
        ['motor_group_list',
-        [['motor_list', [['motor', Type.Motor, 'exp_dmy01', 'Motor to move']], None, 'List of motors'],
+        [['motor_list', [['motor', Type.Motor, 'mot1', 'Motor to move']], None, 'List of motors'],
          ['float', Type.Float, 33, 'Number']],
         None, 'Motor groups']
     ]
--- a/src/sardana/macroserver/macros/examples/parameters.py
+++ b/src/sardana/macroserver/macros/examples/parameters.py
@@ -31,7 +31,7 @@ __all__ = ["pt0", "pt1", "pt2", "pt3", "pt3d", "pt4", "pt5", "pt6", "pt7",
 
 class pt0(Macro):
     """Macro without parameters. Pretty dull.
-       Usage, ex.:
+       Usage from Spock, ex.:
        pt0
        """
 
@@ -44,7 +44,7 @@ class pt1(Macro):
     """Macro with one float parameter: Each parameter is described in the 
     param_def sequence as being a sequence of four elements: name, type, 
     default value and description.
-    Usage, ex.:
+    Usage from Spock, ex.:
     pt1 1
     """
     
@@ -57,7 +57,7 @@ class pt2(Macro):
     """Macro with one Motor parameter: Each parameter is described in the 
     param_def sequence as being a sequence of four elements: name, type, 
     default value and description.
-    Usage, ex.
+    Usage from Spock, ex.
     pt2 mot1
     """
 
@@ -71,8 +71,7 @@ class pt3(Macro):
     parameter types which is repeated. In this case it is a repetition of a 
     float so only one parameter is defined.
     By default the repetition as a semantics of 'at least one'
-    Usages, ex.:
-    pt3 [[1][34][15]]
+    Usages from Spock, ex.:
     pt3 [1 34 15]
     pt3 1 34 15
     """
@@ -89,12 +88,10 @@ class pt3d(Macro):
     parameter types which is repeated. In this case it is a repetition of a 
     float so only one parameter is defined. The parameter has a default value.
     By default the repetition as a semantics of 'at least one'
-    Usages, ex.:
-    pt3d [[1][34][15]]
+    Usages from Spock, ex.:
     pt3d [1 34 15]
     pt3d 1 34 15
-    Usages taken the default value, ex.:
-    pt3d [[1] [] [15]]
+    Usage taken the default value, ex.:
     pt3d [1 [] 15]
     """
 
@@ -110,10 +107,9 @@ class pt4(Macro):
     parameter types which is repeated. In this case it is a repetition of a 
     motor so only one parameter is defined.
     By default the repetition as a semantics of 'at least one'.
-    Usages, ex.:
-    pt3 [[mot1][mot2][mot3]]
-    pt3 [mot1 mot2 mot3]
-    pt3 mot1 mot2 mot3
+    Usages from Spock, ex.:
+    pt4 [mot1 mot2 mot3]
+    pt4 mot1 mot2 mot3
     """
 
     param_def = [
@@ -125,8 +121,7 @@ class pt4(Macro):
 
 class pt5(Macro):
     """Macro with a motor parameter followed by a list of numbers.
-    Usages, ex.:
-    pt5 mot1 [[1][3]]
+    Usages from Spock, ex.:
     pt5 mot1 [1 3]
     pt5 mot1 1 3
     """
@@ -143,8 +138,7 @@ class pt6(Macro):
     """Macro with a motor parameter followed by a list of numbers. The list as
     explicitly stated an optional last element which is a dictionary that defines the
     min and max values for repetitions.
-    Usages, ex.:
-    pt6 mot1 [[1][34][1]]
+    Usages from Spock, ex.:
     pt6 mot1 [1 34 1]
     pt6 mot1 1 34 1
     """
@@ -159,9 +153,8 @@ class pt6(Macro):
     
 class pt7(Macro):
     """Macro with a list of pair Motor,Float.
-    Usages, ex.:
+    Usages from Spock, ex.:
     pt7 [[mot1 1][mot2 3]]
-    pt7 [mot1 1 mot2 3]
     pt7 mot1 1 mot2 3
     """
 
@@ -176,13 +169,11 @@ class pt7(Macro):
     
 class pt7d1(Macro):
     """Macro with a list of pair Motor,Float. Default value for last ParamRepeat element.
-    Usages, ex.:
+    Usages from Spock, ex.:
     pt7d1 [[mot1 1][mot2 3]]
-    pt7d1 [mot1 1 mot2 3]
     pt7d1 mot1 1 mot2 3
     Using default value, ex.:
     pt7d1 [[mot1][mot2 3]] # at any repetition
-    pt7d1 [mot1 1 [mot2]] # only at the last repetition
     
     """
 
@@ -198,13 +189,11 @@ class pt7d1(Macro):
     
 class pt7d2(Macro):
     """Macro with a list of pair Motor,Float. Default value for both ParamRepeat elements.
-    Usages, ex.:
+    Usages from Spock, ex.:
     pt7d2 [[mot1 1][mot2 3]]
-    pt7d2 [mot1 1 mot2 3]
     pt7d2 mot1 1 mot2 3
     Using both default values, ex.:
-    pt7d1 [[][mot2 3][]] # at any repetition
-    pt7d1 [mot1 1 []] # only at the last repetition
+    pt7d2 [[][mot2 3][]] # at any repetition
     """
 
     param_def = [
@@ -219,9 +208,8 @@ class pt7d2(Macro):
 class pt8(Macro):
     """Macro with a list of pair Motor,Float. The min and max elements have been
     explicitly stated.
-    Usages, ex.:
+    Usages from Spock, ex.:
     pt8 [[mot1 1][mot2 3]]
-    pt8 [mot1 1 mot2 3]
     pt8 mot1 1 mot2 3    
     """
     
@@ -239,9 +227,8 @@ class pt9(Macro):
     """Same as macro pt7 but with old style ParamRepeat. If you are writing
     a macro with variable number of parameters for the first time don't even
     bother to look at this example since it is DEPRECATED.
-    Usages, ex.:
+    Usages from Spock, ex.:
     pt9 [[mot1 1][mot2 3]]
-    pt9 [mot1 1 mot2 3]
     pt9 mot1 1 mot2 3
     """
 
@@ -258,8 +245,7 @@ class pt9(Macro):
 class pt10(Macro):
     """Macro with list of numbers followed by a motor parameter. The repeat
     parameter may be defined as first one.
-    Usages, ex.:
-    pt10 [[1][3]] mot1
+    Usage from Spock, ex.:
     pt10 [1 3] mot1
     """
 
@@ -275,7 +261,7 @@ class pt10(Macro):
 class pt11(Macro):
     """Macro with list of numbers followed by a motor parameter. The repeat
     parameter may be defined as first one.
-    Usages, ex.:
+    Usages from Spock, ex.:
     pt11 ct1 [[1][3]] mot1
     pt11 ct1 [1 3] mot1
     """
@@ -293,8 +279,7 @@ class pt11(Macro):
 class pt12(Macro):
     """Macro with list of motors followed by list of numbers. Two repeat
     parameters may defined.
-    Usages, ex.:
-    pt12 [[1][3][4]] [[mot1][mot2]]
+    Usage from Spock, ex.:
     pt12 [1 3 4] [mot1 mot2]
     """
 
@@ -310,8 +295,7 @@ class pt12(Macro):
 class pt13(Macro):
     """Macro with list of motors groups, where each motor group is a list of 
     motors. Repeat parameters may be defined as nested.
-    Usages, ex.:
-    pt13 [[[[mot1] [mot2]]][[[mot3] [mot4]]]]
+    Usage from Spock, ex.:
     pt13 [[mot1 mot2][mot3 mot4]]
 """
 
@@ -328,9 +312,8 @@ class pt13(Macro):
 class pt14(Macro):
     """Macro with list of motors groups, where each motor group is a list of 
     motors and a float. Repeat parameters may be defined as nested.
-    Usages, ex.:
-    pt14 [[[[mot61] [mot62]] 3]  [[[mot63]] 5] ]
-    pt14 [[mot61 mot62] 3  [mot63] 5]
+    Usage from Spock, ex.:
+    pt14 [[[mot61 mot62] 3] [[mot63] 5]]
     """
 
     param_def = [
@@ -347,15 +330,10 @@ class pt14d(Macro):
     """Macro with list of motors groups, where each motor group is a list of 
     motors and a float. Repeat parameters may be defined as nested.
     Default values can be used.
-    Usages, ex.:
-    pt14d [[[[mot61] [mot62]] 3]  [[[mot63]] 5] ]
-    pt14d [[mot61 mot62] 3  [mot63] 5]
     Usages taken default values, ex.:
-    pt14d [[[[mot61] [mot62]]]  [[[mot63]] 5] ]
-    pt14d [[[[mot61] []]]  [[[mot63]] 5] ]
-    pt14d [[[]]  [[[mot63]] 5] ]
-    pt14d [[mot61 mot62] 3  [] 5]
-    pt14d [[mot61 mot62] 3  []] # only at the last repetition
+    pt14d [[[mot61 mot62] 3] [[mot63] []]]
+    pt14d [[[mot61 []] 3] [[mot63] []]]
+    pt14d [[[[]] 3] [[mot03] []]]
     """
 
     param_def = [
--- a/src/sardana/macroserver/macros/examples/parameters.py
+++ b/src/sardana/macroserver/macros/examples/parameters.py
@@ -25,11 +25,15 @@
 
 from sardana.macroserver.macro import *
 
-__all__ = ["pt0", "pt1", "pt2", "pt3", "pt4", "pt5", "pt6", "pt7", "pt8",
-           "pt9"]
+__all__ = ["pt0", "pt1", "pt2", "pt3", "pt3d", "pt4", "pt5", "pt6", "pt7",
+           "pt7d1", "pt7d2", "pt8", "pt9", "pt10", "pt11", "pt12", "pt13",
+           "pt14", "pt14d", "twice"]
 
 class pt0(Macro):
-    """Macro without parameters. Pretty dull"""
+    """Macro without parameters. Pretty dull.
+       Usage, ex.:
+       pt0
+       """
 
     param_def = []
     
@@ -39,7 +43,10 @@ class pt0(Macro):
 class pt1(Macro):
     """Macro with one float parameter: Each parameter is described in the 
     param_def sequence as being a sequence of four elements: name, type, 
-    default value and description"""
+    default value and description.
+    Usage, ex.:
+    pt1 1
+    """
     
     param_def = [ [ 'value', Type.Float, None, 'some bloody float'] ]
     
@@ -49,7 +56,10 @@ class pt1(Macro):
 class pt2(Macro):
     """Macro with one Motor parameter: Each parameter is described in the 
     param_def sequence as being a sequence of four elements: name, type, 
-    default value and description"""
+    default value and description.
+    Usage, ex.
+    pt2 mot1
+    """
 
     param_def = [ [ 'motor', Type.Motor, None, 'some bloody motor'] ]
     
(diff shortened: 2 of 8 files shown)


teresanunez:
Hi, I am making some modification in the code. I found a problem in a concrete case.

teresanunez:
Hi, I have modified the interface following some discussions with Zibi and Marc. Now the elements of each ParamRepeat have to be closed in []. This version is again only a starting point for the discussion
about the final implementation. Thanks for testing and giving your comments.

teresanunez:
This implementation also extends the possible execMacro syntax. Now it is also possible to use, for
example:
 self.execMacro("mv [[exp_dmy01 10] [exp_dmy02 10]]")
using the same spock syntax for the new ParamRepeat implementation. Or for example, for a 
case not supported for the old interfece, one could use something like:

 self.execMacro("paramrepeat_multiple_test [[exp_dmy01 1] [exp_dmy01 2]] 3")

teresanunez:
Some information related to the case in which ParamRepeat is defined containing only one element, I mean, something like:
param_def = [
        ['value_list',
         ParamRepeat(['value_par', Type.Float, 33 , 'Test value']),
         None, 'List of values]]

In this case one can call the macro closing each single parameter into []:
paramrepeat_only_onepar [[1][2][3]]

like for the other cases, or one can use simply:
paramrepeat_only_onepar [1 2 3]

 If, like in this example, the single parameter has a default value and this wants to
 be used, then one have to use give each single value closed into brackets and use
 [] for taking the default value, that means, ex.

 paramrepeat_only_onepar [[1][][3]]

 where the second repetition of the ParamRepeat will take the default value, 33 in the example.

teresanunez:
This PR would solve the issue #321

sagiss:
The review of this pull request is done and it works in the case that we use brackets in all cases. 
The macros with param repeats can be launched from spock using brackets for all RepeatNodes and all RepeatParamNodes.

Now it is possible to do:
paramrepeat_only_onepar [[1][2][3]]

-----

Conclusions:
- It is mandatory to use brackets in param repeats containing more than two elements for each repetition.
E.g:   _mv [[mot61 10] [mot62 20]]_

- To be decided if brackets are done optional or not in the case of a param repeat of a single element.

Option 1: Make mandatory to use brackets even in case of param repeats of single element for each repetition: 
paramrepeat_only_onepar [[1] [2] [3]]

Option2: Make optional the use of brackets in case of param repeats of single elements for each repetition:
paramrepeat_only_onepar [1 2 3]


For both options, the need of using default values would be indicated by empty brackets []:


--

Option1 works already in this pull request.

It will be checked if the implementation of Option2 is possible and does not introduces ambiguities.

We would like to know the opinion of the rest of the Sardana community members. Would you prefer Option1 or Option2 to be implemented.

cpascual:
Hi, @teresanunez said:
> The old spock input syntax is , of course, still valid and, as before, only admit ParamRepeat parameters in the last position. 

but now @sagiss says:
> It is mandatory to use brackets in param repeats containing more than two elements for each repetition

These to asserts seem contradictory. Is it the case that the fist one no longer applies?
If so, IMHO we should consider *very* carefully before accepting this PR (it may even imply a major version change)

teresanunez:
Hi Carlos,
                 if you don't use any brackets the old syntax is applied and it is still valid, as the sentence that  I wrote. If you use brackets, then the new syntax will be considered and in that case you have to use brackets enclosing everything. So both sentences are true: the first apply for the case you don't use brackets at all and the second one for the case you use any brackets.

cpascual:
Thanks @teresanunez for the clarification. It is ok for me (but note I haven't tested or reviewed it)

teresanunez:
Hi, I have implemented the option 2:  

"Make optional the use of brackets in case of param repeats of single elements for each repetition.
Allowing usage of:
paramrepeat_only_onepar [[1] [2] [3]]
and also allowing usage of:
paramrepeat_only_onepar [1 2 3]"

I have been testing it and it seems that it works properly for getting which interface has to be use, but please, Marc, test it again. In case none of the interfaces is valid, one will get an error, but this is what
 also happens just now allowing only one of the interfaces. Well, please, check and make your comments, thanks.

sagiss:
Hi Teresa,

I have been performing tests. It looks promising. 

One thing found is that brackets to indicate default values for single elements when using the syntax with fewer brackets is not allowed.

For example by modifying the macro pt3 indicating a default value in the module:
macros/examples/parameters.py

This does not work and I think it should.
pt3 [10 [] 4]
Error: could not convert string to float: []


Maybe pull request [416](https://github.com/sardana-org/sardana/pull/416) solves it?

sagiss:
Another minor finding also using pt3 is that spock accepts in some cases more brackets than required: 

pt3 should accept this three forms:
pt3 10 5 6
pt3 [10 5 6]
pt3 [[10] [5] [6]]

But it also accepts incorrectly this form (and only reading the 10):
`pt3 [[10 5 6]]`
With 'debug on' activated in spock and executing the line above, we get:
`[START] runMacro Macro 'pt3(['10.0'])`

Thus, only the first values is taken into account. Maybe this could be corrected in an independent issue to this pull request.

teresanunez:
Hi Marc, many thanks for looking about that. About the errors with pt3:
1. the syntax pt3 [10 [] 4] is mixing both interfaces. Up to now this is not possible, spock selects one
 or the other one, but not a mixing of both. The PR 456 does not solve this. I can check if it is possible
 to mix both.
2. the behavoir with pt3 [[10 5 6]] is also normal, the extra parameters are simply ingnored, and this
 syntax is actually  pt3 [[10]] with some extra parameters. This is also like spock works now in a normal
 case, if one put more parameters than the defined ones they are simply ignored. Should we change this
 and through out an error ?.

sagiss:
For point 1 it would be nice to check if it is possible to mix both. Because otherwise default values for individual elements cannot be indicated in the case of using the syntax with fewer square brackets.

For point 2 you have reason; I think it is Ok like it is.

Many thanks, it is a great work

teresanunez:
Ok, many thanks, I look at the point 1.

teresanunez:
Hi Marc, the syntax:

pt3 [10 [ ] 4]

for taking default values should work now. Could you please test it?, many thanks.

sagiss:
It works for the default value
pt3 [10 [ ] 4]

--

But there is still an aspect:

In this case, when I mix brackets and non brackets:
`pt3 [10 [7] [5]]`
The debug output in spock is:
`[START] runMacro Macro 'pt3(['10.0', '4.0', '4.0']) `
Thus, even if values are indicated between brackets, the default values is taken instead of the indicated value.

However when I use brackets everywhere, then it works well as expected without taking the default values:
`pt3 [[10] [7] [5]]`
`[START] runMacro Macro 'pt3(['10.0', '7.0', '5.0'])`

teresanunez:
Yes, sorry, you are right. I put the mix only for the default values ... sorry. I correct it just now.

teresanunez:
Done.

sagiss:
Yes! It works.

From my point of view it is a great work.
The review is done and I think it is ready for integration.

teresanunez:
Marc, many thanks for the review and the comments.
Carlos, could you integrate it?/

cpascual:
>Carlos, could you integrate it?

This week I am quite busy... maybe @amilan  can have a look at it?
If not, I can try next week (sorry!)

teresanunez:
Hi Carlos, no problem, next week or even later is completely fine.
Just whenever you have time. Many thanks.

cpascual:
Note: @sagiss already tested the code thoroughly, but in addition to that it would be good to at least do a quick code review before integration (that's why I do not simply click "on merge")

teresanunez:
Some update about the current implementation, the optional use of [ ] for closing each repetition of a ParamRepeat does not only apply to ParamRepeats with only one element but to all of them.
In case of ParamRepeats with more than one element [] can be also use for taking default values if all
 the elements of the ParamRepeat have a default value defined.
For example, a macro with arguments defined like:

   param_def = [
       [ 'm_p_pair', [ [ 'motor', Type.Motor, 'mot1', 'Motor to move'],
                       [ 'pos',   Type.Float, 3, 'Position to move to'] ],
         None, 'List of motor/position pairs'],
       [ 'm_value', Type.Float, None, 'extra value'],
    ]

mymacro [[mot1 1][mot2 20]] 5

mymacro[mot1 1 mot2 20] 5

mymacro [[] mot2 20] 5

I think we should add in the examples of the file parameters.py the ways in which each macro can be called and perhaps add some more examples for situations that are still not covered there.
I don't know if in the documentation of sardana would also be a place for that.

cpascual:
> I think we should add in the examples of the file parameters.py the ways in which each macro can be called and perhaps add some more examples for situations that are still not covered there.

+1

> I don't know if in the documentation of sardana there would also be a place for that.

I would say that http://sardana-controls.org/en/latest/devel/howto_macros/macros_general.html#repeat-parameters is a reasonable place for this

teresanunez:
I will first add a modified parameters.py file to the PR.
Then I will also modify  the corresponding documentation file.

teresanunez:
Hi, I have  changed the documentation, not adding anything extra but adapting the existing one.
There was not any explicit explanation about the possible uses of the new interface, this  was
 mainly documented in the example files, so I keep it like this. I have extended the examples and
 added the use of any macro in all possible interfaces. If you find that this is not clear enough an explicit explanation could be added to the documentation. Just give your comments. Thanks.

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1166,13 +1166,84 @@ def ParamFactory(paramInfo):
...
+    if hasattr(data, "__iter__"):
+        return [recur_map(fun, elem, keep_none) for elem in data]
+    else:
+        if keep_none is True and data is None:
+            return data
+        else:
+            return fun(data)
+
+                
 def createMacroNode(macro_name, params_def, macro_params):
```
reszelaz: I think that we need to be more "pythonic" in the API of this method and more precisely of the `macro_params` argument.
Now it may receives a sequence of string like this: `['[[mot01', '0]', '[mot02', '0]]']`. The internal lists should be already converted.

reszelaz: I think that this conversion could be advanced up to the [magic command](https://github.com/sardana-org/sardana/blob/develop/src/sardana/spock/spockms.py#L585). This way we could split the input string only once. Now it is split by `genutils.arg_split` and later on joined again to string for the further evaluation.

Just as a curiosity I have looked into the Python Cookbook 3 (chapter 2.19) and inspired on the example found there I have tried to implement a [parser](https://github.com/reszelaz/parser/blob/master/src/parser.py) for what we need. It does not mean that we need to use it. It is just for the reference on how to implement parsing using the regular expression.

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1166,13 +1166,84 @@ def ParamFactory(paramInfo):
         param = SingleParamNode(param=paramInfo)
     return param
 
+def check_member_node_simple(rest_raw, param_node, mem, rep, params_info_len):
+    for member_raw in rest_raw:
```
reszelaz: Docstring is missing, please explain the purpose of the function and its API.

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1166,13 +1166,84 @@ def ParamFactory(paramInfo):
...
+                if len(member_raw) == 1:
+                    member_raw = member_raw[0]
+                    member_node.setValue(str(member_raw))
+        mem += 1
+        mem %= params_info_len
+        if mem == 0:
+            rep += 1
+  
+
+def check_interface(param_raw, param_node):
```
reszelaz: Docstring is missing, please explain the purpose of the function and its API.

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1166,13 +1166,84 @@ def ParamFactory(paramInfo):
...
+            return 0
+        for k, par in enumerate(param):
+            member_node = repeat_node.child(k)
+            if isinstance(member_node, RepeatParamNode):
+                ret = check_interface(par, member_node)
+                if ret ==0:
+                    return ret
+    return 1
+            
+def check_member_node(param_raw, param_node):
```
reszelaz: Docstring is missing, please explain the purpose of the function and its API.

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1191,44 +1262,109 @@ def createMacroNode(macro_name, params_def, macro_params):
...
+        if type(macro_params) == list:
+            new_params = []
+            new_params.append(macro_params)
+            macro_params = new_params
+        # Convert tuple to list
+        macro_params = list(macro_params)
+        macro_params = recur_map(str, macro_params)
+
+    
+    if not new_interface:
```
reszelaz: I wonder if we could move the filling of parameter values to the MacroNode class. Now it contains [`toList`](https://github.com/sardana-org/sardana/blob/develop/src/sardana/taurus/core/tango/sardana/macro.py#L1087) method. What do you think to add one called `fromList`. Since we need to support both, the new interface and the old one (with the plain list of parameters), if the new one fails, it should fallback to the old one.

Another option would be to pass the json encoded list of parameters directly to the RunMacro method. But this would certainly require some adjustments in the MacroServer. This is just an idea to explore.

teresanunez:
Hi Zibi, many thanks for the review and your comments. I will have a look at them carefully.
 Only as first answers, yes, the case of white spaces is already covered, it works with them.
The check functions are exactly for checking if the new or the old interface has to be used.
The implementation already covers both, if one fails look for the other (as you say in one of 
your comments that it has to be done). Thanks again.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Docstring is added.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Docstring is added

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Docstring is added.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Hi Zibi, I am now checking the implementation and use of a fromList function in the MacroNode.

teresanunez: Hi Zibi, in the lines above we are not working with the macro_node but only with the parameters, so it has not sense for me to implement something in the macro_node class only for covering that conversion (since it has still nothing to do with the macro_node). What do you
meant with that?. Thanks.

reszelaz: All the classes `MacroNode`, `RepeatParamNode`, `RepeatNode` and `SingleParamNode` have methods `toList`. This way, from the `MacroNode` object it is possible to generate the list representation of all its parameters (recursively). I was thinking on implementing a reverse mechanism. By passing the complete list representation of parameter values, fill in the `MacroNode` object (also recursively). What do you think, is it feasible?

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Hi Zibi, the problem of changing this conversion to the magic command is that I loose the way I use for distinguishing between old or new interface. I have to get another one, but all the possibilities we mixed for allowing brackets or not brackets in many cases makes difficult to get
a sure one. The one we have now is robust, I think. I would keep this like it is.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Hi Zibi, I have moved to the MacroNode the filling of the parameters (with a fromList method, as you suggested). Actually I have only done the move ... perhaps you also wanted some changes. Should I put this new version in the PR branch, or some where else?.

src/sardana/taurus/core/tango/sardana/macro.py
reszelaz: And what about inspecting the list of parameters - the result of conversion/parsing eventually performed in the magic command. If internally it contains a list, it is the new interface. If not, it is the old one. I think it is something similar to what you already do.

teresanunez: Hi ZIbi, I tried exactly that, but it failed for some case, I will check again if it is possible to
 catch also that case.

src/sardana/taurus/core/tango/sardana/macro.py
reszelaz: Thanks! And do you think that later on it will be possible to redistribute the logic to other classes: `RepeatParamNode`, `RepeatNode`, etc. and its `fromList` methods? If you think that it is possible I can work on it.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: I checked it again, that with list doesn't work, because we don't have a list. One can make it like it was, searching [ and ]. The problem is that passing this to the magic command doesn't work like
 it is, it is not simply moving it, one would have to study again every simple case.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Just like it is I think that it has not sense. It would not simplify anything, well it
 could not be actually used.

src/sardana/taurus/core/tango/sardana/macro.py
teresanunez: Zibi, another reason, even if I convert to list in the magic command, then each element of the
 list is converted again to string when going to macro.py, so I have to make again the conversion
 of each element to a list ... so we don't gain anything, on the contrary, we have to make the same twice.

reszelaz:
@teresanunez said:
>  the optional use of [ ] for closing each repetition of a ParamRepeat does not only apply to ParamRepeats with only one element but to all of them.

I think that this makes the new syntax too open. Let's consider the example of the mv macro. Now it is possible to run it with the following command: `mv [mot01 0]`.
What may make the user think that the following command: `mv [mot01 0] [mot02 0]` will also work. But it will simply ignore the second and the subsequent repetitions.

Another problem may arise when executing a macro with the following parameters:
~~~
    param_def = [
       [ 'numbers_pair', [ [ 'number1', Type.Float, None, 'First number'],
                           [ 'number2',   Type.Float, 2, 'Second number'] ],
         None, 'List of numbers pairs']
    ]
~~~
Now the user can execute this macro in the following way: `mymacro [11 12 13]` with the following three intentions:
1. three pairs of numbers: [11, 2], [12,2] and [13,2]
2. two pairs of numbers: [11, 2] and [12, 13]
3. two pairs of numbers: [11, 12] and [13, 2]

Sardana will finally execute the third option.

So the new syntax introduces ambiguities. There can be a mismatch between the intention of the user and the sardana behavior. This ambiguities already exist in the _old syntax_ that is good but for simple cases. I think we should not introduce new syntax containing ambiguities if they can be avoided.

What do you think if we remove this support?

teresanunez:
Hi Zibi,
            for me it is not a problem, I added because of your requests, but at DESY nobody requested
that, so no problem to take it out. The ambiguity is not there if one understands which brackets can be
 suppressed and which not, but I agree that it is difficult to make the users understand the difference between the different brackets, so, no problem, take it out. It makes the implementation also much more easy.

teresanunez:
Hi Zibi, in my repo fork there is already a version without allowing the suppression of brackets. One
 could go back to it and add the last changes not related to that.

reszelaz:
Ok, we will proceed with the implementation and propose changes soon.

sagiss:
Hi Teresa,

We are reviewing the code, to simplify the interface.
After a discussion with Zibi, we think that it would be easier for users to don't allow the usage of brackets for elements of param repeats of a single element that is repeating. Allowing only the option:
[1, 2, 3]. And not allowing the option: [[1][2][3]].

What do you think about it?
If it is Ok for all of you to allow only the not usage of brackets for single elements inside a param repeat?

In this case we would proceed with the changes. 

We are also developing a battery of tests in order to test many options of input parameters; containing param repeats, etc.

teresanunez:
Hi Marc, it is ok for me. Will you make the changes?. As I wrote before, the first version in
 my fork branch is already without allowing the suppression of the brackets, you can use it
 and add the case for the ParamRepeat with single elements.

doc/source/devel/howto_macros/macros_general.rst
```diff
@@ -596,12 +596,16 @@ parameters with different *flavors*:
...
 
-    * parameters as space separated string (this is not compatible with multiple
-      or nested repeat parameters, furthermore the repeat parameter must be the last one)::
+    * parameters as space separated string::
 
         self.execMacro('ascan %s 0 100 10 0.2' % motor.getName())
-        self.execMacro('mv %s 0' % motor.getName())
-
+        self.execMacro('mv [[%s 0]]' % motor.getName()) 
+        self.execMacro('mv [%s 0]' % motor.getName()) # simplified interface
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

doc/source/devel/howto_macros/macros_general.rst
```diff
@@ -596,12 +596,16 @@ parameters with different *flavors*:
...
+        self.execMacro('mv %s 0' % motor.getName()) # backwards compatibility - see note
+        self.execMacro('mv [[%s 0][%s 20]]' % (motor.getName(), motor2.getName()))
+        self.execMacro('mv [%s 0 %s 20]' % (motor.getName(), motor2.getName())) # simplified interface
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

doc/source/devel/howto_macros/macros_general.rst
```diff
@@ -613,7 +617,9 @@ parameters with different *flavors*:
     use :ref:`repeat parameters <sardana-macro-repeat-parameters>`.
     From Sardana 2.0 the repeat parameter values must be passed as lists of
     items. An item of a repeat parameter containing more than one member is a
-    list. In case when a macro defines only one repeat parameter
```
reszelaz: After implementing changes agreed in the last #405 comments this is not true (execMacro with arguments as concrete types has not changed). Please undo this change if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -25,11 +25,15 @@
...
-__all__ = ["pt0", "pt1", "pt2", "pt3", "pt4", "pt5", "pt6", "pt7", "pt8",
-           "pt9"]
+__all__ = ["pt0", "pt1", "pt2", "pt3", "pt3d", "pt4", "pt5", "pt6", "pt7",
+           "pt7d1", "pt7d2", "pt8", "pt9", "pt10", "pt11", "pt12", "pt13",
+           "pt14", "pt14d", "twice"]
 
 class pt0(Macro):
-    """Macro without parameters. Pretty dull"""
+    """Macro without parameters. Pretty dull.
+       Usage, ex.:
```
reszelaz: Here, and for all below examples, I would say that this is "Usage from spock:"

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -60,20 +70,51 @@ class pt3(Macro):
     """Macro with a list of numbers as parameter: the type is a sequence of
     parameter types which is repeated. In this case it is a repetition of a 
     float so only one parameter is defined.
-    By default the repetition as a semantics of 'at least one'"""
+    By default the repetition as a semantics of 'at least one'
+    Usages, ex.:
+    pt3 [[1][34][15]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -60,20 +70,51 @@ class pt3(Macro):
...
     def run(self, *args, **kwargs):
         pass
+    
+class pt3d(Macro):
+    """Macro with a list of numbers as parameter: the type is a sequence of
+    parameter types which is repeated. In this case it is a repetition of a 
+    float so only one parameter is defined. The parameter has a default value.
+    By default the repetition as a semantics of 'at least one'
+    Usages, ex.:
+    pt3d [[1][34][15]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -60,20 +70,51 @@ class pt3(Macro):
...
+    pt3d [1 34 15]
+    pt3d 1 34 15
+    Usages taken the default value, ex.:
+    pt3d [[1] [] [15]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -60,20 +70,51 @@ class pt3(Macro):
...
+        pass
 
 class pt4(Macro):
     """Macro with a list of motors as parameter: the type is a sequence of
     parameter types which is repeated. In this case it is a repetition of a 
     motor so only one parameter is defined.
-    By default the repetition as a semantics of 'at least one'"""
+    By default the repetition as a semantics of 'at least one'.
+    Usages, ex.:
+    pt3 [[mot1][mot2][mot3]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -83,7 +124,12 @@ def run(self, *args, **kwargs):
         pass
 
 class pt5(Macro):
-    """Macro with a motor parameter followed by a list of numbers"""
+    """Macro with a motor parameter followed by a list of numbers.
+    Usages, ex.:
+    pt5 mot1 [[1][3]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -92,11 +138,16 @@ class pt5(Macro):
...
         pass
-
+ 
 class pt6(Macro):
     """Macro with a motor parameter followed by a list of numbers. The list as
     explicitly stated an optional last element which is a dictionary that defines the
-    min and max values for repetitions"""
+    min and max values for repetitions.
+    Usages, ex.:
+    pt6 mot1 [[1][34][1]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -107,7 +158,12 @@ def run(self, *args, **kwargs):
         pass
     
 class pt7(Macro):
-    """Macro with a list of pair Motor,Float"""
+    """Macro with a list of pair Motor,Float.
+    Usages, ex.:
+    pt7 [[mot1 1][mot2 3]]
+    pt7 [mot1 1 mot2 3]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -117,10 +173,57 @@ class pt7(Macro):
     
     def run(self, *args, **kwargs):
         pass
+    
+class pt7d1(Macro):
+    """Macro with a list of pair Motor,Float. Default value for last ParamRepeat element.
+    Usages, ex.:
+    pt7d1 [[mot1 1][mot2 3]]
+    pt7d1 [mot1 1 mot2 3]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -117,10 +173,57 @@ class pt7(Macro):
...
+    pt7d1 mot1 1 mot2 3
+    Using default value, ex.:
+    pt7d1 [[mot1][mot2 3]] # at any repetition
+    pt7d1 [mot1 1 [mot2]] # only at the last repetition
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -117,10 +173,57 @@ class pt7(Macro):
...
+    
+    def run(self, *args, **kwargs):
+        pass
+
+    
+class pt7d2(Macro):
+    """Macro with a list of pair Motor,Float. Default value for both ParamRepeat elements.
+    Usages, ex.:
+    pt7d2 [[mot1 1][mot2 3]]
+    pt7d2 [mot1 1 mot2 3]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -117,10 +173,57 @@ class pt7(Macro):
...
+    pt7d2 mot1 1 mot2 3
+    Using both default values, ex.:
+    pt7d1 [[][mot2 3][]] # at any repetition
+    pt7d1 [mot1 1 []] # only at the last repetition
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -117,10 +173,57 @@ class pt7(Macro):
...
+
+    
+class pt7d2(Macro):
+    """Macro with a list of pair Motor,Float. Default value for both ParamRepeat elements.
+    Usages, ex.:
+    pt7d2 [[mot1 1][mot2 3]]
+    pt7d2 [mot1 1 mot2 3]
+    pt7d2 mot1 1 mot2 3
+    Using both default values, ex.:
+    pt7d1 [[][mot2 3][]] # at any repetition
```
reszelaz: This should be pt7d2

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -117,10 +173,57 @@ class pt7(Macro):
...
+    def run(self, *args, **kwargs):
+        pass   
 
 class pt8(Macro):
     """Macro with a list of pair Motor,Float. The min and max elements have been
-    explicitly stated"""
+    explicitly stated.
+    Usages, ex.:
+    pt8 [[mot1 1][mot2 3]]
+    pt8 [mot1 1 mot2 3]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -135,7 +238,12 @@ def run(self, *args, **kwargs):
 class pt9(Macro):
     """Same as macro pt7 but with old style ParamRepeat. If you are writing
     a macro with variable number of parameters for the first time don't even
-    bother to look at this example since it is DEPRECATED."""
+    bother to look at this example since it is DEPRECATED.
+    Usages, ex.:
+    pt9 [[mot1 1][mot2 3]]
+    pt9 [mot1 1 mot2 3]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -149,7 +257,11 @@ def run(self, *args, **kwargs):
 
 class pt10(Macro):
     """Macro with list of numbers followed by a motor parameter. The repeat
-    parameter may be defined as first one."""
+    parameter may be defined as first one.
+    Usages, ex.:
+    pt10 [[1][3]] mot1
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -158,11 +270,15 @@ class pt10(Macro):
...
         pass
-
+  
 
 class pt11(Macro):
     """Macro with list of numbers followed by a motor parameter. The repeat
-    parameter may be defined as first one."""
+    parameter may be defined as first one.
+    Usages, ex.:
+    pt11 ct1 [[1][3]] mot1
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -176,7 +292,11 @@ def run(self, *args, **kwargs):
 
 class pt12(Macro):
     """Macro with list of motors followed by list of numbers. Two repeat
-    parameters may defined."""
+    parameters may defined.
+    Usages, ex.:
+    pt12 [[1][3][4]] [[mot1][mot2]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -189,7 +309,11 @@ def run(self, *args, **kwargs):
 
 class pt13(Macro):
     """Macro with list of motors groups, where each motor group is a list of 
-    motors. Repeat parameters may be defined as nested."""
+    motors. Repeat parameters may be defined as nested.
+    Usages, ex.:
+    pt13 [[[[mot1] [mot2]]][[[mot3] [mot4]]]]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
     def run(self, *args, **kwargs):
         pass
 
+       
+class pt14(Macro):
+    """Macro with list of motors groups, where each motor group is a list of 
+    motors and a float. Repeat parameters may be defined as nested.
+    Usages, ex.:
+    pt14 [[[[mot61] [mot62]] 3]  [[[mot63]] 5] ]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    pt14 [[mot61 mot62] 3  [mot63] 5]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

reszelaz: After implementing the proposed changes the correct example is:

`pt14 [[[mot61 mot62] 3]  [[mot63] 5]]`

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
 
+    def run(self, *args, **kwargs):
+        pass 
+       
+class pt14d(Macro):
+    """Macro with list of motors groups, where each motor group is a list of 
+    motors and a float. Repeat parameters may be defined as nested.
+    Default values can be used.
+    Usages, ex.:
+    pt14d [[[[mot61] [mot62]] 3]  [[[mot63]] 5] ]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    pt14d [[mot61 mot62] 3  [mot63] 5]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    Usages taken default values, ex.:
+    pt14d [[[[mot61] [mot62]]]  [[[mot63]] 5] ]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    pt14d [[[[mot61] []]]  [[[mot63]] 5] ]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    pt14d [[[]]  [[[mot63]] 5] ]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    pt14d [[mot61 mot62] 3  [] 5]
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

src/sardana/macroserver/macros/examples/parameters.py
```diff
@@ -200,7 +324,50 @@ class pt13(Macro):
...
+    pt14d [[mot61 mot62] 3  []] # only at the last repetition
```
reszelaz: After implementing changes agreed in the last #405 comments this should not be accepted. Please remove if you agree.

reszelaz: After implementing the proposed changes the correct examples are:
`pt14d [[[mot61 mot62] 3] [[mot63] []]]`
`pt14d [[[mot61 []] 3] [[mot63] []]]`
`pt14d [[[[]] 3] [[mot03] []]]`

src/sardana/macroserver/macros/examples/parameters.py
sagiss: This line (pt11 ct1 [[1][3]] mot1) should be removed.

teresanunez: Hi Marc, it is already removed ...

sagiss:
Hi Teresa,

I have observed that with the modifications added by Zibi and me, we have introduced a regarding your previous commits.

This case, that should work, does not work: pt7d2 [[][mot42]]
(however, if we leave a whitespace it works correctly)

I will work on that and propose you another pull request.
Or if you have some idea of what is happening, I'm open to suggestions.

teresanunez:
Hi Marc, in fact what I have seen is that what you need is to have a space somewhere in the
arguments and then it will work, for example, this:
pt7d2 [[][exp_dmy02][exp_dmy01]]
does not work, but this:
pt7d2 [[][exp_dmy02 1][exp_dmy01]]
woks because there is an space, or
pt7d2 [[ ][exp_dmy02][exp_dmy01]]
or 
pt7d2 [ [][exp_dmy02][exp_dmy01]]
All the combinations having an space somewhere work.

teresanunez:
Hi Marc, I found the problem and a solution for that. I describe it now. Then you can modify it if you want.

teresanunez:
In the function split_macro_parameters (file spockms.py) you check the length of macro_params,
 and only if it is > 1 you search for brackets and create the list of macro_params, but if macro_params
 does not have any space, like in the case:

pt7d2 [[][mot42]]

macro_params have length 1, so you will never search for brackets and create the list correctly, spock
 see only a list with one parameter. If the length of macro_params is 1 you have to search for brackets,
 for deciding about converting or not the parameters to a proper string, for ex. instead of only:

if len(macro_params) > 1:

one need to check something like:

list_of_pars = 0
    if len(macro_params) == 1:
        if  macro_params[0].find('[') > -1:
            list_of_pars = 1

if len(macro_params) > 1 or list_of_pars:

Or any other more elegant way to make it, but this check has to be done. It is not enough with the
 length.

sagiss:
Great Teresa, I have tested your solution and it works well. Could you make a commit with it in order to add it in this pull request?

src/sardana/taurus/core/tango/sardana/test/test_macro.py
```diff
@@ -114,3 +121,175 @@ def verifyXML(self, macro_name, param_def, param_value, expected_xml_rep):
...
+            )
+
+class ParamsTestCase(unittest.TestCase):
+
+    def verifyEncoding(self, param_def, macro_params, expected_params_list):
+        """
+        Helper to verify the correct building of the parameters objects tree.
+        Verify that the list is recreated correctly from the parameters
+        objects tree.
+        :param macro_name: (str) name of the macro
```
sagiss: The comment "param macro_name: (str) name of the macro" should be removed because it is not used anymore.

src/sardana/spock/spockms.py
```diff
@@ -602,3 +602,32 @@ def _removeMacro(self, macro_info):
...
+            list_of_pars = 1
+    # skip empty strings and just one parameter
+    if len(macro_params) > 1 or list_of_pars:
+        params_str = ''
+        for par in macro_params:
+            params_str = params_str + str(par) + ","
+        params_str = params_str[:-1]
+        params_str = params_str.replace("[,", "[").replace(",]", "]").replace(
+            "][", "],[")
+        macro_params = eval(params_str, globals(), Reflector())
```
reszelaz: Using direct eval may be dangerous. Please use [SafeEvaluator](https://github.com/taurus-org/taurus/blob/develop/lib/taurus/core/util/safeeval.py#L35) instead which takes care of allowing only safe code to be evaluated.

reszelaz: Forget about my request. Actually since the eval is still executed on the client side (spock) there is no problem. This could be dangerous if we execute eval in the server.

PR #409, opened 2017-01-09
Warning instead of error if hkl can not be loaded

### LINKED ISSUE #394: Pool should not log an ERROR if hkl is not installed
If the HKL library is not installed, any invocation of the Pool logs this error (which is otherwise harmless , in the sense that it just indicates that you cannot use HKL stuff):

MainThread     ERROR    2016-12-22 10:32:41,590 ModuleManager: Invalid module 

IMHO, the error message is confusing, though. It should be an INFO or WARNING at most instead  (or even probably just a DEBUG). As it is, the message is likely to mislead in case of unrelated problems.

teresanunez:
PR to Fix #394

### DIFF
--- a/src/sardana/pool/poolcontrollers/HklPseudoMotorController.py
+++ b/src/sardana/pool/poolcontrollers/HklPseudoMotorController.py
@@ -48,8 +48,6 @@ import PyTango
 
 from itertools import chain
 
-from gi.repository import GLib
-from gi.repository import Hkl
 
 from taurus.core.util.codecs import CodecFactory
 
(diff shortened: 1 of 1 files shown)


cpascual:
As it is now, I get the message duplicated:
~~~
$ Pool demo2 
(...)
MainThread     WARNING  2017-01-11 12:18:54,881 ControllerManager: HklPseudoMotorController: Not able to load Hkl library. Diffractometer controller can not be used
MainThread     WARNING  2017-01-11 12:18:54,884 ControllerManager: HklPseudoMotorController: Not able to load Hkl library. Diffractometer controller can not be used
~~~

Is it just me?

teresanunez:
I also got it ... but I put it only once ... I seems to be something created by Sardana ...
 Loading twice? (load and reload).

Issue #408, opened 2017-01-05
Param Repeat of one element does not allow brackets


sagiss:
Param Repeat of one element does not allow brackets

We attach the macros used to test this issue in file repeats2.txt: this file has to be renamed to repeats2.py, charged in a macroserver and call the macros existing in the file from spock.
[repeats2.txt](https://github.com/sardana-org/sardana/files/687441/repeats2.txt)

A macro pt20_default has been created. This macro uses as repeat parameter, a repetitions of one float (list of Floats). The default value for a missing element of the list is specified.

The macro pt20_default is called using execmacro from three different macros (repeatsdefaultmac12, repeatsdefaultmac13, repeatsdefaultmac14). The repeatsdefaultmac12 works correctly because the list is complete and default values are not used. 

repeatsdefaultmac13 fails and report an error because internal brackets are used:
('pt20_default', [1, 2, [3]])
Internal brackets for single elements could be optionally accepted are not accepted currently, maybe it could be adapted but it is not mandatory.

However repeatsdefaultmac14 fails and report an error, and this should not be the case. In repeatsdefaultmac14 empty brackets are used in order to indicate that the default value should be used in that place. This can be seen in the spock output:

<pre>
repeatsdefaultmac14
An error occurred while running Macro 'repeatsdefaultmac14() -> 61f298a4-d353-11e6-9891-00163e482412':
float() argument must be a string or a number
</pre>

In order to indicate the usage of default values, empty brackets should be accepted.

teresanunez:
Hi, I have looked at this and solved the problem described with the macro repeatsdefaultmac13,
 now the [], so that the paramrepeat with only one element allow internal macros. The problem
 with the default values is still there, but this is a general problem not only for the paramrepeats with
only one parameter.

teresanunez:
The PR #416 should solve this issue. Also the problem with the default values.

reszelaz:
As agreed in #405 we will not accept mixing list and not list API in the same repeat parameter. So the example `repeatsdefaultmac13` from `repeats2.txt` should not be allowed. Nevertheless we should support empty list as the placeholder for indicating the default value.

teresanunez:
The PR concerning this issue was changed, now repeatsdefaultmac13 from repeats2.txt gives an error.

reszelaz:
Solved by #416.

PR #412, opened 2017-01-10
Accepting brackets for paramrepeats with one element

teresanunez:
This PR solves the issue #408, each repetition of a ParamRepeat with only one element
 can be closed into brackets. The problem of the default values pointed out in #408 concerns
all ParamRepeats and not only the ones with one element.


teresanunez:
I have to check this PR, it introduces another problem.

teresanunez:
I will investigate another implementation. A new pull request will be done.

PR #414, opened 2017-01-11
Fix issue407: default values for repeat parameters

sagiss:
Fix bug 407.

When using execMacro with repeat parameters, the default values
are not taken into account correctly.

This commit fixes the bug.


teresanunez:
I have reviewed and it solves the problem  (it is still open the problem with the ParamRepeats with
 only one element #408 which implies that in that case default values can still not be used).

PR #415, opened 2017-01-12
Hotfix 2.2.3

cmft:
Note for @sardana-org/integrators, after merging into master, it will require also making an annotated tag: 2.2.3. Afterwards I will make a new PR to develop with the bumped version number to 2.2.4-alhpa.


teresanunez:
The tag 2.2.3 in master is created.

PR #416, opened 2017-01-13
Accepting brackets for single paramrepeat and default values

teresanunez:
Hi, this PR should solve the issue #408, allowing to use brackets for the single element paramrepeat
arguments. 
For example a macro like:

 
```
class pt20_default(Macro):
    """Macro with a list of numbers as parameter."""

    param_def = [
       [ 'numb_list', [ [ 'pos', Type.Float, 3.0, 'value'] ], None, 'List of values'],
    ]

        self.output(args)
        pass

```
Could be called like:
```
        params = ('pt20_default', [1, 2, 3])
        self.execMacro(params)

```
or

```
        params = ('pt20_default', [1, 2, [3]])
        self.execMacro(params)

```
Also the issue with the default values should be solved, so one could use:

```
        params = ('pt20_default', [1, 2, []])
        self.execMacro(params)
```

The issue with the default values should be solved in general, also for ParamRepeat with several elements. For example:
```

class pt1_default(Macro):
    """Macro with a list of  float pairs"""

    param_def = [
       [ 'p_p_pair', [ [ 'pos', Type.Float, 21, 'Position1'],
                       [ 'pos',   Type.Float, 22, 'Position2'] ],
         None, 'List of position/position pairs']
    ]

        self.output(args)
        pass
```
 
could be called:

 
```
        params = ('pt1_default', [[2], []])
        self.execMacro(params)
```

and the default values should be correctly taken.

### DIFF
--- a/src/sardana/macroserver/msparameter.py
+++ b/src/sardana/macroserver/msparameter.py
@@ -399,7 +399,8 @@ class ParamDecoder:
                     if len(value) == 0:
                         value = param_def['default_value']
                     else:
-                        value = value[0]
+                        msg = 'API not allowed'
+                        raise WrongParam, msg
                 if value is None:
                     value = param_def['default_value']
                 if value is None:
--- a/src/sardana/macroserver/msparameter.py
+++ b/src/sardana/macroserver/msparameter.py
@@ -395,13 +395,8 @@ class ParamDecoder:
                     value = raw_param.get("value")
                 else:
                     value = raw_param
-                if type(value) == list:
-                    if len(value) == 0:
-                        value = param_def['default_value']
-                    else:
-                        msg = 'Brackets not allowed for single element repetitions'
-                        raise WrongParam, msg
-                if value is None:
+                # None or [] indicates default value
+                if value is None or (type(value) == list and len(value) == 0):
                     value = param_def['default_value']
                 if value is None:
                     raise MissingParam, "'%s' not specified" % name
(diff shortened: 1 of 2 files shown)


sagiss:
I have reviewed the pull request and it works well. 

With this implementation, default parameters can also be used in the first place, by using empty brackets, which seems a good feature:
<pre>
        params = ('pt1_default', [[2, 3], [[], 4]])
        self.execMacro(params)
</pre>

teresanunez:
Marc, many thanks for the review.

src/sardana/macroserver/msparameter.py
```diff
@@ -395,6 +395,11 @@ def decodeNormal(self, raw_param, param_def):
                     value = raw_param.get("value")
                 else:
                     value = raw_param
+                if type(value) == list:
+                    if len(value) == 0:
+                        value = param_def['default_value']
+                    else:
+                        value = value[0]
```
reszelaz: Since in #405 we agreed that the repeat parameters composed from just one member are not passed as list I would raise `WrongParam` exception here.

teresanunez:
Hi Zibi, I have changed this, now the brackets closing single elements are not allowed. I put an
 WrongParam exception, as you said, but be aware that all errors in this function are shown in
 spock as:

float() argument must be a string or a number

independently of where or why the error is.

sagiss:
I have verified the behavior reported by Teresa 

spock reports:
float() argument must be a string or a number

We should find the cause of this.

teresanunez:
Hi marc, I have just sent a commit with a way of solving the problem with the error that is shown in
 spock, only to discuss it here.

reszelaz:
With the solution proposed by @teresanunez the correct exception with the self explaining reason is raised now. I have just slightly modified it and moved it to the `decodeRepeat` instead of `decodeNormal` (see the last commit message). I proceed with the merge and closure of #408.

PR #419, opened 2017-01-23
Fix  #418 TaurusSequencerWidget fails with stopped MS

### LINKED ISSUE #418: TaurusSequencerWidget fails with stopped MS
TaurusSequencerWidget.SetModel raises an TypeError when the chosen MS is not connected.

``` python
  File "/home/cfalcon/tools/sardana-sardana.git/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py", line 814, in setModel
    self.connect(newModelObj, Qt.SIGNAL("macrosUpdated"), self.macroComboBox.onMacrosUpdated)
TypeError: arguments did not match any overloaded call:
  QObject.connect(QObject, SIGNAL(), QObject, SLOT(), Qt.ConnectionType=Qt.AutoConnection): argument 1 has unexpected type 'NoneType'
  QObject.connect(QObject, SIGNAL(), callable, Qt.ConnectionType=Qt.AutoConnection): argument 1 has unexpected type 'NoneType'
  QObject.connect(QObject, SIGNAL(), SLOT(), Qt.ConnectionType=Qt.AutoConnection): argument 1 has unexpected type 'NoneType'
```

cmft:
TaurusSequencerWidget.SetModel raises an Type error when the
chosen MS is not connected and the door is not defined.

Fix it, checking the value of the new model obj before connect the
signal.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/motion.py
+++ b/src/sardana/taurus/core/tango/sardana/motion.py
@@ -70,7 +70,7 @@ class Moveable:
     def getLastMotionTime(self):
         raise NotImplementedError
 
-    def getTotalLastMotionTime():
+    def getTotalLastMotionTime(self):
         raise NotImplementedError
 
     def abort(self, wait_ready=True, timeout=None):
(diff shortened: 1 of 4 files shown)


reszelaz:
I was not able to reproduce the bug described in #418. Neither the requestor was able to reproduce it. So I will reject the ticket and this PR. 

I have cherry-picked the last two commits which are useful. The first of them fixes buggy behaviour when using a tango host different than the default one. This bug was affecting only Taurus 4 and in order to fix it one needs the latest version of Taurus 4.

PR #420, opened 2017-01-27
DummyMotorController: related to issue #30

teresanunez:
Trying to run a continuous scans with the sep6 (ascanct macro) and using as a motor one
from a DummyMotorController, it can happen that the velocity computed for running the scan
is equal to the BaseRate. In that case one get an exception but one does not see where
the problem comes from, since the exception comes before the computed velocity is
actually set to the device. This PR solves that problem. The implementation of the DummyMotorController would accept infinity acceleration and deceleration and zero
 acceleration and deceleration times. The PR also include a work around added by Zibi
 to his sep6 branch and mentioned in issue #30. The PR also solves issues from #30.

### DIFF
--- a/src/sardana/pool/poolcontrollers/DummyMotorController.py
+++ b/src/sardana/pool/poolcontrollers/DummyMotorController.py
@@ -22,7 +22,7 @@
 ##############################################################################
 
 import time
-from math import pow, sqrt
+from math import pow, sqrt, isinf
 
 from sardana import State, SardanaValue
 from sardana.pool.controller import MotorController
@@ -151,11 +151,15 @@ class Motion(BaseMotion):
     def setAccelerationTime(self,at):
         """Sets the time to go from minimum velocity to maximum velocity in seconds"""
         at = float(at)
-        if at <= 0:
-            raise "Acceleration time must be > 0"
-
+        if at < 0:
+            at = 0
+            
         self.accel_time = at
-        self.accel = (self.max_vel - self.min_vel) / at
+        if at != 0:
+            self.accel = (self.max_vel - self.min_vel) / at
+        else:
+            self.accel = float('INF')
+            
 
         self.__recalculate_acc_constants()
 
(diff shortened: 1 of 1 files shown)


src/sardana/pool/poolcontrollers/DummyMotorController.py
```diff
@@ -569,8 +588,9 @@ def PreStartOne(self, axis, pos):
     def StartOne(self, axis, pos):
         #raise Exception("Cannot move on StartOne")
         idx = axis - 1
+        m = self.m[idx]
```
sagiss: Is it necessary? 
'm' does not seem to be used.

src/sardana/pool/poolcontrollers/DummyMotorController.py
```diff
@@ -165,12 +169,15 @@ def getAccelerationTime(self):
...
+        if dt < 0:
+            dt = 0
 
         self.decel_time = dt
-        self.decel = (self.min_vel - self.max_vel) / dt
-
+        if dt != 0:
+            self.decel = (self.min_vel - self.max_vel) / dt
+        else:
+            self.decel = float ('INF')
```
sagiss: I think that instead of self.decel=float('INF') in this case it should be self.decel=float('-INF').
It has not any major implication in the code, but the sign should be negative because it is a deceleration.

src/sardana/pool/poolcontrollers/DummyMotorController.py
teresanunez: I don't know why this is there ... I probably took an old version not cleaned up, I didn't put. 
Yes, it should not be there. Thanks.

src/sardana/pool/poolcontrollers/DummyMotorController.py
teresanunez: I agree, you can change it to -inf. Thanks.

sagiss:
I have observed a problem applying the pull request.
Before the change, the acceleration time and deceleration time could not be set to 0. Now they can.

If we set for a dummy motor the acceleration time to 0, and the deceleration time to a certain number (4 for example), and we launch:
<pre>
mot01.acceleration = 0
mot01.deceleartion = 4
%dscan mot01 0 30 3 0.1
</pre>

In this case we are in presence of a small movement. We get an error of one of the asserts that is not fulfilled:
assert(self.curr_max_vel_pos >= self.init_pos)

I'm am trying to get a look to this things, but if acceleration times of 0 are allowed, many things changes. 
We should maybe treat the four cases:
1) acceleration time != 0     and     deceleration time !=0
2) acceleration time == 0     and     deceleration time ==0 
3) acceleration time != 0     and     deceleration time ==0
4) acceleration time == 0     and     deceleration time !=0

teresanunez:
Hi Marc, thanks for the check, and sorry,  I forgot to check some possible combinations. Do you want to work on that, or should I also have a look?.

sagiss:
Hi Teresa. I'm having a look to it for the moment. If I have problems I will let you know. Thanks

PR #422, opened 2017-01-30
Improve main in macrobutton.py module

jordiandreu:
The code enables to execute any macro (except param repeat
with multiple values) from a macrobutton. Two arguments
are required to initalize the widget: door name and macro name.
Once the widget is running, the macro name can be changed and
the interface to supply the parameters is updated automatically.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
@@ -385,8 +385,15 @@ if __name__ == '__main__':
             param_names = []
             default_values = []
             for p in pars:
-                param_names.append(p['name'])
-                default_values.append(p['default_value'])
+                ptype = p['type']
+
+                if isinstance(ptype, list):
+                    for pr in ptype:
+                        param_names.append(pr['name'])
+                        default_values.append(pr['default_value'])
+                else:
+                    param_names.append(p['name'])
+                    default_values.append(p['default_value'])
 
             self.macro_name = macro_name 
             return param_names, default_values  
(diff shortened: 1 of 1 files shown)


src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
```diff
@@ -329,97 +329,138 @@ def abort(self):
...
+            # Create TOP layout
+            self.w_arg = Qt.QWidget()
+            self.w_arg.setLayout(Qt.QGridLayout())
+            col = 0
+            self.w_arg.layout().addWidget(Qt.QLabel('macro name'), 0, col)
+            self.w_macro_name = Qt.QLineEdit()
+            self.w_arg.layout().addWidget(self.w_macro_name, 1, col)
+
+            _argEditors = []
+            for name in p_names:
```
reszelaz: This does not consider param repeats composed from more than one params.

In general I would limit the purpose of this widget to very simple cases of macros with repeat parameters:
* only one repeat parameter (no nested repeat parameters are allowed)
* repeat parameter located as the last one

In case of repeat parameters with more than one param e.g. `mv` macro the widget should populate two `QLineEdit` instances, one for the moveable and one for the position. Now it populates only one of them.

jordiandreu: We can easily make the widget able to accept repeat parameters in any position, but only a single value could be supplied.

reszelaz: I would not go for it. Let's keep the "demo" simple. In sardana we have already a generic widget to execute macros - macroexecutor.

PR #425, opened 2017-02-06
Fix bug#423 Macrobutton fails parsing string param. with spaces

cmft:
MacroButton fails to execute macros which its argument is a string with spaces


reszelaz:
I agree on the changes.

PR #429, opened 2017-02-22
Add option for node duplication in macroexecutor (issue 426)

sagiss:
Add option for node duplication in macroexecutor (issue 426).
Add button in macroexecutor for allowing RepeatNode duplication.

### DIFF
--- a/src/sardana/macroserver/macros/examples/parameters.py
+++ b/src/sardana/macroserver/macros/examples/parameters.py
@@ -259,8 +259,9 @@ class pt10(Macro):
   
 
 class pt11(Macro):
-    """Macro with list of numbers followed by a motor parameter. The repeat
-    parameter may be defined as first one.
+    """Macro with counter parameter followed by a list of numbers, followed by
+    a motor parameter. The repeat parameter may be defined in between other
+    parameters.
     Usages from Spock, ex.:
     pt11 ct1 [1 3] mot1
     """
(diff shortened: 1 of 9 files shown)


sagiss:
Tests Ok:
- Tested with macros generic 'wm' and 'mv'.
- Tested with 'manytomos' specific macro from beamline scientists.

sagiss:
After testing sequencer it presented failures. The nodes where not duplicated till clicking on the button for adding nodes. Then a refresh of the view was done. 

I will propose changes in the code.

sagiss:
Review: The icon gtk-copy is not visible using sequencer with taurus3.  Fix it.

cpascual:
Oooops!! 
Sorry, I wanted to push just the last commit (a6a9859), but I pushed also all the changes in develop since you branched. 
Anyway, that is not beautiful, but neither harmful... Just ignore from a08bd50 till c3ccf03 if you are reviewing

cpascual:
The feature seems to work fine using Taurus4, both for macroexecutor and for the sequenceeditor, **but** I found the following issues:

- It did not work (the duplicate button did not appear) when run in a Suse12 (PyQt4.4, taurus3,...) neither in  macroexecutor nor in the sequenceeditor.
- I expected (maybe unjustifiably?)  that I would be able to duplicate whole macro nodes in the sequencer just as one can duplicate a param repeat node... but that is not the case.
- The copy button (at  least in my system) changes very little when it is disabled (it makes it difficult to tell if the action is enabled or disabled)

Of the three, only the first seems important to me, and if it is difficult to implement, I would agree tomerge this PR and open a bug for porting it to older libs.

PR #440, opened 2017-03-22
Forcing readout from hw

teresanunez:
Changes only affecting the data format used at DESY. 
It is harmless to merge it to the develop branch.

### DIFF
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -137,11 +137,11 @@ class FIO_FileRecorder(BaseFileRecorder):
             all_motors = self.macro.findObjs('.*', type_class=Type.Motor)
             all_motors.sort()
             for mot in all_motors:
-                pos = mot.getPosition()
+                pos = mot.getPosition(force=True)
                 if pos is None:
                     record = "%s = nan\n" % (mot)
                 else:
-                    record = "%s = %g\n" % (mot, mot.getPosition())
+                    record = "%s = %g\n" % (mot, mot.getPosition(force=True))
                     
                 self.fd.write( record)
             self.fd.flush()
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -141,7 +141,7 @@ class FIO_FileRecorder(BaseFileRecorder):
                 if pos is None:
                     record = "%s = nan\n" % (mot)
                 else:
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi Zibi, thanks, should I merge it by myself?. 
Yes, it makes the same as PreScanSnapshot but it is putting the output in a standard DESY
format file.

cpascual:
I haven't looked at the code, but based on the comments above, I wonder if the same functionality could be implemented using the PreScanSnapshot API of the recorder...

teresanunez:
Hi Carlos, it was implemented by Thorsten already years ago, it is for keeping the same format the people from DESY was used to (the fio format). The only change now is that we add force=True in the argument of getPosition. I will keep it like it is, so that we have this format independent and like the users
 want it here.

reszelaz:
I think it is safe to merge this change now, so I proceed with it.

I think that using the `PreScanSnapshot` instead of the in-place readouts will not affect the way you save data in the file (format). It is configurable by the user - it allows to select just a subset of the motors and not all of them as you use it now. Apart of that the `PreScanSnapshot` is read only once and the values are passed to all the configured recorders. So in my opinion it has a lot of benefits over the way you do it in the FIO recorder and it is the "standard" way. I see only one drawback, that it will require to update the `PreScanSnapshot` group whenever you add a new motor to the system - now you read all of the motors.

src/sardana/macroserver/recorders/storage.py
```diff
@@ -137,11 +137,11 @@ def _startRecordList(self, recordlist):
             all_motors = self.macro.findObjs('.*', type_class=Type.Motor)
             all_motors.sort()
             for mot in all_motors:
-                pos = mot.getPosition()
+                pos = mot.getPosition(force=True)
                 if pos is None:
                     record = "%s = nan\n" % (mot)
                 else:
-                    record = "%s = %g\n" % (mot, mot.getPosition())
+                    record = "%s = %g\n" % (mot, mot.getPosition(force=True))
```
reszelaz: Is it necessary to read it again? Since it is not None, it should have the value already.
I would use: `record = "%s = %g\n" % (mot, pos)`

teresanunez: Hi Zibi, I see it like you ... I ask Thorsten, if he has a reason for repeating the readout ...
Thanks for the review.

teresanunez:
Hi Zibi, there was not reason for the double readout. It is corrected now. Many thanks.

PR #443, opened 2017-04-01
Fix #442: allow escaped quotes in parameters (spock)

### LINKED ISSUE #442: backward comatibility problems of senv
Hi,

Since #434 has been already closed I create another ticket for the bug, i.e.

    senv ScanFile "[\"file.nxs\",\"file.fio\"]"

no more works.

Bests,
Jan

reszelaz:
Use negative lookbehind assertion in the regular expression describing
quoted parameter in order to allow passing quotes escaped by backslash.

Also avoid string cutting in order to exclude quotes and single quotes
from the string parameters. Do it by explicit extraction of the given group.

### DIFF
--- a/src/sardana/spock/parser.py
+++ b/src/sardana/spock/parser.py
@@ -32,7 +32,7 @@ import collections
 
 # Token specification
 PARAM = r"(?P<PARAM>[^\[\]\s]+)"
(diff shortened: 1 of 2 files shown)


reszelaz:
@sardana-org/integrators can you review this? @jkotan had reported that it works for him.

PR #450, opened 2017-04-11
Add flake8 check on diff to travis-ci

reszelaz:
Implements flake8 check on the diff e.g. only on the changes introduced in the pull request. In case the pull request is not flake8 compliant (PEP8 and others) it fails the travis-ci. The tests are run anyway.

Apart of that the included script has more applications - see documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,11 @@ python:
   - "2.7"
 
 before_install:
+  - sudo apt-get update -qq
+  # install flake8 to perform python code style check in the script part
+  - sudo apt-get install -qq python-flake8
+
+install:
   # run reszelaz/sardana-test docker container (Debian8 with sardana-deps)
   - docker pull reszelaz/sardana-test
   - docker run -d --name=sardana-test -h sardana-test --volume=`pwd`:/sardana reszelaz/sardana-test
@@ -21,5 +26,7 @@ before_install:
   - docker exec sardana-test supervisorctl start MacroServer
 
 script:
+  # run flake8 check on all python files in the project
+  - ci/flake8_diff.sh
   # run the full testsuite
   - docker exec sardana-test sardanatestsuite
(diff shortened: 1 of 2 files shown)


cpascual:
Great addition!

Just one comment:
Maybe it would be even better if instead of running the flake test in the same travis build as the testsuite, you used a [travis build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) to split this into 2 travis runs. This way, Travis would run 2 builds: one for the test suite and one for flake.
This could allow to "automate" a message in case of failure saying, "check PEP8 on your PR" and give a link to the travis result for flake8.

(it is just an idea... I would also be ok with the current approach)
If you like the idea and want to implement it, you can have a look at the taurus travis yml

cpascual:
One more comment: maybe the flake8 install could be done in the docker already, instead of in travis

reszelaz:
I agree that we should improve the readability of travis-ci logs in case of the flake8 failure. Now one needs to read the logs carefully to figure out what failed. I will check how it improves with the matrix configuration.

Regarding running the flake8 check in the docker container, I don't mind moving it there. But do you see a strong benefits of doing it in this way, apart of having a more self-consisted docker image? Furthermore, since the script is using some travis environment variables, it will require exporting them to the container, which I think will make the configuration a bit complicated.

In any case, if there are no negative comments, I vote for merging it as it is now, and postpone the improvements for the future. But first it is necessary to merge #446, @cpascual can you also look at it as the PEP8 expert :) ?

cpascual:
> Regarding running the flake8 check in the docker container

actually, I withdraw my proposal. I think it is cleaner in the travis machine instead of the docker. 

>  I vote for merging it as it is now

fine with me

cpascual:
I merged, it as is.
@reszelaz , would you open a new issue (enhancement) with the matrix thing?

reszelaz:
Done. See #452.

PR #454, opened 2017-04-20
Copy-paste bug in read_Value function

teresanunez:
We found a, I guess, copy-paste bug in the read_Value function of the TwoD and OneD experimental channels, introduced with the sep6 implementation. This PR only correct the copy-paste error.


reszelaz:
Very good! I have also seen it: https://github.com/reszelaz/sardana/commit/3b25fc3b3b50bba78dc3a4ad3f26ecbf55a1ee69 when implementing a proof-of-concept support for 1D in contscan. But it is not yet ready to be merged. So I will merge your solution. Furthermore your solution also corrects the same copy&paste error for 2D.

PR #457, opened 2017-04-25
1D support in continuous scans

reszelaz:
This development was done still before integration of the SEP6 between @droldan and myself. Now I have just updated it with the latest develop (there were some PEP8 issues to merge manually) in order to facilitate the eventual merge. @rhomspuron had already tried this branch in a real setup at Alba with the Mythen detector - please report here any issues you observed so we could fix them.

The implementation includes:
- Data transfer between the Pool and MacroServer:
  - Data attribute added to the OneDExpChannel
  - correctly treating data chunks reported by the 1D controllers
  - there are some optimizations related to passing data pending but I would address them in the future.
- DummyOneDController was prepared to be used with the HardwareTrigger synchronization


rhomspuron:
We installed this version of sadana on CLAESS (BL22) and the user feedbacks are positive. On the other hands we found some problems:
* The interpolation with the Mythen (1d channel) block the system, the Mythen uses trigger by hardware, but for one reason the recorder does not receive the data when the interpolation is enabled.
* After multiple ctl+c on spock of the ascanct, the Pool is hanged. We use a macro qExasf which calls the ascanct. 
* One of the two adlink card controller randomly is in state moving even though the DS is in state ON, this happen on the step scan and synchronization by hardware.
 We are working to solve these problem

reszelaz:
> The interpolation with the Mythen (1d channel) block the system, the Mythen uses trigger by hardware, but for one reason the recorder does not receive the data when the interpolation is enabled.

This is already solved.

> After multiple ctl+c on spock of the ascanct, the Pool is hanged. We use a macro qExasf which calls the ascanct.

I'm not able to reproduce this problem. Do you know how to do it? Are you able to reproduce it directly with the `ascanct` and dummy controllers? 

> One of the two adlink card controller randomly is in state moving even though the DS is in state ON, this happen on the step scan and synchronization by hardware.
> We are working to solve these problem

I assume that this is not a problem with Sardana but with a specific controller.

reszelaz:
I have just rewritten this with the `PoolExpChannelDevice` class introduced in #469 .

I have also fixed some other bugs such as:
* SPEC recorder problems with handling 1D's missing data indicated by "NaN"
* ascanct problems with 1D channels synchronized by software trigger

I think it is now ready for review and eventual integration. Any volunteer among @sardana-org/integrators ?

teresanunez:
Hi Zibi, thanks for the implementation. I can review it.

teresanunez:
I have reviewed this and for me it is working as expected and without errors. I have not tested it with 
real hardware trigger generator since I do not have one ready for tests just now. But I would say
that the PR could be merged. Should we wait for any other review or could I merge it now?.

reszelaz:
If you have tested it with dummies, and everything works I would say it is fine to accept it. I bet that for @rhomspuron or other users it will be easier to test it with the real hardware if it is already in develop. We could treat possible bugs as issues later on. Thanks for the tests!

teresanunez:
Yes, I have tested it with dummies. So I would merge. Thanks to you.

Issue #458, opened 2017-05-03
IORegisterController WriteOne method gives a integer as value


rhomspuron:
The IORegisterController WriteOne method gives a integer in the value. The attribute Value in the definition is a float:

`
class IORegisterController(Controller, Readable):
    """Base class for a IORegister controller. Inherit from this class to
    implement your own IORegister controller for the device pool.
    """

    #: .. deprecated:: 1.0
    #:     use :attr:`~Controller.axis_attributes` instead
    predefined_values = ()

    #: A :class:`dict` containing the standard attributes present on each axis
    #: device
    standard_axis_attributes = {
        'Value'       : { 'type' : float,
                          'description' : 'Value', },
    }

    #: A :obj:`str` representing the controller gender
    gender = 'I/O register controller'
    
    def __init__(self, inst, props, *args, **kwargs):
        Controller.__init__(self, inst, props, *args, **kwargs)

    def WriteOne(self):
        """**IORegister Controller API**. Override if necessary."""
        pass
`

In addition the base class  IORegisterController the method WriteOne does not receives the axis and the value, but the PoolIoregister call the method write_one:

`
 def write_register(self, value, timestamp=None):
        self._aborted = False
        self._stopped = False
        if not self._simulation_mode:
            if timestamp is None:
                timestamp = time.time()
            self.set_write_value(value, timestamp=timestamp, propagate=0)
self.controller.write_one(self.axis, value)
`

The method write_one is implemented on the class PoolController and it call the Method WriteOne with axis and value: 

`
    # START SPECIFIC TO IOR CONTROLLER ---------------------------------------

    def write_one(self, axis, value):
        self.ctrl.WriteOne(axis, value)

    # END SPECIFIC TO IOR CONTROLLER -----------------------------------------

`

PR #459, opened 2017-05-03
Issue #458: IORegisterController WriteOne method gives a integer as value

rhomspuron:
This is a possible solution for the bug.

### DIFF
--- a/src/sardana/macroserver/macros/ioregister.py
+++ b/src/sardana/macroserver/macros/ioregister.py
@@ -37,7 +37,7 @@ class write_ioreg(Macro):
 
     param_def = [
         ['input/output register', Type.IORegister, None, 'input/output register'],
-        ['data', Type.Integer, None, 'value to be send']
+        ['data', Type.Float, None, 'value to be send']
     ]
 
     def run(self, ioreg, data):
@@ -55,13 +55,14 @@ class read_ioreg(Macro):
     ]
 
     result_def = [
-        ['input/output register value', Type.Integer, None, 'value read from ' +
+        ['input/output register value', Type.Float, None, 'value read from ' +
          'the input/output register']
     ]
 
     def run(self, ioreg):
         name = ioreg.getName()
         data = ioreg.readIORegister(force=True)
+        data = float(data)
         o = "Reading " + name + " register "
         self.debug(o)
         return data
--- a/src/sardana/pool/controller.py
+++ b/src/sardana/pool/controller.py
@@ -1435,6 +1435,7 @@ class IORegisterController(Controller, Readable):
         'Value': {'type': float,
                   'description': 'Value', },
     }
+    standard_axis_attributes.update(Controller.standard_axis_attributes)
 
     #: A :obj:`str` representing the controller gender
     gender = 'I/O register controller'
--- a/src/sardana/pool/controller.py
+++ b/src/sardana/pool/controller.py
@@ -1442,6 +1442,6 @@ class IORegisterController(Controller, Readable):
     def __init__(self, inst, props, *args, **kwargs):
         Controller.__init__(self, inst, props, *args, **kwargs)
 
-    def WriteOne(self):
+    def WriteOne(self, axis, value):
         """**IORegister Controller API**. Override if necessary."""
         pass
--- a/src/sardana/tango/pool/IORegister.py
+++ b/src/sardana/tango/pool/IORegister.py
@@ -33,7 +33,7 @@ import sys
 import time
 
 from PyTango import DevFailed, Except
-from PyTango import DevVoid, DevLong
+from PyTango import DevVoid, DevLong, DevDouble
 from PyTango import DevState, AttrQuality
 from PyTango import READ_WRITE, SCALAR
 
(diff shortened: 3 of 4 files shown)


teresanunez:
Hi, I have had a look at this and the issue with float/long is solved. What I see is (but it was already like this) that if I write the attribute Value with a ATK panel open, Value is not updated until I read it from
 any other client. It seems that  there is something wrong with the events. Could you check if you also see something strange there?. Thanks.

reszelaz:
What @teresanunez describes must be related to https://github.com/sardana-org/sardana/issues/21.
I think that the `PoolIORAcquisition` is never started, so no one checks the state neither update the value and its quality. Anyway I would treat it as a separate issue.

Regarding the `Value` type, you went for the double and not for the integer. Do you have in mind any IOR's where integer in not enough?

Finally, I was also looking at the TANGO new attribute type `DevEnum`, whigh sounds interesting from the IOR `Value` attribute point of view. The only thing is that its values must be consecutive integers starting from 0...  And I think that in Sardana we would like to maintain a free choice of the IOR `Value`. But personally, I have never used this new attribute type, is there anyone with more experience?

rhomspuron:
The IOR can be used to control a motor, in this case the the value should be a double, in other words, the DiscretePseudoMotorController could be controller with the IOR. 

The value can be an integer, but in this case its meaning change from value to index like the DevEnum or the case of the TangoAttrIORCtrl, we should decide if the value is an index of a list of values or it is the value which we wants to write in the hardware.

reszelaz:
Yes, we need to agree on the IOR interface. As a baseline analysis we could start from reviewing of what we have already in the [controllers repository](https://sourceforge.net/p/sardana/controllers.git/ci/master/tree/python/ioregister/) and contacting the authors and asking them what is missing in the current interface or how it could be improved. Meanwhile let me start brainstorming...

From the one hand, we could interpret the IOR as a generic element which allows to write/read from a given hardware register a value. This value may be of any type: integer, float, boolean (maybe string as well?) but the hardware usually expects a fixed type for a given register. In the Sardana core, the attribute does not have any type associated to it (at least yet). The type definition that started the whole discussion is used so far by the Tango layer only. Ideally, the IOR `Value` Tango attribute type should not be fixed but dynamic allowing to specify a different type for each IOR axis. This is already possible for the Motor `Position` attribute so I think that we should also allow that for the IOR `Value` attribute. The IOR generic widget could benefit from that providing a specific write/read subwidget.
The controller developer just needs to override the [GetAxisAttributes](https://github.com/sardana-org/sardana/blob/develop/src/sardana/pool/controller.py#L481) method. Some small development would be still necessary to make it work for the IOR. The Motor’s [implementation](https://github.com/sardana-org/sardana/blob/develop/src/sardana/tango/pool/Motor.py#L423) could serve as an inspiration.

So I refine my question, what should be the default type of the IOR `Value` attribute? This has also an impact on the `write_ioreg` macro argument type.

Following this definition of the IOR we should keep in mind that creating an IOR for each register of a hardware controller would mean creating many Pool elements, what in the Tango world also implies many Tango devices. This would allow to control the write process with an action (thread), and having a state associate to each register. An alternative solution would be to use the axis extra attributes of an element. We should document the pros and cons of each of this approaches.

From the other hand it may be interesting to, instead of operating the IOR with the raw values, operate it with the more meaningful value descriptions. At Alba we do it with the [TangoAttributeIORCtrl](https://sourceforge.net/p/sardana/controllers.git/ci/master/tree/python/ioregister/TangoAttrIORCtrl.py). Since we are already discussing the IOR interface maybe it is a good moment to think, if it has sense to include the required logic (calibration and labels) in the generic IOR interface. Again, the IOR generic widget could benefit from that providing a specific write/read subwidgets. My previous comment about the `DevEnun` type is related to that. What are your opinions about it?

Regarding changes of the attribute default type we have made a similar change for the [DiscetePseudoMotorController](https://github.com/sardana-org/sardana/blob/develop/src/sardana/pool/poolcontrollers/DiscretePseudoMotorController.py). Which I think was not necessary. I have reverted 7add752 and it does not provoke any problems with the `mv` macro as stated in #92. @rhomspuron and @saggiss could you cross-check it, please?

reszelaz:
I have submitted https://github.com/rhomspuron/sardana/pull/1 that allows to freely change the IOR `Value` data type between `int`, `float` and `bool`. With this, I think that we could change the default IOR `Value` data type to double as you suggest. The only missing thing would be to change the macro parameter type of `write_ioreg` and the macro result type of the `read_ioreg` from `Integer` to `Double`. This way these macros will compatible with any type of IOR.

Note that it would be trivial to add one more possible data type `str`. Does anyone see it interesting? If there are no voices I will keep just the three types for the moment.

teresanunez:
Up to now we have not had any request for the str type there.

reszelaz:
Let's leave the `str` data type out of the scope of this feature request.

Now, the `read_ioreg` macro result is of type `Float`. But I think that we could refactor this macro in order to follow the `wm` behavior. That means, to print via the door output the IOR value instead of using the magic command output. This way we could print the value respecting its data type e.g. 1, 99.9 or True. But this is of course out of the scope of this PR. Just some idea for the future. 

I think that everything is already implemented so I'm proceeding with the merge.

PR #461, opened 2017-05-08
Solve issue-455: close file on stopping the scan

teresanunez:
Hi Jan, this is only a first try. Could you test if the files are correctly close with this change?.
Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,16 +24,19 @@ This file follows the formats and conventions from [keepachangelog.com]
   in use (SEP6)
 - `synchronizer` parameter to the Measurement Group configuration (SEP6)
 - `latency_time` parameter to the experimental channel controllers (SEP6)
-- `ApplyInterpolation` enviroment variable, applicable to `ascanct` & co.
+- `ApplyInterpolation` environment variable, applicable to `ascanct` & co.
   (SEP6)
 - "How to write a counter/timer controller" documentation (SEP6)
 - "How to write a trigger/gate controller" documentation (SEP6)
 
 ### Changed
-- `ascanct` & co. macro parameters to more resamble parameters of step scans
+- `ascanct` & co. macro parameters to more resemble parameters of step scans
   (SEP6)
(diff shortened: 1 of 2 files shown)


teresanunez:
Jan, the solution is not fine, it can not be like this. It is only to check if the files are correctly closed if
the scan is interrupted (if you don't interrupt it you will get an error).

jkotan:
Hi Teresa,

It works great! 

Best regards,
Jan

teresanunez:
Perfect, thanks for testing.

teresanunez:
I have added endstatus to the ScanHistory.
@reszelaz could you have a look to the implementation and give your opinion?. Thanks.

reszelaz:
Yes, this is exactly how I see it. Since we are simplifying the logic - now we will treat different types of scan terminations equally I think that we can also simplify the `GScan.step_scan` implementation (which BTW I see hard to follow as it is now). I have opened a PR to your branch. Could you take a look? I have just performed a couple of tests, so it is recommended to make some more extensive tries.

Actually when I was refering to adding the end status to the environment, I was just refering the [ScanDataEnvironment](https://github.com/sardana-org/sardana/blob/develop/src/sardana/macroserver/scan/scandata.py#L210), but I don't mind adding it to the `ScanHistory` as well in the global MacroServer environment. It could be usefull there as well.

reszelaz:
When finally integrating this PR it is important to review what are doing the already existing recorders e.g. `OutputRecorder` in the `endRecordList` so it is coherent with the end status.

teresanunez:
Hi Zibi, many thanks for your comments. I have checked the different ouputs and I found an issue concerning the OutputRecorder. If the scan is interrupted the deadtime and the motion dead time are shown as zero, since the data for the calculation is only stored at the end of all the steps (in environment
 variables). This can be corrected if the corresponding environment variables are filled during the scan (with the temporary data summed up). I have tried it and it seems to work fine. Do you see any problem on that?. I could modify the PR adding this.

reszelaz:
If there is no extra time overhead due to this modifications we could change it as you say.

PR #462, opened 2017-05-18
Adapting to four circle diffractometers

teresanunez:
Some macros were not suited for four circle diffractometers. I have adapted them. They are now
 been in use in one of our beamlines with a E4CH diffractometer.

### DIFF
--- a/src/sardana/macroserver/macros/hkl.py
+++ b/src/sardana/macroserver/macros/hkl.py
@@ -1,8 +1,8 @@
 
 """
-    Macro library containning diffractometer related macros for the macros
+    Macro library containning diffractometer related macros for the macros 
     server Tango device server as part of the Sardana project.
-
+    
 """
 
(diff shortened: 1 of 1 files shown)


cpascual:
Hi @teresanunez , it looks like you are the [first victim](https://travis-ci.org/sardana-org/sardana/builds/233526314#L1659) of the [new PEP8 Taliban Robot](#450)...  ;)

teresanunez:
Sorry, I don't know why it has failed now ... is there any automatic way of correcting this?, I am getting crazy.

reszelaz:
The errors are only W291 and W293 (whitespace). It is very easy to spot them in the GitHub PR "changes" tab. These are whitespace changes that are highlighted a little bit more dark than the other. You should eliminate them. `autopep8` would also serve for that.

teresanunez:
I made it now with autopep8 and I put the code even in a worse state ... I think I leave it like this for now, I can not continue with this.

PR #463, opened 2017-05-19
Adapting hkl macros to 4 circle diffractometers

teresanunez:
A new try solving the pep8 problems. Many thanks @reszelaz and @cpascual for your help.


reszelaz:
I won't have time to make the functional tests in the following month. But I have looked at the code and it looks safe to apply now. It should not break anything cause it basically adds conditions to the macros based on the number of motors. So the six-circle diffractometer should work as before.

teresanunez:
Hi @reszelaz , many thanks for looking at that. @cpascual could you perhaps already merge it?

cpascual:
Ok, I merge it trusting  @reszelaz's impresion...  blame him if the macros kill your dog ;)

PR #469, opened 2017-05-28
Support pseudo counters and 0Ds in continuous acquisition e.g. continuous scan.

reszelaz:
First of all, the SEP6 did not fully implement the 0D channels in continuous acquisition (the Tango layer was missing). This was not explicitly said in the SEP6 document, so this PR corrects the document.
 
This PR adds the following features:
- Pseudo counters are supported in continuous acquisition e.g. continuous scans, like `ascanct`. With the exception of pseudo counters that are not based on any physical channel i.e. controller’s `counter_roles` tuple is empty.
- 0D channels are supported in continuous acquisitions. e.g. continuous scans, like `ascanct`.
 
Below I make a summary of changes, grouped by concepts:
 
### Buffering:
- Buffering of the acquired values on the Pool side, necessary to implement the pseudo counters, was moved from the `BufferedAttribute` class to a new, dedicated `SardanaBuffer` class. This class is intended to be aggregated by the experimental channels, similarly as instances of the `SardanaAttributes` e.g. `Value`, `CurrentValue`. In my opinion this lead to a better separation of concerns between these two concepts: buffer and attribute.
- Value buffers of the physical and pseudo channels, similarly as value attributes are connected via sardana events, with the following difference. The value attribute event does not execute the calculation yet, but the access the pseudo value triggers the calculation. In case of the value buffer, the calculation is executed when the physical buffer is filled. This has one drawback - at the moment, the pseudo counters that are not included in the acquisitions are calculated as well. The idea is to make the pseudo counters parts of the acquisition action (totally passive), so we could calculate only the pseudo counters which indeed participate in the acquisition.
- Buffers are continuously cleaned from physical values that already participated in all pseudo buffers calculations. However in some special cases the very last chunks may not be cleaned. This resting values in the buffers should be cleaned in the acquisition actions’ finish hook, but it is not possible to implement right now - the actions are independent and may finish in an aleatory order. As a workaround we perform the final cleanup at the next acquisition start.
 
### 0D:
- A part of the core API was changed in order to be more consistent with the new concept of value buffer.
    - `BaseAccumulation.append_value` -> `BaseAccumulation.append`
    - `Value.get_value_buffer` -> `Value.get_accumulation_buffer`
    - `Value.append_value` -> `Value.append_buffer`
    - `PoolZeoDExpChannel.get_value_buffer` -> `PoolZeoDExpChannel.get_accumulation_buffer`
    - `PoolZeoDExpChannel.value_buffer` -> `PoolZeoDExpChannel.accumulation_buffer`

For the moment I did not implement the backwards compatibility layer for them. IMHO nobody is using these core classes outside of Sardana. If requested we could think of a compatibility layer.
- OD `ValueBuffer` Tango attribute is marked as deprecated in favor of `AccumulationBuffer` attribute.
 
### Others:
- Base classes `PoolExpChannelDevice` and `PoolExpChannelDeviceClass` were implemented. They are intended to be the base classes for all the experimental channels.
 

Actions to be done in the future:
* Before the official release we should make the deprecation warnings more explicit (version number, and expiration date)
* When 0D’s `ValueBuffer` deprecation expires, we should rename all the experimental channel `Data` attributes to `ValueBuffer` attributes.
 
The development/commit history may seems a little bit chaotic, sorry about that.

@sardana-org/integrators please take a look on this PR, it is somehow blocking integration of #457, which should be rewritten and benefit from the `PoolExpChannelDevice` base class.


teresanunez:
Hi @reszelaz, I have tried this code and I get the error:

 File "/home/tnunez/sardana_zibi_contscan_pc/src/sardana/macroserver/scan/gscan.py", line 1955, in _go_through_waypoints
    waypoint_info = self.prepare_waypoint(waypoint, start_positions)
  File "/home/tnunez/sardana_zibi_contscan_pc/src/sardana/macroserver/scan/gscan.py", line 1900, in prepare_waypoint
    interval_displacement = total_displacement / self.macro.nr_interv
  File "/home/tnunez/sardana_zibi_contscan_pc/src/sardana/macroserver/macro.py", line 2402, in __getattr__
    (type(self).__name__, name))
AttributeError: 'ascanct' object has no attribute 'nr_interv'

I have taken today the code from the branch contscan_pc of your fork.

reszelaz:
That's strange. It fails in a place where it does not have too much sense. It looks like the macro that you run, the `ascanct` is from some previous revision of sardana. There was a change before accepting the SEP6 where we have changed the `nr_of_points` to `nr_interv` to be as much as possible equivalent to the step scan `ascan`.

I have just tested it and in my case it worked. I have used the measurement group with one counter, one 0D and a pseduo counter based on those two. But I think that the error that you observe is not related to the measurement group at all.

teresanunez:
Sorry, you are right. I forgot to change the Macropath property to the new one.

teresanunez:
Hi @reszelaz, I have tested this again,  now with the right macros and using zeroD and pseudocounters in the continuous acquisition. I think it is fine, I have not found problems. Do you want me to integrate it?.

reszelaz:
Thanks for the review! Yes, go ahead with the merge. I will later on update the channgelog.

reszelaz:
The changelog is now updated.

Issue #452, opened 2017-04-13
Improve travis-ci output readability when flake8 check fails


reszelaz:
Travis always executes 2 main checks - flake8 and sardanatestsuite. It is difficult to figure out which of these two was the reason of failure. One of the possibilities of the improvement would be to use the configuration matix as explained in this https://github.com/sardana-org/sardana/pull/450#issuecomment-293496858.

cpascual:
Seeing how frustrating the experience was for a seasoned developer such as @teresanunez in #462, I think we should try to implement this kind of urgently (or risk losing some contributions).

Another related improvement, could be to make the flake8_diff.sh script to echo the following:

- a suggested autopep8 command that would fix the issues 
- the command that the user can run locally to check before pushing 

or alternatively,...
- write a generic version of the above instructions in the CONTRIBUTING.md file and let the flake8_diff.sh point to it.

teresanunez:
Many thanks @cpascual, I think these are the main points ... Once you got the error after your first attempt (in case, you were not aware of checking your code locally before making the PR),
it shoud be pointed out how the code could be checked before pushing again and the output of this check should be a command for fixing the issues. I think this would help a lot. I would show the information compact and clear (for the specific case) in the PR itself.

reszelaz:
First of all sorry about the inconveniences. I totally agree on the relevance of this issue. But this week I won't have time to improve it, I could start from the next weekend.

cpascual:
Fixed by #470

PR #470, opened 2017-05-29
Improve flake8 output in travis-ci (#452)

reszelaz:
This PR aims to improve the output of the flake8 check in travis-ci - #452.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,27 +9,31 @@ python:
   - "2.7"
 
 env:
-  # run flake8 check on all python files in the project
-  - TEST_CMD="ci/flake8_diff.sh"
-  # run the full testsuite
-  - TEST_CMD="docker exec sardana-test sardanatestsuite"
+  - TEST="flake8"
+  - TEST="testsuite"
 
 before_install:
-  - sudo apt-get update -qq
   # install flake8 to perform python code style check in the script part
-  - sudo apt-get install -qq python-flake8
+  - if [ $TEST == "flake8" ]; then sudo apt-get update -qq ; fi
+  - if [ $TEST == "flake8" ]; then sudo apt-get install -qq python-flake8; fi
 
 install:
   # run reszelaz/sardana-test docker container (Debian8 with sardana-deps)
-  - docker pull reszelaz/sardana-test
-  - docker run -d --name=sardana-test -h sardana-test --volume=`pwd`:/sardana reszelaz/sardana-test
+  - if [ $TEST == "testsuite" ]; then docker pull reszelaz/sardana-test; fi
+  - if [ $TEST == "testsuite" ]; then docker run -d --name=sardana-test -h sardana-test --volume=`pwd`:/sardana reszelaz/sardana-test; fi
+
   # wait approx. 10 s (supervisor starts mysql and Tango DB)
-  - sleep 10
+  - if [ $TEST == "testsuite" ]; then sleep 10; fi
+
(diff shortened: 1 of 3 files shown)


cpascual:
The commit c592f92 seems ok to me.
Is this still WIP? (the title says so)

Are you planing to also improve the output as per https://github.com/sardana-org/sardana/issues/452#issuecomment-302637286 ?

reszelaz:
Yes, I will either improve the output or the documentation.
I will try to have it for tomorrow.

reszelaz:
Now it is ready for review and eventual integration.

cpascual:
With the latest commit (d478876) you introduced a bug: the script returns 0 even if flake8 found errors

cpascual:
I am fixing it myself and merging

.travis.yml
```diff
@@ -8,6 +8,12 @@ services:
 python:
   - "2.7"
 
+env:
+  # run flake8 check on all python files in the project
+  - TEST_CMD="ci/flake8_diff.sh"
+  # run the full testsuite
+  - TEST_CMD="docker exec sardana-test sardanatestsuite"
+
```
cpascual: I just realised that this implementation makes the flake8 test also run a lot of the `install` part that is useless for it. This has 2 problems:
- It wastes time (the flake8 test could be much faster)
- It mixes cases for failure: e.g., if a commit introduces problems that make  `python setup.py install` to fail, the flake8 test would report an error that is actually due to the unrelated install issue

To fix this, we could use [stages](https://docs.travis-ci.com/user/build-stages) but maybe it is simpler to just do the install steps conditionally, as in the following example::

```yaml
(...)
env:
  - TEST_CMD="ci/flake8_diff.sh" INSTALL="N"
  - TEST_CMD="docker exec sardana-test sardanatestsuite"  INSTALL="Y"

(...)

install:
  - if [ $INSTALL == "Y" ]; then docker pull reszelaz/sardana-test ; fi
  - if [ $INSTALL == "Y" ]; then docker run -d --name=sardana-test -h sardana-test --volume=`pwd`:/sardana reszelaz/sardana-test ; fi
  - (...)

(...)


```

reszelaz: Using stages would mean changing the approach completely,  since they run sequentially and  a stage failure makes the whole pipe to break. That would mean loosing information in case of a failure - we will only learn about the state of one of the checks.

Apart of the option that you suggest (conditional install) we could also revert just the commit c592f92, accepting the two others and run one job only. 

I will investigate a little bit more on that..

cpascual: > we could also revert just the commit c592f92, accepting the two others and run one job only.

IMHO that would be a pity. I would go for conditional install

reszelaz:
I have added conditional tests preparations and it improved the jobs durations. Now the `flake8` takes only 37 s.

The travis configuration file may be hard to follow now. In order to improve it, we could move the whole `testsuite` part to a script (both `install` and `script` sections commands), let's call it `testsuite_docker.sh`. With that, we could just run `falke8_diff.sh` or `testsuite_docker.sh` in the `script` section depending on the env.
 
This way we will gain some readability in the travis configuration file but in the jobs output all theses commands, that now are separate, will be treated as one. Now each of the commands of the `install` section is by default folded, so its output is hidden until someone unfold it. Moving it to a script we will loose this making the output less readable. 

At the end, I think that the less readable configuration file is less harmful than the less readable jobs output - the one that the users read. So I would leave it as it is. But I would appreciate your opinions.

Issue #471, opened 2017-05-31
Provide data for 'ct' macro


droldan:
Our users often wants to acquire in a macro in the middle procedure, like before/after movements or before/after scan.
In order to resolve this we would like to provide data in 'ct' macro to allow getData() method provided by the macro object.

Expected (in a macro):

```
ct = self.execMacro('ct 2')
data =  ct.getData()
```

reszelaz:
Resolved by #682

PR #472, opened 2017-05-31
Provide data to 'ct' macro

droldan:
Proposal for Issue #471

'ct' macro does not allow data access.
Add access to data via getData() in 'ct' macro.

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -676,7 +676,13 @@ class ct(Macro):
         for line in table.genOutput():
             self.output(line)
 
-
+        # Prepare data result to be extracted by getData()
+        self._data = {}
+        self._data['mntGrp'] = self.mnt_grp_name
+        self._data['integ_time'] = integ_time
+        self._data['data'] = zip(names, counts)
+        self.debug("Data: %r ", self._data)
+        
 class uct(Macro):
     """Count on the active measurement group and update"""
 
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -678,7 +678,7 @@ class ct(Macro):
 
         # Prepare data result to be extracted by getData()
         self._data = {}
-        self._data['mntGrp'] = self.mnt_grp.name
+        self._data['mnt_grp'] = self.mnt_grp.name
         self._data['integ_time'] = integ_time
         self._data['data'] = zip(names, counts)
         self.debug("Data: %r ", self._data)
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -678,7 +678,7 @@ class ct(Macro):
 
         # Prepare data result to be extracted by getData()
         self._data = {}
-        self._data['mntGrp'] = self.mnt_grp
+        self._data['mntGrp'] = self.mnt_grp.name
         self._data['integ_time'] = integ_time
         self._data['data'] = zip(names, counts)
         self.debug("Data: %r ", self._data)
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi Daniel,
                  I have test it and I get the error:

'ct' object has no attribute 'mnt_grp_name'

The name is:

mnt_grp

If you change it:
        self._data['mntGrp'] = self.mnt_grp_name

to:
        self._data['mntGrp'] = self.mnt_grp

it works.

droldan:
Thanks Teresa, ctrl+c ctrl+v mistake...
I have pushed the fix, added the mntGrp name instead the object.

src/sardana/macroserver/macros/standard.py
```diff
@@ -676,6 +676,12 @@ def run(self, integ_time):
         for line in table.genOutput():
             self.output(line)
 
+        # Prepare data result to be extracted by getData()
+        self._data = {}
+        self._data['mntGrp'] = self.mnt_grp.name
+        self._data['integ_time'] = integ_time
```
reszelaz: I think it is better to use the same type of key naming, either CamelCase or underscore, but always the same. I prefer underscore. So here we could use 'mnt_grp'. What do you think?

src/sardana/macroserver/macros/standard.py
```diff
@@ -676,6 +676,12 @@ def run(self, integ_time):
...
+        self._data['data'] = zip(names, counts)
+        self.debug("Data: %r ", self._data)
```
reszelaz: Do we really need this debug? The results are already printed with the output.

src/sardana/macroserver/macros/standard.py
droldan: Thanks @reszelaz, off course! Pushed the changes.

src/sardana/macroserver/macros/standard.py
droldan: Pushed the changes, deleted debug line.

reszelaz:
I would start with a simpler data definition and if we have a strong argument to extend it we could discuss it in another issue. For the moment I propose to not include in the data definition neither the measurement group name nor the integration time. I think that the first one is not necessary because we have already the names of the involved channels. The second one is not necessary because the macro’s data is usually accessed from a wrapper macro that creates the `ct` macro object and runs it. In this way, the wrapper macro knows the set integration time - a parameter of the `ct` macro. Furthermore, the set integration time should not be used for the data post-processing but the real integration time should be used instead (coming from the timer channels) because they may be different. So we could start with just the pairs of the channels’ references and their acquired data.

This PR uses the channels labels as data references e.g. “ct01” (internal channel) or “sys/tg_test/1/short_scalar” (external channel). When assessing if this is the correct way to go I have cross checked how it is done in the scan framework and I found some inconsistencies. In the case of the scans the moveables use names e.g. “mot01”, the internal channels use the full names without the scheme part e.g. “pc255:10000/expchan/ctctrl01/1” and the external channels use the proper URIs e.g. “tango://pc255:10000/sys/tg_test/1/short_scalar”. I think that this is a bug and we should agree and unify the data references in the scan data. Following this decision, I would use the same for the `ct` data. 

Knowing that a scan may involve motors coming from different pools, defined in different databases, I think that we need the proper URIs as references. I will create a separate issue that will address this so we could discuss on how to implement this change and its backward compatibility policy.

If my proposal regarding the data references in scans gets accepted I would also use the proper URI’s in the `ct` data. In both cases, scans and `ct`, the use of the proper URIs may introduce extra difficulties to the end user when manipulating data. The users usually do not know the full names of the Sardana elements so I propose that we develop some utilities/recipes to ease the data manipulation.

reszelaz:
Since this PR still requires some agreements and modifications I propose to move it out if the Jul17 milestone.

reszelaz:
@droldan thanks for this proposal. We will provide this feature when #630 will get integrated.

Issue #473, opened 2017-05-31
Allow measurementGroup as a parameter in 'ct' macro.


droldan:
Following the [Issue-471](https://github.com/sardana-org/sardana/issues/471), we want to allow the measurement group as a optional parameter in 'ct' macro to enable acquire in a specific measurement group.
This will help to avoid change the 'ActiveMntGrp' environment variable in each ct measurement and make it more flexible

reszelaz:
Solved with #940.

PR #474, opened 2017-05-31
Provide MeasurementGrp as a parameter in ct macro

droldan:
Proposal for Issue #473 
'ct' macro acquire the ActiveMntGrp, it would be good to pass it
via optional parameter.
Add measurement group as a optional parameter.
Fully backward compatible

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -640,19 +640,19 @@ class ct(Macro):
     env = ('ActiveMntGrp',)
 
     param_def = [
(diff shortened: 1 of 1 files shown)


reszelaz:
Thanks for this proposal! I agree that it would be interesting to pass a MeasurementGroup to the `ct` macro. But I don't like to pass it as a String parameter and would prefer to pass it as a MeasurementGroup parameter. This way we could benefit from the context tab completion in spock and from the parameter validation done in the core. In order to properly implement it as an optional MeasurementGroup we would need to implement #285 first and I propose to do it in this order. So I'm rejecting this PR because after implementing #285 the implementation will be different and much simpler.

PR #468, opened 2017-05-25
Fix #467 - catching and handling exception in a motion ending in alarm

LJBD:
This is a very simple fix of #467 and a PEP8-ification of the module the error was in.
Also, when I've printed that value of `state` variable that caused the exception here, I got what I've reported in #456 , so:
```
["TypeError: 'TaurusDevState' object is not callable\n"]
```
But I guess that that is another issue. Only now have I learned why I've seen it and not the exception: it turns out that @daneos has hot-fixed it.

PR #476, opened 2017-06-02
Adapt state to taurus4 in hkl macros

teresanunez:


reszelaz:
This are exactly the same changes that in #468 so I merged without functional tests. Thanks!

PR #478, opened 2017-06-08
[WIP] SEP17: Ongoing acquisition formalization and implementation

reszelaz:
This Pull Request aims to introduce the SEP17 (see its main text [here](https://github.com/reszelaz/sardana/blob/27be5bd852325fa61216c54077b1bd0f1a9fb438/doc/source/sep/SEP17.md)).

    Title: Ongoing acquisition formalization and implementation
    Abstract:
     This SEP aims to formalize the ongoing acquisition of experimental
     channels and implements the missing aspects of this feature.

It presents three options and we should still choose one of them and then implement the necessary changes. I have elaborated the most the first option cause it has the most sense to me, but I'm also opened for the rest two or any other. I would appreciate any feedback.

### DIFF
--- a/doc/source/sep/SEP17.md
+++ b/doc/source/sep/SEP17.md
@@ -6,12 +6,12 @@
     URL: http://www.sardana-controls.org/sep/?SEP17.md
     License: http://www.jclark.com/xml/copying.txt
     Abstract:
-     This SEP aims to formalize the ongoing acqiuisition of experimental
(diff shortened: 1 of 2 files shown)


reszelaz:
When discussing this SEP we should take into account what was reported by @jkotan in https://github.com/sardana-org/sardana/issues/662#issuecomment-349417197.

PR #479, opened 2017-06-14
Accept additional args in flake8_diff.sh

cpascual:
The flake8_diff.sh script is a custom wrapper on flake8 used for
CI, but also locally for performing checks before pushing.
Add support for passing arguments to flake8_diff.sh so that it
propagates them to flake8 (useful e.g. if your local version of
flake8 behaves differently to that in the CI machine).


reszelaz:
It will be useful, thanks!

PR #480, opened 2017-06-16
Complete Spock Logging (issue #102)

### LINKED ISSUE #102: Complete Spock Logging (SF#161)
Hello,
      some of our users want to have a file with the complete output from spock after each session (not only the history with the commands that have been sent).
I would like to start the discussion about the possibility of implementing this
 logging feature and to request this implementation.
Many thanks.

                             Regards,
                                             Teresa  

Original Ticket: [sardana/tickets/161](https://sourceforge.net/p/sardana/tickets/161)

teresanunez:
Fix #102 
This PR has to be evaluated using the PR taurus-org/taurus#469


cpascual:
Hi @teresanunez, On your first commit you had [both pep8 problems and a testsuite error](https://travis-ci.org/sardana-org/sardana/builds/243647383?utm_source=github_status&utm_medium=notification). After fixing the pep8 problems you still have the [testsuite error](https://travis-ci.org/sardana-org/sardana/jobs/244410961#L3736):

```
======================================================================

FAIL: test_read_ioreg_macro_runs (sardana.macroserver.macros.test.test_ioregister.IORegisterTest)

Testing read_ioreg with macro_runs(macro_params=['ior02'], wait_timeout=1, macro_name='read_ioreg')

----------------------------------------------------------------------

  File "/usr/local/lib/python2.7/dist-packages/taurus/test/base.py", line 106, in newTest

    return helper(**helper_kwargs)

  File "/usr/local/lib/python2.7/dist-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/macros/test/base.py", line 216, in macro_runs

    self.assertFinished('Macro %s did not finish' % macro_name)

  File "/usr/local/lib/python2.7/dist-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/macros/test/base.py", line 186, in assertFinished

    self.assertIn(state, finishStates, msg)

AssertionError: Macro read_ioreg did not finish; State history=[u'start', u'finish', u'exception']

----------------------------------------------------------------------
```

Could this be related to the fact that it needs taurus-org/taurus#469 to be tested?

teresanunez:
I don't think so ... in Taurus I have only changed something in the Logger (added a function) and the
 error talks about a macro for the ioregister ... But this part is exactly like in develop ...

cpascual:
Can you try to run the testsuite locally and see if it is ok?

Then you could use the specific test for debugging:

```
python -m unittest sardana.macroserver.macros.test.test_ioregister.IORegisterTest
```

teresanunez:
Hi @cpascual, testsuite is not ok, but I get also errors taking the develop branch and unittest
 is also not ok.

teresanunez:
Hi @cpascual, I solved the problem the problem with testsuite, I found it now running
 the tests separately. Thanks for the help.

cmft:
Hi all,
I propose to implement this feature using a logging filehandler (It seems more simple and does not requires changes in Taurus).

I did a quick test (base on @teresanunez  code). What do you think?
``` 
import time
from sardana.macroserver.macro import *
import logging

class MyMacro(Macro):
    def __init__(self, *args, **kwargs):
        Macro.__init__(self, *args, **kwargs)
        # Add file handler
        try:
            self.logging_onoff = self.getEnv("LogMacroOnOff")
        except:
            self.logging_onoff = False

        if self.logging_onoff:
            self.output("LOGGING ON")
            log_file = "/tmp/spock_output.log"
            self.fileHandler = logging.FileHandler(log_file)
            self.fileHandler.setFormatter(self.log_format)
            self.logger = self.getLogger()
            self.logger.addHandler(self.fileHandler)

    def __del__(self):
      self.logger.removeHandler(self.fileHandler) 
```

teresanunez:
Hi @cmft, thanks for looking at this. I have to check it. I don't know now if the fileHandler is already
 defined or not and what is written ... I will try to look at this tomorrow.

teresanunez:
Hi @cmft, I have checked this. I integrated in the class Macro, using an environment variable for activating it. It works fine, but I am not sure if this is how the users want to have the file. I will discuss with them. So we can postpone this until I get an answer. Many thanks. I will also write now something in the Taurus PR.

reszelaz:
I was trying what @cmft proposed and I think it will not work well with the hooks or macros called from within other macros. At least this is what I've observed:
1. The messages logged from the _inner_ macros e.g. hooks are logged twice.
2. The messages logged by the MacroExecutor object also appears in the file:

```
SardanaTP.W004 DEBUG    2017-07-10 16:11:08,741 Door/zreszela/1.MacroExecutor: [ END ] runMacro Macro 'MyMacroOut() -> 1'
SardanaTP.W004 DEBUG    2017-07-10 16:11:08,741 Door/zreszela/1.MacroExecutor: Job ended (stopped=False, aborted=False)
```

Nevertheless, I think that the solution with a FileHandler instead of [extending the taurus logging system](https://github.com/taurus-org/taurus/pull/469) is the way to go. Similarly, as it is done for the "Report" feature of the MacroServer. 

Having in mind that this PR still requires some work I propose to move it out of the Jul17 milestone and continue developing this feature without rush.

teresanunez:
@reszelaz I have observed the two points you mentioned, and I don't like them, I do not think that
 our users would accept it, but they have still not have time to test it. The output in the 2 point can be modified with settings in the Logging, but in any case ... No problem if you move it out from the Jul17 milestone.

reszelaz:
> The output in the 2 point can be modified with settings in the Logging

What settings are these? Is it possible to provide a short snippet here to demonstrate it? Or you have it in some branch already implemented. I was just reviewing what is in the current branch of this PR and the example from https://github.com/sardana-org/sardana/pull/480#issuecomment-309672179 and I couldn't find it.

teresanunez:
Hi @reszelaz, I mean settings for changing the format of the output, for example with:

formatter = logging.Formatter("%(asctime)s %(message)s")
self.fileHandler.setFormatter(formatter)

one gets an outupt like:

2017-07-11 12:19:34,211    0          1         1      0.0446622     1         0      0.0172672
2017-07-11 12:19:34,211    0          1         1      0.0446622     1         0      0.0172672
2017-07-11 12:19:34,211 [START] runMacro Macro 'gc_macro() -> -8'
2017-07-11 12:19:34,211 [START] runMacro Macro 'gc_macro() -> -8'

Perhaps it is possible to define an own file handler based in the python logging where one could
 put somehow a filter to the messages of the MacroExecutor. In any case I think I prefer to implement
something from scratch.

teresanunez:
Hi @reszelaz, I have checked if it would be possible to add some customized filter to the Logging
 FileHandler and yes, it is possible ... one can take out the messages from the MacroExecutor or
 avoid the duplication of messages ... customized the format ... it is really flexible. I think we could
 go for this option. What do you think?

reszelaz:
@teresanunez sorry for the late reply!
I think that in order to get rid of the duplicated logs in case of the nested macros we should consider [this](https://docs.python.org/2/library/logging.html#logging.Logger.propagate) recommendation. It says that the handler should be added to the higher logger instance only, in our case the top most macro. In this case we should look for a proper place in the [MacroExecutor](https://github.com/sardana-org/sardana/blob/develop/src/sardana/macroserver/msmacromanager.py#L816) class.

teresanunez:
HI @reszelaz, many thanks for the information. I will have a look.

teresanunez:
Hi @reszelaz, I have now added the filehandler to the MacroExecutor, as you suggested, and the messages are not repeated any more, nice, thanks. Without setting anything extra, simply adding the handler I get something like this:

[START] runMacro Macro 'lsm(['.*']) -> c39744a6-6bb0-11e7-9b5b-901b0e40a03c'
        Name    Type       Controller   Axis
 ----------- ------- ---------------- ------
   exp_dmy01   Motor   dummy_mot_ctrl      1
   exp_dmy02   Motor   dummy_mot_ctrl      2
   exp_dmy03   Motor   dummy_mot_ctrl      3
    mnchrmtr   Motor     omsvme58_exp      8
     motoms1   Motor     omsvme58_exp      1
     motoms2   Motor     omsvme58_exp      2
     motoms3   Motor     omsvme58_exp      3
     motoms4   Motor     omsvme58_exp      4
     motoms5   Motor     omsvme58_exp      5
     motoms6   Motor     omsvme58_exp      6
[ END ] runMacro Macro 'lsm(['.*']) -> c39744a6-6bb0-11e7-9b5b-901b0e40a03c'
Job ended (stopped=False, aborted=False)
Configuration changed
ExtraColumns is not defined
SharedMemory is not defined. Use "senv SharedMemory sps" to enable it
[START] runMacro Macro 'ascan(exp_dmy01, 1.0, 10.0, 3, 1.0) -> c8c6ef1c-6bb0-11e7-9b5b-901b0e40a03c'
Operation will be saved in /tmp/test_[ScanId].lima (Lima)
Scan #1818 started at Tue Jul 18 14:01:08 2017. It will take at least 0:00:04.000392
Moving to start positions...
 #Pt No   exp_dmy01    t01       c03       c02        dt   
[START] motion
[ END ] motion
[START] acquisition
[ END ] acquisition
   0          1         1      0.0395526  50.4819   0.344789
[START] runMacro Macro 'gc_macro() -> -1'
[ END ] runMacro Macro 'gc_macro() -> -1'
[START] motion
[ END ] motion
[START] acquisition
[ END ] acquisition
   1          4         1      139.424   111.057   1.61898 
[START] runMacro Macro 'gc_macro() -> -1'
[ END ] runMacro Macro 'gc_macro() -> -1'
[START] motion
[ END ] motion
[START] acquisition
[ END ] acquisition
   2          7         1      26.9498   1.01381   2.84508 
[START] runMacro Macro 'gc_macro() -> -1'
[ END ] runMacro Macro 'gc_macro() -> -1'
[START] motion
[ END ] motion
[START] acquisition
[ END ] acquisition
   3          10        1      0.0621481  174.425   4.11347
[START] runMacro Macro 'gc_macro() -> -1'
[ END ] runMacro Macro 'gc_macro() -> -1'
Operation saved in /tmp/test_01818.lima (Lima)
Scan #1818 ended at Tue Jul 18 14:01:14 2017, taking 0:00:05.352674.Dead time 25.3% (motion dead time 0.6%)
[ END ] runMacro Macro 'ascan(exp_dmy01, 1.0, 10.0, 3, 1.0) -> c8c6ef1c-6bb0-11e7-9b5b-901b0e40a03c'
Job ended (stopped=False, aborted=False)


I guess one could now play with formats/filters ... Should we go for that?.

                            Regards,
                                             Teresa

teresanunez:
A first disadvantage that I see is that one can not control the turn on/off of the logger from the spock
environment, since it is set one the macroserver is started, and not when a running macro is accessible.
At least I have not found the way to set it somewhere else without duplicating messages.

reszelaz:
Actually, I was thinking to still attach the file handler to the macro object but do it somewhere in the `MacroExecutor` scope, for example in the `runMacro` method. In order to avoid multiple logs when macros are nested or hooks are used, the FileHandler should be attached only to the top-most macro. This could be achieved by consulting the `getParentMacro` method of the Macro object, if it is None, that means that it is the top-most macro.

BTW, as an alternative to using the environment variables to control this feature, we could use the Door’s attributes. But this way we will lose the possibility to enable logging for just a subset of macros i.e. environment variable set on the macro level, and I think this is interesting.

teresanunez:
HI @reszelaz, it could well be that I am doing something wrong ... but the solution you propose doesn't work for me. In fact I had already tried, every time a macro is called (parent macros) the number of
 messages in the log file is increased by one ...

reszelaz:
Could you please push the code to github so I could give it a try locally.

teresanunez:
@reszelaz, I solved now the problem, removing the handler a the end. It works fine now.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1243,6 +1245,8 @@ def run(self, params, asynch=True):
             # return self._macro_pointer.getResult()
 
     def _jobEnded(self, *args, **kw):
+        if self.logging_onoff:
```
reszelaz: If we remove the handler in `_jobEnded` that mean that in case of synchronous executtion the logger would not be executed. I would move the removal at the end of `runMacro` method.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1294,31 @@ def __runXMLMacro(self, xml):
     _runXMLMacro = __runXMLMacro
 
     def runMacro(self, macro_obj):
+
         name = macro_obj._getName()
         desc = macro_obj._getDescription()
         door = self.door
+        self.macro_obj = macro_obj
```
reszelaz: I think this is dangerous. We can not maintain a reference to the macro_obj, otherwise the memory could not be freed after the macro execution. The macro_pointer reference and the environment variable `PreserveMacroObj` is related to that.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1294,31 @@ def __runXMLMacro(self, xml):
...
+            try:
+                self.logging_onoff = macro_obj.getEnv("LogMacroOnOff")
+            except:
+                self.logging_onoff = False
+            if self.logging_onoff:
+                try:
+                    self.logging_path = macro_obj.getEnv("LogMacroPath")
+                except:
+                    macro_obj.setEnv("LogMacroPath", "/tmp")
+                    self.logging_path = macro_obj.getEnv("LogMacroPath")
```
reszelaz: Maybe we could go for variables like: `LogMacroDir` and `LogMacroFile`, similarly to the `ScanDir` and `ScanFile`.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1294,31 @@ def __runXMLMacro(self, xml):
...
+                log_file = self.logging_path + "/spock_session.log"
```
reszelaz: I would not hard code the file name. Maybe `ScanFile` variable could be used?

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1294,31 @@ def __runXMLMacro(self, xml):
...
+                self.macro_obj.fileHandler = logging.FileHandler(log_file)
```
reszelaz: I think that we don't need these *monkey patching* (`fileHandler` and `logger`). Actually it would be enough to keep the `fileHandler` reference locally if we remove the handler in `runMacro` as well (see my previous comment).

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1294,31 @@ def __runXMLMacro(self, xml):
...
+                formatter = logging.Formatter(
+                    "%(levelname)-8s %(asctime)s %(name)s: %(message)s")
```
reszelaz: What do you think to allow the user to change the formatter? Maybe another environment variable?

reszelaz:
It looks very promising and I think that this is the way to go! I have just added some review comments.

The next step would be to look into the *filters*. This could be used to convert/remove the `MacroExecutor` logs and maybe the Taurus deprecation warnings. But only if the users are currently interested in this.

Regarding the environment variable names I would use names starting with underscore, just to indicate their private scope. 

And as with other new macros let's start voting for the names:
- currently we have `logmacro_on` and `logmacro_off`
- I'm in favor of one macro with parameter e.g. `logmacro on|off` or `logmacro True|False`
- another option could be `history` or `script` with parameters..

Any other proposals?

src/sardana/macroserver/msmacromanager.py
teresanunez: Done

src/sardana/macroserver/msmacromanager.py
teresanunez: Not needed any more (it was only used in the call done in jobEnded, now moved to runMacro). I remove it.

src/sardana/macroserver/msmacromanager.py
teresanunez: No problem, I will change it.

src/sardana/macroserver/msmacromanager.py
teresanunez: Our users wanted hard coded names. In the last version they have we had to append to the logfile name the name of the node where spock was running. They wanted like this. If you agree, I could put the hardcoded name in case LogMacroFile is not defined. Is this fine for you?

src/sardana/macroserver/msmacromanager.py
teresanunez: In fact, this is there because it was needed in my previous implementation (without using the Handler) and I did not clean the files ... waiting for what would be decided to be used. I can remove it now.

src/sardana/macroserver/msmacromanager.py
teresanunez: It is fine for me. I can leave this as default in case the environment variable is not defined.

teresanunez:
@reszelaz many thanks for the review. I have worked on your comments. The environment variables are now without underscore, but I also think that it could be added if we want the users to use only the macros. The name of the macros in your second point are fine for me.

src/sardana/macroserver/msmacromanager.py
reszelaz: I agree.

src/sardana/macroserver/msmacromanager.py
reszelaz: Ok for me.

src/sardana/macroserver/msmacromanager.py
reszelaz: I think you can go ahead with this as well.

reszelaz:
Thanks @teresanunez, all your changes seems fine to me.

So just as a summary, this feature would add the following environment variables:
- `LogMacroOnOff` (If we go for macro `logmacro [True|False]` or `logmacro [on|off]`I would shorten this variable to `LogMacro`)
- `LogMacroDir`
- `LogMacroFile`
- `LogMacroMode`
- `LogMacroFormatter` (Personally I think that `LogMacroFormat` is more accurate since this varible will store the format of the message not the formatter itself)

Regarding the proposal of converting the `logmacro_on` and `logmacro_off` into one macro with parameters. I do not have any strong opinion about the parameter type: `Boolean` vs. `String`. The bool option gives us a benefit of type validation implemented in the core and easier to implement spock contextual `Tab` completion. Also the macroexecution widgets could populate the parameter editor with the True & False options. The string option would be more similar to what we have now for the `debug` magic command.

And the last thing, @teresanunez do you think that the message filtering part is still missing?

teresanunez:
Hi @reszelaz, thanks for your comments. I can change the LogMacroOnOff to LogMacro, no problem, I will do. About Boolean or String ... well, we could go for Boolean.
The filtering ... without filtering there are so many unuseful things in the file (from the user point of view) that I am not sure that our users will accept it. But we could try and see, what do you think?.

reszelaz:
Great! So go ahead with the macro name change, please. Are you also fine with shorening the `LogMacroFormatter`?

I also have an impression that the filtering will be demanded by the users. The thing is that I won't have time in the short term to implement it. If you are willing to try I could later make the code review and tests.

teresanunez:
Yes, I will also make the change of LogMacroFormatter to LogMacroFormat.
I will have a look at the filtering.

teresanunez:
LogMacroMode, for selecting if the log file has to be updated or a new one has to be created, does not
work since using the FileHandler, I will also look at that.

teresanunez:
@reszelaz, I have tested the change of the log messages format setting it in the LogMacroFormat 
environment variable and it does not work, the format is not understood. I have try several ways, like:

senv LogMacroFormat "%(levelname)-8s %(asctime)s %(name)s: %(message)s"
senv LogMacroFormat "\"%(levelname)-8s %(asctime)s %(name)s: %(message)s\""

But in all cases I get in the log file the string that I set, I mean:

%(levelname)-8s %(asctime)s %(name)s: %(message)s

and not the real log messages any more. Do you already see what I am doing wrong?

teresanunez:
Sorry, I see it now. There is an error in the code.

teresanunez:
Hi @reszelaz, I have implemented a default filter converting a default output like:

2018-01-05 11:40:03,063 Door/issue102/1.MacroExecutor: [START] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c90c8470-f204-11e7-bc82-901b0e40a03c'
2018-01-05 11:40:03,063 Door/issue102/1.Macro[ascan]: Scan #42 started at Fri Jan  5 11:40:03 2018. It will take at least 0:00:02.048528
2018-01-05 11:40:03,063 Door/issue102/1.Macro[ascan]:  #Pt No    mot17      ct17      ct18      ct19      ct20       dt   
2018-01-05 11:40:03,063 SScan: [START] motion
2018-01-05 11:40:03,925 SScan: [ END ] motion
2018-01-05 11:40:03,926 SScan: [START] acquisition
2018-01-05 11:40:03,933 SScan: [ END ] acquisition
2018-01-05 11:40:03,933 Door/issue102/1.Macro[ascan]:    0         1         0         0         0         0      0.862778
2018-01-05 11:40:03,934 SScan: [START] motion
2018-01-05 11:40:04,554 SScan: [ END ] motion
2018-01-05 11:40:04,554 SScan: [START] acquisition
2018-01-05 11:40:04,570 SScan: [ END ] acquisition
2018-01-05 11:40:04,571 Door/issue102/1.Macro[ascan]:    1        5.5        0         0         0         0      1.49128 
2018-01-05 11:40:04,572 SScan: [START] motion
2018-01-05 11:40:05,188 SScan: [ END ] motion
2018-01-05 11:40:05,188 SScan: [START] acquisition
2018-01-05 11:40:05,203 SScan: [ END ] acquisition
2018-01-05 11:40:05,204 Door/issue102/1.Macro[ascan]:    2         10        0         0         0         0      2.12502 
2018-01-05 11:40:05,205 Door/issue102/1.Macro[ascan]: Scan #42 ended at Fri Jan  5 11:40:05 2018, taking 0:00:02.141716. Dead time 100.0% (motion dead time 97.9%)
2018-01-05 11:40:05,244 Door/issue102/1.MacroExecutor: [ END ] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c90c8470-f204-11e7-bc82-901b0e40a03c'


that I find really confusing for the user, in something like this:


2018-01-05 11:37:19,923 Door/issue102/1.MacroExecutor: [START] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> 67d090a2-f204-11e7-b3b9-901b0e40a03c'
2018-01-05 11:37:19,924 Door/issue102/1.Macro[ascan]: Scan #41 started at Fri Jan  5 11:37:19 2018. It will take at least 0:00:02.048528
2018-01-05 11:37:19,925 Door/issue102/1.Macro[ascan]:  #Pt No    mot17      ct17      ct18      ct19      ct20       dt   
2018-01-05 11:37:20,807 Door/issue102/1.Macro[ascan]:    0         1         0         0         0         0      0.86803 
2018-01-05 11:37:21,442 Door/issue102/1.Macro[ascan]:    1        5.5        0         0         0         0      1.50493 
2018-01-05 11:37:22,070 Door/issue102/1.Macro[ascan]:    2         10        0         0         0         0      2.13288 
2018-01-05 11:37:22,071 Door/issue102/1.Macro[ascan]: Scan #41 ended at Fri Jan  5 11:37:22 2018, taking 0:00:02.146527. Dead time 100.0% (motion dead time 97.8%)
2018-01-05 11:37:22,144 Door/issue102/1.MacroExecutor: [ END ] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> 67d090a2-f204-11e7-b3b9-901b0e40a03c'

I will commit the code with this default filter.
We could also use an environment variable, LogMacroFilter, to set additional filters if needed.
What do you think about this?, should we also allow disabling the default filters that I have implemented?.

teresanunez:
If I set to format to:

senv LogMacroFormat "%(asctime)s: %(message)s"

I get an output very similar to what our users have now with my manually implementation, so I think
that there will not be problems to get this in our beamlines.
The output would be:

2018-01-05 11:54:14,311: [START] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c4702258-f206-11e7-a9c6-901b0e40a03c'
2018-01-05 11:54:14,312: Scan #44 started at Fri Jan  5 11:54:14 2018. It will take at least 0:00:02.048528
2018-01-05 11:54:14,313:  #Pt No    mot17      ct17      ct18      ct19      ct20       dt   
2018-01-05 11:54:15,194:    0         1         0         0         0         0      0.870536
2018-01-05 11:54:15,823:    1        5.5        0         0         0         0      1.49689 
2018-01-05 11:54:16,449:    2         10        0         0         0         0      2.12415 
2018-01-05 11:54:16,449: Scan #44 ended at Fri Jan  5 11:54:16 2018, taking 0:00:02.137161. Dead time 100.0% (motion dead time 98.0%)
2018-01-05 11:54:16,498: [ END ] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c4702258-f206-11e7-a9c6-901b0e40a03c'

reszelaz:
I like the way you implemented the generic `LogFilter` class it may be very powerful. Actually I was thinking about something much simpler. Just an environment variable to set a python class (with module that aggregates it) - it would need to be importable on the machine where the MacroServer runs or be added to they `PythonPath` property. This way we delegate it to the user to develop and maintain this code. In the documentation we could say that this class must be a valid python logging filter redirecting to the official python documentation. I think that this simpler solution would be easier to maintain in the long term. What do you think?

Then in the core I would implement just one class, with the default filters that you consider **minimal** in order to make the logs relatively readable. Then, I think we would always need this class to be used as a filter.

We should also warn the users in the docs that the format of this logs may change, so we do not recommend to use it for extracting information programmatically e.g. parsing.

I must miss something, but in https://github.com/sardana-org/sardana/pull/480#issuecomment-355528067 you mentioned that you have just used a custom format. That means that the default filter is still in use, isn’t it? Then why do you see in the output lines like:
```
2018-01-05 11:54:14,311: [START] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c4702258-f206-11e7-a9c6-901b0e40a03c'
```
and 
```
2018-01-05 11:54:16,498: [ END ] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c4702258-f206-11e7-a9c6-901b0e40a03c'
```
Shouldn’t they be discarded by the default filters?

And just for the curiosity: do you know if it is possible to modify the message (or other record fields) with the filter?

teresanunez:
> I like the way you implemented the generic LogFilter class it may be very powerful. Actually I was thinking about something much simpler. Just an environment variable to set a python class (with module that aggregates it) - it would need to be importable on the machine where the MacroServer runs or be added to they PythonPath property. This way we delegate it to the user to develop and maintain this code. In the documentation we could say that this class must be a valid python logging filter redirecting to the official python documentation. I think that this simpler solution would be easier to maintain in the long term. What do you think?

I think we could maintain both:
- an environment variable setting what you want to filter (field and word), so the users do not have
to take care of defining a function and loading a module 
- another environment variable for setting an user defined filter, for those who really want/need to
 make more customized changes.
Both solutions require a certain knowledge about the output, but this can be given as examples.



> I must miss something, but in #480 (comment) you mentioned that you have just used a custom format. That means that the default filter is still in use, isn’t it? Then why do you see in the output lines like:
> 
> 2018-01-05 11:54:14,311: [START] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c4702258-f206-11e7-a9c6-901b0e40a03c'
> 
> and
> 
> 2018-01-05 11:54:16,498: [ END ] runMacro Macro 'ascan(mot17, 1.0, 10.0, 2, 0.0) -> c4702258-f206-11e7-a9c6-901b0e40a03c'
> 
> Shouldn’t they be discarded by the default filters?

No, the default filters do not discard those lines because are the only ones where one can see which macro is running. The last one (the one with END could be discarded, but the first one has to be there, where we could do is modified it).

> And just for the curiosity: do you know if it is possible to modify the message (or other record fields) with the filter?

Yes, in the filter one can modify all the record fields. If you want to modify them, we could agree in default modifications and leave any other one for the possible user defined filters.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1319,56 @@ def __runXMLMacro(self, xml):
...
+                    self.logging_mode = macro_obj.getEnv("LogMacroMode")
+                except:
+                    self.logging_mode = 0
+                try:
+                    logging_path = macro_obj.getEnv("LogMacroDir")
+                except:
+                    macro_obj.setEnv("LogMacroDir", "/tmp")
+                    logging_path = macro_obj.getEnv("LogMacroDir")
+                    
+                log_file = logging_path + "/spock_session_" + socket.gethostname() +".log"
```
reszelaz: Here I have two comments:
- I would not use "spock" in the file name - other clients e.g. macroexecutor could also take a profit from this feature
- maybe in the file name we could use the door name instead of the hostname. Otherwise if we run two clients, each on a separate door this would log to the same file.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1319,56 @@ def __runXMLMacro(self, xml):
...
+                logging_onoff = False
+            if logging_onoff:
+                try:
+                    self.logging_mode = macro_obj.getEnv("LogMacroMode")
+                except:
+                    self.logging_mode = 0
+                try:
+                    logging_path = macro_obj.getEnv("LogMacroDir")
+                except:
+                    macro_obj.setEnv("LogMacroDir", "/tmp")
```
reszelaz: Also two comments:
- Do you think it is necessary to set the environment variable, or just silently use the default one?
- I'm not sure, but this may not work on Windows. What if you use `os.path.join` and `os.sep`. For an example look in `sardana.tango.core.util.prepare_logging`.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -1290,9 +1319,56 @@ def __runXMLMacro(self, xml):
...
+                    logging_path = macro_obj.getEnv("LogMacroDir")
+                    
+                log_file = logging_path + "/spock_session_" + socket.gethostname() +".log"
+
+                if self.logging_mode:
+                    if os.path.isfile(log_file):
+                        try:
+                            os.system("vrsn -s %s" % log_file)
```
reszelaz: What `vrsn` does? It does not look like a platform independent thing.

It must to do something with the append/new file feature. What about using the `RotatingFileHandler` from python logging?

reszelaz:
@teresanunez, thanks for the clarification. I have now review the code and added some code review comments.

Regarding customization of filters by the user, if with what we will leave by default would be enough for the current use cases, I would still not open to the user the possibility to change the filtering. If it is requested in the future, we could always add it *a posteriori*.

I still have problems to understand the `LogFilter` class. Should it discard or allow the `param` string in the `field`? 

I agree that we need to leave the `[START]` message. And I do not mind to leave the `[ END ]` messages if the users are fine with that. If could be even useful as a clear separator between the macros execution.

src/sardana/macroserver/msmacromanager.py
teresanunez: I have changed the building of the path according to your suggestion. But I still kept the set of the
MacroLogDir environment. I am not sure if it is better to set it or not ... But perhaps it is useful for the users ... if they activate the logging and they don't know which the default directory is, they will see it listing the environment. What do you think?

teresanunez:
> Regarding customization of filters by the user, if with what we will leave by default would be enough for the current use cases, I would still not open to the user the possibility to change the filtering. If it is requested in the future, we could always add it a posteriori.

I agree.

> I still have problems to understand the LogFilter class. Should it discard or allow the param string in the field?

It is discarded.



> I agree that we need to leave the [START] message. And I do not mind to leave the [ END ] messages if the users are fine with that. If could be even useful as a clear separator between the macros execution.

Perfect, I leave it like it is.

Many thanks.

src/sardana/macroserver/msmacromanager.py
teresanunez: As far as I could read and test, the RotatingFIleHandler allows the rotation of the files based on the size, but what the users want is to have one file for each macro (in case they run scans, so that each scan is in one file). I do not get
this with a RotatingFileHandler.

reszelaz: I've never used it this way so it may not work. But please try to set the `maxBytes=0` or `backupCount=0` in the constructor (I think that these are even the default values) and then call the `doRollover` when you wan to change the file. This is how I interpret the [docs](https://docs.python.org/2/library/logging.handlers.html#rotatingfilehandler).

BTW why the users want to have a dedicated log file per scan? Isn't the role of the `ScanFile`?

src/sardana/macroserver/msmacromanager.py
reszelaz: An alternative would be to print the info message after enabling the logging where the logs will be saved. If you think it is not enough then leave the environment setting.

src/sardana/macroserver/msmacromanager.py
teresanunez: Hi @reszelaz, I found the way to do this. It is commited.

src/sardana/macroserver/msmacromanager.py
teresanunez: Hi @reszelaz, I implemented now like this ... but it is a little bit annoying because one gets a message like:
Log file saved in /tmp

everytime a macro is started. One could put it in debug instead of info ... If it is enough ... the users would say.

reszelaz: I agree it would be too much to print it everytime. So let's leave the environment variable setting.

src/sardana/macroserver/msmacromanager.py
reszelaz: I posted #689 that fixed the issue with door's name.

src/sardana/macroserver/msmacromanager.py
reszelaz: Great that it works! I have tested it and the current behavior is that as long as the mode is set to 1 (using the `logmacro` macro) every macro execution starts a new file. Is it correct? Then to simply change the file one would need to execute the `logmacro` macro twice, isn't it? Like for example:
```
logmacro True 1
logmacro True 0
```

src/sardana/macroserver/msmacromanager.py
reszelaz: I'm thinking we are getting close to the integration of this PR, could you already remove these references? This will make the review process cleaner.

src/sardana/macroserver/msmacromanager.py
```diff
@@ -74,7 +77,32 @@
...
+    # module    -> __init__
+    # message   -> 2         10        0         0         0         0      2.11293 
+
+    def __init__(self, param=None, field=None):
+        self.param = param
+        self.field = field
+
+    def filter(self, record):
+        allow = True
+        if self.param != "MacroExecutor" and "MacroExecutor" in record.name:
```
reszelaz: It took me some time to understand the logic of `LogFilter`. What do you think to refactor it to make it more intuitive. If I remember correctly, this precise condition is for filtering out some messages in case of executing nested macros. Maybe adding an extra argument to the constructor, let's call it `logger` and make the matching with the regular expression could make it easier to follow.

reszelaz:
@teresanunez, sorry for the delay. I think I have responded to all open issues (there was one still from some previous reviews). In general we are quite close to be ready to integrated.

I've just found the following problem, could you check it:
```
Door_zreszela_1 [9]: umv mot01 10
     mot01
   10.0000
An error occurred while running Macro 'umv([['mot01', '10.0']]) -> ec2950bc-00fc-11e8-ae2d-f8b156a48f79':
local variable 'logging_onoff' referenced before assignment
```

src/sardana/macroserver/msmacromanager.py
teresanunez: Yes, it works exactly like you describe. Thanks for testing.

reszelaz: Thanks, it looks good! Just an idea, no need to change it:
What about having a specific macro `logmacro_rollover` that would do this? Then, I think that we could remove the `mode` argument of the `logmacro`.

src/sardana/macroserver/msmacromanager.py
teresanunez: I thinks it is done.

teresanunez:
> I've just found the following problem, could you check it:
> 
> Door_zreszela_1 [9]: umv mot01 10
>      mot01
>    10.0000
> An error occurred while running Macro 'umv([['mot01', '10.0']]) -> ec2950bc-00fc-11e8-ae2d-f8b156a48f79':
> local variable 'logging_onoff' referenced before assignment

I think it is solved with the cleanup  I did. I do not get the message.

teresanunez:
> I've just found the following problem, could you check it:
> 
>     Door_zreszela_1 [9]: umv mot01 10
>     mot01
>     10.0000
>     An error occurred while running Macro 'umv([['mot01', '10.0']]) -> ec2950bc-00fc-11e8-ae2d-f8b156a48f79':
>     local variable 'logging_onoff' referenced before assignment
> 
> I think it is solved with the cleanup I did. I do not get the message.

No, it is not. I am looking.

teresanunez:
> I've just found the following problem, could you check it:
> 
>     Door_zreszela_1 [9]: umv mot01 10
>     mot01
>     10.0000
>     An error occurred while running Macro 'umv([['mot01', '10.0']]) -> ec2950bc-00fc-11e8-ae2d-f8b156a48f79':
>     local variable 'logging_onoff' referenced before assignment
> 
>     I think it is solved with the cleanup I did. I do not get the message.
> 
> No, it is not. I am looking.

It is now solved. It was related with the macros having a parent macro, in which logging_onoff was not defined. One could solve it in different ways. I simply defined logging_onoff for all macros.

src/sardana/macroserver/msmacromanager.py
teresanunez: Do you mean with an argument to set it on or off?

reszelaz: Maybe I did not understand well this feature. Are the users also interested in producing one log file per one macro execution. Or, in other words, would they continue executing macros when the mode is 1?
If not, and this is just to make a rollowever and then return to the mode 0, I think that the parameter is not necessary. The macro execution would do the rollover and that's it.

teresanunez: They requested to have it automatically, per one macro execution.

src/sardana/macroserver/msmacromanager.py
reszelaz: I'm not sure if this assignments are necessary:
`macro_obj.fileHandler = logging.handlers.RotatingFileHandler(log_file, backupCount = bck_counts)`
`macro_obj.logger = macro_obj.getLogger()`.
Or in other words does the macro_obj require special references to `fileHandler` and `logger`.

src/sardana/macroserver/msmacromanager.py
reszelaz: Aaa, ok. Then forget about my proposal. I understood it wrong. THanks!

src/sardana/macroserver/msmacromanager.py
teresanunez: I remove them, they are not needed.

src/sardana/macroserver/msmacromanager.py
teresanunez: Hi @reszelaz, sorry, I do not understand which is the change that you propose. It is using a matching with regular expresions ... the line you mention is just an exception for keeping the line in which the macro name and arguments appear.

src/sardana/macroserver/msmacromanager.py
reszelaz: Great thanks!

src/sardana/macroserver/msmacromanager.py
reszelaz: Sorry for the late reply! I think it would be easier if I propose a change directly instead of explaining it. Then you will evaluate it. Soon I will make a PR to your branch.

reszelaz:
I also think that it would be clenaer to encapsulate all the logic in a class, let's call it `LogMacroManager` that would be instantiated with the macro object. It could expose two methods e.g. `start` and `stop` or `enable` and `disable` that would do the job. This way we would not _clutter_ the already complex `MacroExecutor.runMacro` method. What do you think? If its not clear, I can propose a change with a PR to your branch and then we can discuss it. As you prefer.

Apart of that I will propose a change to the `LogFiler` as I commented previously.

Then after fixing small flake8 errors it would be ready for integration:)

Many thanks!

teresanunez:
Hi @reszelaz, thanks for your review. I think it is better if you make PRs for the last two changes you hav propose. I guess this would be fasteer. Many thanks.

PR #482, opened 2017-06-19
Avoid import OrderedDict from taurus

### LINKED ISSUE #481: Avoid import OrderedDict from taurus
Sardana imports `OrderedDict` module from `taurus.external` and it will be dropped in the next release.
See taurus [#223](https://github.com/taurus-org/taurus/issues/223) 

All occurrences of `from taurus.external.ordereddict import OrderedDict` must be replaced.

e.g.
```
try:
    from collections import OrderedDict
except ImportError:
    # For Python < 2.7
    from ordereddict import OrderedDict
```

cmft:
Sardana imports OrderedDict module from taurus.external and it will
be dropped in the next release.

Fix #481

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,8 @@ requires = [
     'itango (>=0.0.1)',
     'taurus (>= 3.6.0)',
     'lxml (>=2.1)',
+    # ordereddict is necessary for Python < 2.6
+    'ordereddict'
 ]
 
 console_scripts = [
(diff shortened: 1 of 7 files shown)


reszelaz:
I have just added an explicit requirement to setup.py for Python < 2.7. Keep in mind that `requires` does not install the required packages when using `pip` but just describe the packages relationships.
Thanks!

PR #483, opened 2017-06-22
Fix bug when updating macro arguments

jordiandreu:
A string value is expected as argument for updateMacroArgument
method. However, this value comes from any valid signal and
could be any other object.

Remove the regular expression statement before string conversion.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
@@ -230,7 +230,7 @@ class MacroButton(TaurusWidget):
             self.macro_args.append('')
         # some signals may come with other than string argumenst e.g. int
         # so convert them to string
-        value = Qt.from_qvariant(value, str)
+        value = str(value)
         # string arguments may contain spaces so encapsulate them in quotes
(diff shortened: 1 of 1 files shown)


reszelaz:
Thanks for your contribution! I have some comments...

We need to keep the conditional encapsulation of the string parameters within quotes in order to be compatible with parameters that contain spaces. It was introduced in b49f62eb3 

What I suggest is to relax the `updateMacroArgument` API to accept arguments of other types than `str` and convert these values with `Qt.from_qvariant`.

Please check my proposal and feedback.

cpascual:
> What I suggest is to relax the updateMacroArgument API to accept arguments of other types than str and convert these values with Qt.from_qvariant.

Do not use Qt.from_qvariant. It is not a standard method and provided only [for backwards compatibility](http://pyqt.sourceforge.net/Docs/PyQt5/pyqt_qvariant.html). Better use standard python casting

reszelaz:
Thanks! I have already changed the code. @jordiandreu please check it now.

jordiandreu:
I have already checked the code proposed by @reszelaz and it works fine. So, in my opinion you can go ahead.

reszelaz:
Thanks for the feedback. I've just merged it.

PR #484, opened 2017-06-28
Pip install requirements

reszelaz:
Handle install requirements when using `pip`, ~~also conditionally install `ordereddict` when using Python < 2.6~~. 
Apart of that update the documentation to install with `pip` instead of `easy_install`.


reszelaz:
Unfortunately the conditional specifier does not work with older setuptools, so it can not be done for the ordereddict.

PR #485, opened 2017-06-28
Add timescan

reszelaz:
This PR provides timescan (solves #104).

In order to reuse as much as possible from the existing scan framework I made a decision to relax the “motion” requirements of the `GScan` class so it could be used in scans which do not involve moveables. An alternative solution that I have evaluated was to develop a lower level abstract class e.g. `BaseScan` that would just take care of the basic scan setup and leave the `GScan` strongly linked to the “motion” concepts. I was not sure if this refactoring is really necessary and I consider it more complicated to implement, that’s why I decided to leave only the `GScan` class. But of course we can still do this refactoring in the future, if necessary.

The summary of changes:

- `timescan` macro was added that executes equidistant time scans.
- `TScan` class was developed that may serve to develop other types of timescans e.g. non-equidistant.
- `CAcquisition` class was developed that provides the common logic to `CTScan` and `TScan` such as, processing `ValueBuffer` events, validating compatibility of the measurement group, etc.
- `timescan` macro similarly to the `ascanc` & co. macros does not provide the correct information about the deadtime at the end of the scan. IMHO this can be treated later as an issue.

This PR also fix some issues related with the recent changes in Taurus4 related to the ranges.

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -149,7 +149,7 @@ class aNscan(Hookable):
                                      self._period_generator, moveables, env, constrains, extrainfodesc)
             elif mode == ContinuousHwTimeMode:
                 self.nr_interv = scan_length
-                self.nr_of_points = self.nr_interv + 1
+                self.nr_points = self.nr_interv + 1
                 mg_name = self.getEnv('ActiveMntGrp')
                 mg = self.getMeasurementGroup(mg_name)
                 mg_latency_time = mg.getLatencyTime()
@@ -245,8 +245,8 @@ class aNscan(Hookable):
             'post-move') + [self._fill_missing_records]
         step["post-move-hooks"] = post_move_hooks
         step["check_func"] = []
-        step["active_time"] = self.nr_of_points * (self.integ_time +
-                                                   self.latency_time)
+        step["active_time"] = self.nr_points * (self.integ_time +
+                                                self.latency_time)
         step["positions"] = []
         step["start_positions"] = []
         starts = self.starts
@@ -321,7 +321,7 @@ class aNscan(Hookable):
 
     def _fill_missing_records(self):
         # fill record list with dummy records for the final padding
-        nb_of_points = self.nr_of_points
+        nb_of_points = self.nr_points
         scan = self._gScan
         nb_of_records = len(scan.data.records)
         missing_records = nb_of_points - nb_of_records
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -1485,3 +1485,35 @@ class a4scanct(aNscan, Macro):
                       nr_interv, integ_time, mode=ContinuousHwTimeMode,
                       latency_time=latency_time
                       ** opts)
+
+
+class timescan(Macro, Hookable):
+    """Do a time scan over the specified time intervals. The scan starts
+    immediately. The number of data points collected will be nr_interv + 1.
+    Count time is given by integ_time. Latency time will be the longer one
+    of latency_time and measurement group latency time.
+    """
+
+    param_def = [
+        ['nr_interv', Type.Integer, None, 'Number of scan intervals'],
+        ['integ_time', Type.Float,   None, 'Integration time'],
+        ['latency_time', Type.Float, 0, 'Latency time']]
+
+    def prepare(self, nr_interv, integ_time, latency_time):
+        self.nr_interv = nr_interv
+        self.nr_points = nr_interv + 1
+        self.integ_time = integ_time
+        self.latency_time = latency_time
+        self._gScan = TScan(self)
+
+    def run(self, *args):
+        for step in self._gScan.step_scan():
+            yield step
+
+    def getTimeEstimation(self):
+        mg_latency_time = self._gScan.measurement_group.getLatencyTime()
+        latency_time = max(self.latency_time, mg_latency_time)
+        return self.nr_points * (self.integ_time + latency_time)
+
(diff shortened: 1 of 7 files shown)


teresanunez:
Hi @reszelaz, thanks for this implementation, our users are also very interested on that.
I have tried to test it, and I get the error:

An error occurred while running Macro 'timescan(10, 1.0, 0.0) -> 0a7488d2-5d80-11e7-8181-901b0e40a03c':
subscribeEvent() takes at most 3 arguments (4 given)

I don't know if it has something to do with my set up, but I do not see what I could change.

                     Regards,
                                    Teresa

LJBD:
A quick note: at SOLARIS, @daneos have implemented something vaguely similar.

reszelaz:
> subscribeEvent()` takes at most 3 arguments (4 given)

You need to use the latest taurus (either develop or support-3.x).

Also keep in mind that this branch has still some errors (travis-ci is in red). They affect step scans when using Taurus3. I have them already fixed locally. I will commit soon. But the timescan should be working..

> at SOLARIS, @daneos have implemented something vaguely similar.

Great! So we need to merge it into a one solution. There is no sense to maintain two versions of timescan.

teresanunez:
Hi @reszelaz, I tried with taurus4 ... but well, perhaps I had some version mix somewhere.
 I sill try again after the travis problems are solved. Many thanks.

reszelaz:
Now the issues with Taurus3/Taurus4 compatibilities are resolved. To test it please remember to use the latest revisions of Taurus develop/support-3.x branches. Thanks!

teresanunez:
Hi @reszelaz, thanks. It works fine for me now.

teresanunez:
@reszelaz I observed one thing ... I would expect that the counters/timers behaves in the
 same way if I make the scan with a dummy motor or using timescan (this was the purpose of that,
 wasn't it), but this is not the case. If I make the scan with a dummy motor the readouts are:

> ascan mt_dmy01_ts 1 10  5 1 
Pt No   mt_dmy01_ts  mt_timer1_ts  mt_counter1_ts     dt   
   0           1            1              2         0.412021
   1          2.8           1              2         1.82655 
   2          4.6           1              2         3.24354 
   3          6.4           1              2         4.65269 
   4          8.2           1              2         6.06447 
   5           10           1              2         7.47921 

But the same with timescan gives as readout:

> timescan 5 1

  Pt No   mt_timer1_ts  mt_counter1_ts     dt   
Relative timestamp (dt) column contains theoretical values
   0           1              2            0    
   1          nan            nan           1    
   2           1              2            2    
   3          nan            nan           3    
   4           1              2            4    
   5          nan            nan           5    

If I put the third argument in the timescan (the latency time), then I get values for all
 points ... but never for the last one (even with latency times of 3 seconds, which is
 by far larger than the time the motor needs to move). Is this what the users would
 expect when they make a time scan?, I am missing perhaps something ...

teresanunez:
Thanks @reszelaz, now it is fine for me. I understand, that the people has to set a latency
 time if this is needed by timer/counters.

reszelaz:
Thanks for pointing this out. There was a bug that I have already fixed. So now you can see the last point (nr_interv + 1).

Regarding the "skipped" acquisitions you need to use some latency_time. In the case of my PC, with dummy controllers, latency_time of 0.04 s is enough to overcome skipped points. With the real hardware this will depend on the controller's implementation, the number of controllers/channels in the measurement group, etc.

teresanunez:
@reszelaz, yes, thanks. One thing more, if the latency time is needed, should it not be better to
 don't put it as optional?, but force the users to set it?. Or put some default value like 0.04 and
 not 0 ...

reszelaz:
Good point!

The real hardware controllers have a chance to report the preferred latancy time with the controller parameter `latency_time`. This, however, describes the latency time needed between two consecutive acquisitions in the hardware synchronization. The measurement group exposes the `latency_time` attribute which is the maximum of the latency times of all the controllers belonging to it. The scan selects the higher of these two: the `latency_time` attribute of the measurement group and the `latency_time` macro parameter. So if the macro parameter is 0 then the measurement group's latency time is used. Up to now everything looks nice in the hardware acquisition...

In the software acquisition, the latency time of the controllers does not have too much sense. It may be interesting to set a `latency_time` attribute for the measurement group with the empirically measured value. This value may depend on the controllers and channels used in the measurement group, the computational power and the current load of the PC where the Pool runs, the network traffic that may affect communication with the hardware, etc. Since the latency time used within the software synchronization is so complex to determine I prefer not hardcode it.

What we could do is either:

1. Change the `latency_time` attribute of the measurement group to writable (now it is read only) and allow the user to assign a value to it. So if the user sets a value to lower than the maximum of the latency time of all the controllers, it will be ignored. And if he sets a value to higher than the maximum of the latency time of all the controllers it will used. In order to reset it, one would need to set it to 0 so again the value reported by the controllers would be used.
2. Define the environment variable `MntGrpLatencyTime` that would be used if the user does not pass the `latency_time` argument to the macro.

I'm more in favor of the first option, cause for me the latency time it is somehow related to the definition of the Measurement Group. I'm not too much fan of hardcoding a default value of the `latency_time` macro parameter.

teresanunez:
Hi @reszelaz, thanks, I see your arguments and I agree with them. It can stay like it is ... Setting
 a latency time in the MG requires also a knowledge about that from the users ... so they can set
 it in the argument of the timescan macro, no need to have more extra places.

reszelaz:
Ok, then let's wait for the feedback from the users. We can add either of these two options at any time or even think more on how to make it more user-friendly. Do you think it is ready for integration? If so could you please accept it?

Issue #486, opened 2017-06-30
TwoDExpChannel: 'Value' object has no attribute 'value_buffer'


jkotan:
Hi,

I've just wanted to test sardana from the current develop branch and with 2D channel (pilatus)
I get the following error:

    SardanaTP.W001 DEBUG    2017-06-30 07:47:53,062 p09/door/haso228k.01: Traceback (most recent call last):
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/msmacromanager.py", line 1251, in runMacro
	for step in macro_obj.exec_():
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/macro.py", line 2274, in exec_
	for i in it:
      File "/usr/lib/python2.7/dist-packages/sardana/macroserver/macros/scan.py", line 256, in run
	for step in self._gScan.step_scan():
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/scan/gscan.py", line 877, in step_scan
	self.start()
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/scan/gscan.py", line 832, in start
	self.data.start()
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/scan/scandata.py", line 289, in start
	self.datahandler.startRecordList(self)
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/scan/recorder/datarecorder.py", line 56, in startRecordList
	recorder.startRecordList(recordlist)
      File "/home/jkotan/testsar/lib/python2.7/site-packages/sardana-2.2.5_alpha-py2.7.egg/sardana/macroserver/scan/recorder/datarecorder.py", line 115, in startRecordList
	self._startRecordList(recordlist)
      File "/usr/lib/python2.7/dist-packages/sardananxsrecorder/nxsrecorder.py", line 808, in _startRecordList
	self.__setNexusDevices()
      File "/usr/lib/python2.7/dist-packages/sardananxsrecorder/nxsrecorder.py", line 438, in __setNexusDevices
[...]

PR #489, opened 2017-07-03
Solving issue #486

teresanunez:


teresanunez:
Hi @reszelaz, I simply removed what you have added for CS ... but I think it has not sense to have
 it there. Well, let's see what you think. Thanks.

reszelaz:
You are right, this protection is not necessary because 2D's are still not supported by continuous scans. This code can be removed. I'm accepting your PR. Thank you!

PR #492, opened 2017-07-05
Fix expconf when using taurus --tango-host option

reszelaz:
This issue was originally reported in Taurus:
https://github.com/taurus-org/taurus/issues/79 but it appeared to be
a bug in Sardana. Fix it by respecting the tango database when creating
PyTango.Group object.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/macroserver.py
+++ b/src/sardana/taurus/core/tango/sardana/macroserver.py
@@ -188,11 +188,13 @@ class ExperimentConfiguration(object):
         elif isinstance(scan_file, (str, unicode)):
             scan_file = [scan_file]
         ret['ScanFile'] = scan_file
-        mnt_grps = macro_server.getElementNamesOfType("MeasurementGroup")
+        mnt_grps = macro_server.getElementsOfType("MeasurementGroup")
+        mnt_grps_names = [mnt_grp.name for mnt_grp in mnt_grps.values()]
+        mnt_grps_full_names = mnt_grps.keys()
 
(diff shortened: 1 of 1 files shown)


cpascual:
I reviewed the code (but did not test) and it seems ok for me. Waiting for confirmation from @jkotan  to merge it.

@reszelaz : would it be easy / useful to add an automatic test for this?  ... I am thinking in avoiding regressions and use it as an example for similar issues (I suspect this won't be the last bug related to short vs full names)

reszelaz:
> would it be easy / useful to add an automatic test for this? ... I am thinking in avoiding regressions and use it as an example for similar issues (I suspect this won't be the last bug related to short vs full names)
(this is not a request, just an open question)

I like the idea of having an automated test for that but I think that it won't be easy to implement with the infrastructure that we have right now.

If we try to reuse the "demo1" environment with the real Pool and MacroServer device servers running in a docker container, as we do for the macro tests, it would require launching another instance of it to be able to access a different TANGO_HOST. I estimate that it will take me some time.

We could try to patch PyTango to have a properly isolated unit test, not dependent on the "demo1" infrastructure, but I don't have experience with that and it would take me time to do it.

So I would abandon this unless there are other ideas on how to implement it.

teresanunez:
Hi @cpascual, I have tested it at DESY and it is working.

jkotan:
Hi Zibi,

unfortunately, currently I do not have VMs set to communication between two different domains (i.e. using only FQDN). But from my tests inside one domain it works OK.

Bests,
Jan

PR #495, opened 2017-07-12
Fix Problem in loading a favorite macro in macroexecutor

### LINKED ISSUE #441: Problem in loading a favorite macro in macroexecutor
When loading a macroexecutor previously saved favourite macro (with parameter values stored) in the same macroexecutor a bug appears.

And it does not let you to change the parameters of the recently loaded favourite macro.

<pre>
  File "/homenfs/mrosanes/PycharmProjects/sardana/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macroparameterseditor/model.py", line 63, in flags
    not node.type() in globals.EDITOR_NONEDITABLE_PARAMS):
  File "/homenfs/mrosanes/PycharmProjects/sardana/src/sardana/taurus/core/tango/sardana/macro.py", line 502, in type
    return self._type
AttributeError: 'SingleParamNode' object has no attribute '_type'
</pre>

cmft:
Macroexecutor raises an Attribute error if you try to execute
a "favorite macro" after restarting the macroexecutor.

Fix #441 

The flake8 could fail cause a URL in a comment  exceed the 79 characters

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macroexecutor.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macroexecutor.py
@@ -690,6 +690,12 @@ class TaurusMacroExecutorWidget(TaurusWidget):
         self.tabMacroListsWidget.addTab(
             self.historyMacrosViewer, "History Viewer")
         splitter.addWidget(self.tabMacroListsWidget)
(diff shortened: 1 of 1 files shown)


reszelaz:
Your solution solves the issue. I merge it.
Note that the older versions of PyQt e.g. 4.4.3 were not affected by this bug.

PR #503, opened 2017-07-14
Properly synchronize stopping of macro and its reserved objects

### LINKED ISSUE #8: interrupted and/or stopped dscan does not return to initial position (SF#7)
Note: This is an internal ALBA ticket (RT#21632)that is passed to the sardana tracker.

Sun Jan 29 15:59:26 2012 ffauth - Ticket created 
From: 	F Fauth

Two cases,

1- I launched a dscan but interrupted with ctrlC, is it normal the motor did not go to the initial position
2- I launched a dscan, it started correctly but could not finish because of soft limit. Is it normal the motor did not go to the initial position. ? Is it normal that the scan starts though he will not be able to reach the final position of teh scan ?

Reported by: cpascual ( http://sf.net/u/cpascual )

Original Ticket: [sardana/tickets/7](https://sourceforge.net/p/sardana/tickets/7)

reszelaz:
The macro reserved objects e.g. motion or measurement groups are stopped
when the user stops the macro e.g. Ctrl+C in spock, or directly door's
StopMacro command. This stop is executed in a worker thread which is
comletly desynchronized with the macro execution thread.

Fix it forcing the MacroExecutor to wait for the stopping thread to finish its job
before executing the on_stop or on_abort methods. Also protect macros that
catch the StopException by waiting in the mAPI decorator.

Fixes #8.

@sardana-org/integrators any volunteer to integrate it?

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -308,12 +308,14 @@ def mAPI(fn):
             if self._shouldRaiseStopException():
                 if is_macro_th:
                     self.setProcessingStop(True)
+                self.executor._waitStopDone()
                 raise StopException("stopped before calling %s" % fn.__name__)
         ret = fn(*args, **kwargs)
         if not self.isProcessingStop():
             if self._shouldRaiseStopException():
                 if is_macro_th:
                     self.setProcessingStop(True)
+                self.executor._waitStopDone()
                 raise StopException("stopped after calling %s" % fn.__name__)
         return ret
     return new_fn
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -37,6 +37,7 @@ import copy
 import inspect
 import functools
 import traceback
+import threading
 
 from lxml import etree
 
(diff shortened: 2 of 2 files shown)


teresanunez:
HI @reszelaz, I can try it.

teresanunez:
Hi @reszelaz, one comment, the point 2 in #8 is not solved (I know you only meant the point 1 with
this PR), should be #8 in any case closed or should I integrate this without closing it?.

reszelaz:
Good point! I completely forgot about it. Let me investigate it a little bit and see if it is possible to fix it for Jul17.

reszelaz:
I paste here the user request so it facilitates the discussion:
> 2- I launched a dscan, it started correctly but could not finish because of soft limit. Is it normal the motor did not go to the initial position. ? Is it normal that the scan starts though he will not be able to reach the final position of teh scan ?

I have checked the first question "Is it normal the motor did not go to the initial position?" and in my case, it goes to the initial position. What I do to test it is:
1. Position the `mot01` in 0.
2. Set the software positive limit to 1.5 using Jive.
3. Start: `dscan mot01 -2 2 4 0.1`

The scan does not finish, it stops after point 3 (`mot01` at 1) with the following exception: `Set value for attribute Position is above the maximum authorized (at least element 0)`. But then the motor returns to the initial position 0. So I consider this as a correct behavior. 

Regarding the second question: "Is it normal that the scan starts though he will not be able to reach the final position of the scan?" This already has its own issue #46.

So I think that we could close #8 if you accept this PR.

teresanunez:
Hi @reszelaz, thanks, I did not realize that there was another issue for the second question of the
 point 2, so, I agree with you, #8 can be closed accepting this PR.

PR #511, opened 2017-07-28
Fix DumbRecorder

mrosanes:
Fix DumbRecorder.
- DataRecorder class constructor needs and extra argument. Add it.
- Fix DumbRecorder without using 'time' library.


reszelaz:
This makes the DumbRecorder functional. Thanks!

PR #513, opened 2017-07-28
Add macro example for get data

mrosanes:
Add macro example for executing a Macro from inside another Macro,
get its data, and calculate a result using these data.


reszelaz:
Thanks for the example! I'm merging it.

PR #532, opened 2017-08-07
Import OrderedDict from taurus.external

mrosanes:
For python versions < 2.7, import OrderedDict from
taurus.external.ordereddict

### DIFF
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -47,7 +47,7 @@ try:
     from collections import OrderedDict
 except ImportError:
     # For Python < 2.7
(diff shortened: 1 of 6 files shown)


reszelaz:
Thanks for this contribution! However, I would try to avoid merging it. This problem affects only python 2.6 and I doubt if anyone apart from us (Alba) still uses this version. Furthermore, Sardana now requires ordereddict and it is just a question of installing this dependency in order to avoid this problem. The dependency can be installed from PyPI or a system package - I was even able to find one for openSUSE 11.1 that was released in 2008.

cpascual:
... also, taurus.external.ordereddict was deprecated and has been removed in Taurus 4.1.0

PR #533, opened 2017-08-07
Macro move_async may fail if movement could not be started

reszelaz:
Move of a motor may not started if, for example, it was requested to
move out of the limits. In this case does not wait for the movement to
finish.

### DIFF
--- a/src/sardana/macroserver/macros/examples/motion.py
+++ b/src/sardana/macroserver/macros/examples/motion.py
@@ -42,10 +42,10 @@ class move_async(Macro):
                  ]
 
     def run(self, moveable, pos):
+        motion = self.getMotion([moveable])
+        self.info('initial position: %s' % motion.readPosition())
+        _id = motion.startMove([pos])
         try:
-            motion = self.getMotion([moveable])
-            self.info('initial position: %s' % motion.readPosition())
-            _id = motion.startMove([pos])
             # Do whatever here (while the moveable is moving)
             self.info('state: %s' % motion.readState())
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi @reszelaz, I have reviewed this. I have checked the case in which the motor can not be moved becaues the requested move is outside of the limits. For me neither like it was nor with your changes the macro wait for the movement. With your changes is better because one can see the cause of the error:
An error occurred while running Macro 'move_async(mot01, 11.0) -> ca02c0e6-ae8a-11e7-addc-901b0e40a03c':
Set value for attribute Position is above the maximum authorized (at least element 0)


Before applying your change one gets the error:
An error occurred while running Macro 'move_async(mot01, 10.0) -> 19c137ee-ae8a-11e7-addc-901b0e40a03c':
local variable '_id' referenced before assignment

because startMove gives an exception and does not return.
In any case the change is fine to me.

reszelaz:
Thanks for the review @teresanunez. What I observe is that in the case of a move within the allowed range the macro waits for the end of movement. And in the case when the requested position is outside of the limits it does not wait and ends with the error: `Set value for attribute Position is above the maximum authorized (at least element 0)`. But in the last case the movement was not started. I think this behavior is correct, isn't it?

teresanunez:
Hi @reszelaz, yes, I think, that the behaviour you describe is correct. It is also what I see.

reszelaz:
Great, then I'me merging it. Thanks!

PR #535, opened 2017-08-09
Fix warnings of deprecation warnings related to Taurus4

mrosanes:
Fix warnings of deprecation warnings related to Taurus4, 
by maintaining Taurus3 compatibility.

### DIFF
--- a/src/sardana/spock/spockms.py
+++ b/src/sardana/spock/spockms.py
@@ -570,7 +570,12 @@ class SpockMacroServer(BaseMacroServer):
             # TODO: when it becomes possible to do:
             # some taurus.Device.<attr name> = <value>
             # replace device_proxy with element
-            device_proxy = element.getObj().getHWObj()
+            try:
+                # Taurus4 usage
+                device_proxy = element.getObj().getDeviceProxy()
+            except:
+                # Taurus3 usage
+                device_proxy = element.getObj().getHWObj()
             genutils.expose_variable(element.name, device_proxy)
         return element
 
--- a/src/sardana/taurus/core/tango/sardana/macroserver.py
+++ b/src/sardana/taurus/core/tango/sardana/macroserver.py
@@ -326,8 +326,12 @@ class BaseDoor(MacroServerDevice):
             # TODO: For Taurus 4 compatibility
             from taurus.core import TaurusDevState
             self._old_sw_door_state = TaurusDevState.Undefined
+        try:
+            self.stateObj.addListener(self.stateChanged)
+        except:
+            # Fallback for Taurus3
+            self.getStateObj().addListener(self.stateChanged)
 
-        self.getStateObj().addListener(self.stateChanged)
 
         for log_name in self.log_streams:
             tg_attr = self.getAttribute(log_name)
--- a/src/sardana/taurus/core/tango/sardana/macroserver.py
+++ b/src/sardana/taurus/core/tango/sardana/macroserver.py
@@ -41,6 +41,7 @@ from lxml import etree
 
 import PyTango
 
+import taurus
 from taurus import Device, Factory
 from taurus.core.taurusmanager import TaurusManager
 from taurus.core.taurusbasetypes import TaurusEventType, TaurusSWDevState, \
@@ -59,7 +60,7 @@ from .pool import getChannelConfigs
 from .macro import createMacroNode
 
 CHANGE_EVT_TYPES = TaurusEventType.Change, TaurusEventType.Periodic
-
+taurus_version = taurus.Release.version
 
 def recur_map(fun, data, keep_none=False):
     """Recursive map. Similar to map, but maintains the list objects structure
@@ -574,12 +575,11 @@ class BaseDoor(MacroServerDevice):
         return result
 
     def stateChanged(self, s, t, v):
-        self._old_door_state = self.getState()
-        try:
-            self._old_sw_door_state = self.getSWState()
-        except:
-            # TODO: For Taurus 4 compatibility
+        if taurus_version.split(".")[0] == '4':
             self._old_sw_door_state = self.state
+        else:
+            # Taurus 3 compatibility
+            self._old_sw_door_state = self.getSWState()
 
     def resultReceived(self, log_name, result):
         """Method invoked by the arrival of a change event on the Result attribute"""
--- a/src/sardana/taurus/core/tango/sardana/macroserver.py
+++ b/src/sardana/taurus/core/tango/sardana/macroserver.py
@@ -95,7 +95,11 @@ class Attr(Logger, EventGenerator):
             self.fireEvent(None)
         elif type != TaurusEventType.Config:
             if evt_value:
-                self.fireEvent(evt_value.value)
+                if taurus_version.split(".")[0] == '4':
+                    self.fireEvent(evt_value.rvalue)
+                else:
+                    # Taurus 3 compatibility
+                    self.fireEvent(evt_value.value)
             else:
                 self.fireEvent(None)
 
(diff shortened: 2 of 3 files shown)


mrosanes:
This pull request would allow to: 
- Avoid many deprecation warnings which appear just opening spock; improving the user experience.
- Reduce the amount of logs appearing when running the MacroServer.

However the decision was taken to do not deal for the moment with the Sardana deprecation warnings coming from Taurus4, as it could probably give problems of compatibilities with other code developed on using the Sardana code.

src/sardana/spock/spockms.py
```diff
@@ -570,7 +570,12 @@ def _addElement(self, element_data):
             # TODO: when it becomes possible to do:
             # some taurus.Device.<attr name> = <value>
             # replace device_proxy with element
-            device_proxy = element.getObj().getHWObj()
+            try:
+                # Taurus4 usage
+                device_proxy = element.getObj().getDeviceProxy()
+            except:
```
LJBD: I wouldn't use such a broad `except` clause since it might overshadow some other exceptions happening inside `getDeviceProxy`. I'd use `except AttributeError` - AFAIK that's what's going to appear here with Taurus 3.

src/sardana/taurus/core/tango/sardana/macroserver.py
```diff
@@ -326,8 +342,12 @@ def __init__(self, name, **kw):
             # TODO: For Taurus 4 compatibility
             from taurus.core import TaurusDevState
             self._old_sw_door_state = TaurusDevState.Undefined
+        try:
+            self.stateObj.addListener(self.stateChanged)
+        except:
```
LJBD: Again the same thing as above. I'm not sure what error would happen here, presumably also `AttributeError`.

src/sardana/taurus/core/tango/sardana/pool.py
```diff
@@ -221,12 +224,20 @@ def eventReceived(self, evt_src, evt_type, evt_value):
         if evt_value is None:
             v = None
         else:
-            v = evt_value.value
+            try:
+                v = evt_value.rvalue
+            except:
```
LJBD: The same thing here. Also probably `AttributeError`.

src/sardana/taurus/core/tango/sardana/pool.py
```diff
@@ -506,7 +517,11 @@ def _information(self, tab='    '):
         try:
             # TODO: For Taurus 4 / Taurus 3 compatibility
             if hasattr(self, "stateObj"):
-                state_value = self.stateObj.read().rvalue
+                try:
+                    state_value = self.stateObj.read().rvalue
+                except:
```
LJBD: The same thing here. Also probably `AttributeError`.

src/sardana/taurus/core/tango/sardana/pool.py
```diff
@@ -1482,7 +1497,11 @@ def on_configuration_changed(self, evt_src, evt_type, evt_value):
         if evt_type not in CHANGE_EVT_TYPES:
             return
         self.info("Configuration changed")
-        self._setConfiguration(evt_value.value)
+        try:
+            self._setConfiguration(evt_value.rvalue)
+        except:
```
LJBD: The same thing here. Also probably `AttributeError`.

src/sardana/taurus/core/tango/sardana/pool.py
```diff
@@ -1890,10 +1909,20 @@ def on_elements_changed(self, evt_src, evt_type, evt_value):
         elif evt_type not in CHANGE_EVT_TYPES:
             return
         try:
-            elems = CodecFactory().decode(evt_value.value, ensure_ascii=True)
+            try:
+                elems = CodecFactory().decode(evt_value.rvalue,
+                                              ensure_ascii=True)
+            except:
```
LJBD: The same thing here. Also probably `AttributeError`.

src/sardana/taurus/core/tango/sardana/pool.py
```diff
@@ -1890,10 +1909,20 @@ def on_elements_changed(self, evt_src, evt_type, evt_value):
...
+                # Fallback for Taurus3
+                elems = CodecFactory().decode(evt_value.value,
+                                              ensure_ascii=True)
         except:
             self.error("Could not decode element info")
-            self.info("value: '%s'", evt_value.value)
+            try:
+                self.info("value: '%s'", evt_value.rvalue)
+            except:
```
LJBD: The same thing here. Also probably `AttributeError`.

src/sardana/taurus/core/tango/sardana/macroserver.py
```diff
@@ -94,7 +95,11 @@ def eventReceived(self, src, type, evt_value):
             self.fireEvent(None)
         elif type != TaurusEventType.Config:
             if evt_value:
-                self.fireEvent(evt_value.value)
+                if taurus_version.split(".")[0] == '4':
```
cpascual: This would fail for "Taurus 5.0.0"
Use `>=` instead (and cast to int)

... or even better, use:

```python
from distutils.version import LooseVersion
if LooseVersion(taurus_version) >= '4':
    ...
```

src/sardana/taurus/core/tango/sardana/macroserver.py
```diff
@@ -120,14 +125,25 @@ def clearLogBuffer(self):
 
     def eventReceived(self, src, type, evt_value):
         if type == TaurusEventType.Change:
-            if evt_value is None or evt_value.value is None:
-                self.fireEvent(None)
+            if taurus_version.split(".")[0] == '4':
```
cpascual: same as above...

src/sardana/taurus/core/tango/sardana/macroserver.py
```diff
@@ -570,12 +590,11 @@ def _runMacro(self, xml, synch=False):
         return result
 
     def stateChanged(self, s, t, v):
-        self._old_door_state = self.getState()
-        try:
-            self._old_sw_door_state = self.getSWState()
-        except:
-            # TODO: For Taurus 4 compatibility
+        if taurus_version.split(".")[0] == '4':
```
cpascual: same...

cpascual:
As a general comment: this PR changes return values of `.value` by `.rvalue`.
Have you checked that this works? 
Note that `.rvalue` may return a Quantity object when `.value` returns a numeric type, so the methods receiving this should be able to handle the Quantities correctly. 

A more conservative approach would be to return `.rvalue.magnitude` *if it is a Quantity*, but this would lead to very ugly code all over Sardana and would make it difficult to later on properly adapt to Taurus 4 (that is why we originally decided to keep Sardana using "Taurus 3" style until the moment we can drop support of Taurus3).

And this brings an even more general issue with this PR:
I am personally not sure that this PR is the best approach. IMHO the forks in the code for taurus3-taurus4 compatibility should be kept at a minimum in order to facilitate the eventual cleaning of taurus3 inherited code.

My point is that being flooded with deprecation warnings is *a feature*, not a bug.

But if the current warnings are too annoying and actually degrade usability, note that a crude mechanism already exists for silencing repeated Taurus3 deprecation warnings (it was done as a quick hack and there is a lot of room for improvement). I propose to put some effort in improving it (maybe even allowing to silence *all* deprecation warnings on-demand)

cpascual:
> But if the current warnings are too annoying and actually degrade usability, note that a crude mechanism already exists for silencing repeated Taurus3 deprecation warnings (it was done as a quick hack and there is a lot of room for improvement). I propose to put some effort in improving it (maybe even allowing to silence all deprecation warnings on-demand)

Just for completitude, I checked with mrosanes and the current mechanism is enough for what we want. the usage is easy. An application using taurus (e.g., spock) can do the following to silence all warnings and only give a final info message:

```python
# as soon as possible:
from taurus import tauruscustomsettings
tauruscustomsettings._MAX_DEPRECATIONS_LOGGED = 0

# do whatever
...

# just before exit:
from taurus.core.util.log import _DEPRECATION_COUNT
from taurus import info
info('\n*********************\n%s', _DEPRECATION_COUNT.pretty())

```

mrosanes:
After discussion with @cpascual, we decided to change the approach in order to allow choosing how many deprecation warnings related to Taurus are displayed by setting the variable tauruscustomsettings._MAX_DEPRECATIONS_LOGGED.

This pull request will be closed, and a new one has been opened to deal with it:
https://github.com/sardana-org/sardana/pull/550

PR #541, opened 2017-08-09
Add relctrl macro

MikeFalowski:
From sardana workshop.

### DIFF
--- a/src/sardana/macroserver/macros/expert.py
+++ b/src/sardana/macroserver/macros/expert.py
@@ -28,7 +28,7 @@ from __future__ import print_function
 __docformat__ = 'restructuredtext'
 
 __all__ = ["addmaclib", "commit_ctrllib", "defctrl", "defelem", "defm",
-           "defmeas", "edctrl", "edctrllib", "prdef", "relctrlclass",
+           "defmeas", "edctrlclass", "edctrllib", "prdef", "relctrlclass",
            "rellib", "relmac", "relmaclib", "send2ctrl", "udefctrl",
            "udefelem", "udefmeas", "sar_info"]
 
@@ -220,7 +220,7 @@ class send2ctrl(Macro):
 ##########################################################################
 
 
-class edctrl(Macro):
+class edctrlclass(Macro):
     """Returns the contents of the library file which contains the given
        controller code."""
 
--- a/src/sardana/macroserver/macros/expert.py
+++ b/src/sardana/macroserver/macros/expert.py
@@ -27,10 +27,10 @@ from __future__ import print_function
 
 __docformat__ = 'restructuredtext'
 
-__all__ = ["addmaclib", "commit_ctrllib", "defctrl", "defelem", "defm",
-           "defmeas", "edctrlclass", "edctrllib", "prdef", "relctrlclass",
-           "relctrllib", "rellib", "relmac", "relmaclib", "send2ctrl",
-           "udefctrl", "udefelem", "udefmeas", "sar_info"]
+__all__ = ["addctrllib", "addmaclib", "commit_ctrllib", "defctrl", "defelem",
+           "defm", "defmeas", "edctrlclass", "edctrllib", "prdef",
+           "relctrlclass", "relctrllib", "rellib", "relmac", "relmaclib",
+           "send2ctrl", "udefctrl", "udefelem", "udefmeas", "sar_info"]
 
 import sys
 import traceback
@@ -318,6 +318,19 @@ class relctrllib(Macro):
     def run(self, ctrl_library):
         pool = ctrl_library.getPoolObj()
         pool.ReloadControllerLib(ctrl_library.name)
+
+
+class addctrllib(Macro):
+    """Adds the given controller library code to the pool server filesystem.
+    """
+
+    param_def = [["ctrl_library_name", Type.String, None,
+                  "The module name to be loaded (without extension)"]]
+
+    def run(self, ctrl_library_name):
+        # TODO: make it compatible with multiple pools
+        pool = self.getPools()[0]
+        pool.ReloadControllerLib(ctrl_library_name)
 
 
 class relctrlclass(Macro):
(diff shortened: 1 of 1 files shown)


reszelaz:
Thanks for this contribution! This macro will be certainly useful.
Please correct the [flake8 errors](https://travis-ci.org/sardana-org/sardana/jobs/262678108) and add documentation to the macro (you can get inspired on the other reload-type macros).

reszelaz:
@MikeFalowski I hope you don't mind I've fixed these tiny issues. Let's hope now we could accelerate the integration of this macro in Sardana.

Before integrating it, I would like to hear your opinions about the macro name. The proposed name is `relctrl`.
We could also consider `relctrlcls` or `relctrlclass` since what we actually reload is the controller class.

My doubt comes from the fact that the user perceive a different hirerarchy of objects in case of controllers and macros. This is because the macro instances are silently created and destroyed by Sardana on the macro execution request. While the controller instances are rather long-lived objects created and destroyed explicitelly by the user, for example using the `defctrl` macro. I don't know if it would not be confusing to use the `relctrl` macro that would actually reload the controller class.

I've tried to grouped the macro names for interacting with macros and controllers in form of a table (I've used edctrlcls, relctrlcls, lsctrlcls names in the table, but I'm totally opened for other propsals):

Controller related macros | Macro related macros
---|---
lsctrl | -
defctrl | -
udefctrl | -
send2ctrl | -
edctrlcls [1] | edmac
edctrllib | edmaclib
commit_ctrllib | -
addctrllib [2] | addmaclib
relctrlcls [3] | relmac
relctrllib | relmaclib
lsctrlcls [4] | lsmac
lsctrllib | lsmaclib

[1] This macro already exists with name edctrl but is buggy (and I think was always buggy..), so I don't think that we would need to maintain the backwards compatibility if we decide to rename it to edctrlcls.
[2] This macro does not exist yet but we would certainly need it.
[3] This is the new macro that this PR adds.
[3] This macro does not exists and we could add it.

jordiandreu:
I have no preference yet for the naming convention but, let me put some more noise to the discussion with the aim to shed some light on the issue.

If relmac goes with relmaclib...
...then relctrl shall be paired with relctrllib...or if
...relctrlcls shall be paired with relctrlclslib.., but not mixed?

teresanunez:
The point is that macros and controllers are not 'equivalent', as Zibi says above. The naming that
Zibi proposes is coherent with what is already use for the Pool attributes:

ControllerList
ControllerClassList
ControllerLibList

I think relctrl is not right. relcrtlcls and relctrlclslib is IMO more precise than relctrlcls and relctrlib, but, well, I think is not necessary to add the cls to rectrllib ... and in any case it is compatible with the Pool attributes. Only one opinion, no problems with any decission.

mrosanes:
Another idea, but maybe it is not really feasible:

- For macros: merge relmac and relmaclib into a single macro: relmac (loading a given macro, could also load the libraries associated and needed by the given macro)

- For controllers: merge relctrlcls, relctrlclslib, etc, into a single macro: relctrlcls.

teresanunez:
Well, actually relmac reloads the whole library, the only difference was the argument
 ... so perhaps keep only relmaclib ??? ... the problem could be that the people doesn;t know
the name of the library ...

mrosanes:
You have reason @teresanunez; or for example, have a single macro: relmaclib, that could accept as argument, the name of the macro and the name of the library, and both would work.

My comment was not correctly specified, it was going more in the direction to merge rellib (which loads libraries used by the macros, but not containing macros), with relmaclib and relmac (which does contain macros). In this case only a single macro would remain, let's say 'relmaclib', which would load the macro library and the libraries used by the given macro library. But it is only an idea.

reszelaz:
First of all, thanks to all for your comments!

@teresanunez has a very good point with the attribute names on the server side.
I will use table to correlate the server API with the macro names.

So in case of the MacroServer and macro related macros we have:

MacroServer API | Macro name
---|---
MacroList attribute | lsmac
MacroLibList attribute | lsmaclib
ReloadMacro command | relmac
ReloadMacroLib command | relmaclib

So in case of the Pool and controller related macros we have:

Pool API | Macro name
---|---
ControllerList attribute | lsctrl
ControllerClassList attribute | lsctrlcls [*]
ControllerLibList attribute | lsctrllib
ReloadControllerClass command | relctrlcls [*]
ReloadControllerLib command | relctrllib [*]

[*] new macros - I used names proposen in the previous table

Regarding the first proposal of @mrosanes, to merge relmac and relmaclib into one macro relmac and relctrlcls and relctrllib into another macro relctrlcls, from the end user point of view it would be enough. As @teresanunez explains later, the relmac macro already reloads the whole library. I think this is a bug, and I hope that at some point we would be able to reload just a single macro. So, in the current situation, the correct thing to do would be to merge it into relmaclib. But here, I agree with @teresanunez that the users not necessarily know the macro library name. So finally, I would leave two macros as it is now. Furthermore, I remember to use many times the relmalib macro. So, IMHO, it could be useful to the engineers.

Regarding the second proposal of @mrosanes, to accept two different types of arguments for relmaclib: MacroCode and MacroLibrary, this will be difficult to implement with the present features of MacroServer. The only thing that would be possible is to define this argument as string type. But this way we would lose the context autocompletion in spock and a special combobox in the macroexecution widgets. 

Regarding the third proposal of @mrosanes, to merge relmac and rellib into one macro relmac and relmaclib and rellib into one macro relmaclib, I like this idea. But I would not eliminate the rellib macro and do it as an extra parameter to the relmac and relmaclib that would be by default False. So, if the user would like to reload a macro “my_experiment” and all the internally used python libraries, he would need to execute "relmac myexperiment True".
We should keep in mind that this operation may affect another macros that use the same python libraries. So by default I would not do that, just for the security reasons. I think the rellib macro should be called explicitly cause it may have bigger impact on the whole system. Btw. I think it would be tricky to implement this proposal.

reszelaz:
After recompiling all the comments I think that the best would be use the names proposed in https://github.com/sardana-org/sardana/pull/541#issuecomment-331932759. I have renamed the relctrl to relctrl and edctrl to edctrlcls and added relctrllib and addctrllib macros. Thanks to all for your contribution!

PR #545, opened 2017-08-11
Hotfix 2.3.2

reszelaz:
Fix provides metadata to be compliant with the PyPI (now it is more strict).


reszelaz:
I merge it directly cause it affects only the project metadata and is not critical to the software operation.

PR #548, opened 2017-08-16
Allow to stop a Macro when it is paused

mrosanes:
Allow to stop a Macro (by aborting it) when this macro is paused.


teresanunez:
I have reviewed this and it works fine.

reszelaz:
I've also checked and it looks ok. I'm merging it. Thanks to both!

PR #549, opened 2017-08-16
Fix property macro_status

mrosanes:


reszelaz:
Yes, this is a copy&paste error. I'm merging it. Thanks!

PR #551, opened 2017-08-18
Issue283

mrosanes:
Add option in sequencer for providing input file
    
Add option in macro sequencer for providing a macro sequence xml
input file.
    
A sequence of macros can be loaded from the command line execution of
the sequencer using the option: --file (or -f).
    
Example:
sequencer --file foo.xml

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py
@@ -578,21 +578,7 @@ class TaurusSequencerWidget(TaurusWidget):
         self.saveSequenceAction.setEnabled(False)
         self.emit(Qt.SIGNAL("currentMacroChanged"), None)
 
-    def onOpenSequence(self):
-        if not self._sequenceModel.isEmpty():
-            if Qt.QMessageBox.question(self,
-                                       "Open sequence",
-                                       "Do you want to save existing sequence?",
-                                       Qt.QMessageBox.Yes,
-                                       Qt.QMessageBox.No) == Qt.QMessageBox.Yes:
-                self.onSaveSequence()
-                self.tree.clearTree()
-
-        sequencesPath = self.sequencesPath()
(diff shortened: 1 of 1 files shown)


reszelaz:
@mrosanes I agree on the changes and I also tested it. It works nicelly. Could you please correct the PEP8 errors. Afterwards, I will integrate it immediatelly. Thanks!

mrosanes:
Done for the extra indentation in order to fix PEP8.
However PEP8 also accepts to not adding this extra indentation:

https://www.python.org/dev/peps/pep-0008/#indentation:

<pre>
When the conditional part of an if-statement is long enough to require that it be written across multiple 
lines, it's worth noting that the combination of a two character keyword (i.e. if), plus a single space, plus 
an opening parenthesis creates a natural 4-space indent for the subsequent lines of the multiline 
conditional. This can produce a visual conflict with the indented suite of code nested inside the 
if-statement, which would also naturally be indented to 4 spaces. This PEP takes no explicit position on 
how (or whether) to further visually distinguish such conditional lines from the nested suite inside the 
if-statement. Acceptable options in this situation include, but are not limited to:

# No extra indentation.
if (this_is_one_thing and
    that_is_another_thing):
    do_something()

# Add a comment, which will provide some distinction in editors
# supporting syntax highlighting.
if (this_is_one_thing and
    that_is_another_thing):
    # Since both conditions are true, we can frobnicate.
    do_something()

# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
</pre>

mrosanes:
Maybe, according to the last comment on PEP8, we could make flake8 more permissive in order to accept to do not add extra indentation between a long if statement broken in two line, and the block of code inside the if.

reszelaz:
Thanks for correcting the PEP8 error. Now, I proceed with the integration. As you suggested I have proposed to relax the E129 error on the flake8 check done by travis-ci.

droldan:
Good news, it will be very useful for the scientist on try to relaunch the sequencer remotely! 
Thanks @mrosanes and @reszelaz!

PR #552, opened 2017-08-21
Report error number when loading sequence from txt

mrosanes:
The sequencer allows to import text files with macro sequences.

If a macro from the sequence in the text file contains an error,
report the line where this error is found.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py
@@ -862,10 +862,9 @@ class TaurusSequencerWidget(TaurusWidget):
         error_line = 0
         for macroNode in newRoot.allMacros():
             error_line += 1
-
             try:
-                macroServerObj.recreateMacroNodeAndFillAdditionalInfos(macroNode)
-
+                macroServerObj.recreateMacroNodeAndFillAdditionalInfos(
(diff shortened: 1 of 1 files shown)


teresanunez:
I have reviewed this. It works as described, but I see a problem with macros that do not
need arguments and are put in the text file without them. For example, a sequence with
the macro lsm, if I put in the the text file to load:

mv mot01 10
lsm

the sequence is loaded but when I try to run it I get the error:

'nothing to repeat'

The right way to put it in the text file is:

lsm .*

but I don't think that this is easy to see for the user. 
Do you also see this ?.
I would say that if the macro is loaded as:

lsm
 
it should run, and if it does not run an error should appear when loading ... but
 indicating something about putting 

lsm .* 

in the text file, if not it could be dificult to figure out.

reszelaz:
I agree with @teresanunez. It should be possible to run lsm without the need to explicitly pass parameters. `.*` is the default value.

mrosanes:
I agree with both of you. But in my opinion this is another bug. It already exist in develop without the integration of this pull request and I see it as an independent issue.
@teresanunez, if you agree with that, could you report it?

teresanunez:
Hi @mrosanes, perfect, I will create an issue with that problem. Then IMO this PR could be merged.

reszelaz:
Great! @teresanunez could you please proceed with the merge then.

mrosanes:
Thanks for the integration @teresanunez!

PR #557, opened 2017-08-23
Fix Door method read_ElementList

mrosanes:
Get the MacroServer elements when executing the Door method
read_ElementList.

### DIFF
--- a/src/sardana/tango/macroserver/Door.py
+++ b/src/sardana/tango/macroserver/Door.py
@@ -325,10 +325,9 @@ class Door(SardanaDevice):
         value = attr.get_write_value()
         self.door.get_input_handler().input_received(value)
 
-    #@DebugIt()
     def read_ElementList(self, attr):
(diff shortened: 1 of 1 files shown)


mrosanes:
This pull request deals with issue https://github.com/sardana-org/sardana/pull/557

reszelaz:
Thanks for this contribution, but I would change a little bit the strategy. XXXList attributes are more oriented to exposing the user friendly information in form of strings, see Pool's InstrumentList or MotorList. The Elements attributes are foreseen for exposing all elements to the clients in form of encoded stream of data e.g. Pool's elements, MacroServer elements. I propose to eliminate this attribute at the Door level since we already have it on the MacroServer level.   Any other opinions?

cmft:
IMO clean up the API and  avoid  multiples way to do the same thing  is  better. So, I will remove it.

I created a PR #653 to do that.

reszelaz:
#653 removes this attribute. Closing this PR.

PR #558, opened 2017-08-23
Issue221: Present the door list in an ordered fashion

mrosanes:
Present the door list in an ordered fashion

When creating a new spock profile, present the door list
in an ordered way. Done for ipython >= 1.0.

- First: the doors containing aliases are presented in
alphabetical order.

- Second: the doors that does not contain an alias are presented
at the end of the list also ordered alphabetically.


teresanunez:
Hi, I have reviewed this and it works as described. IMO it can be merged.

reszelaz:
Nice! I also think that this could be merged.

PR #559, opened 2017-08-24
Reimplement macrobutton handleEvent

mrosanes:
Macrobutton text is set to the Door name if MacroServer
is stopped. This is a not desired behavior, and the text
of the macrobutton should remain the same.

Reimplement macrobutton handleEvent in order to avoid text
setting when MacroServer is down, or when macroserver is
reconnected.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/macrobutton.py
@@ -90,7 +90,7 @@ class MacroButton(TaurusWidget):
 
     # Override default implementation of handleEvent from TaurusWidget
     # in order to avoid button's text being lost on the MS restart.
-    # More detais in #293 and taurus-org/taurus#635 
+    # More detais in #293 and taurus-org/taurus#635
     def handleEvent(self, evt_src, evt_type, evt_value):
         pass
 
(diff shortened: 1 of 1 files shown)


cmft:
IMO the bug is in taurus, and how is implemented the handleEvent in (taurusbasecontainer).  It does not respect an hard-code set text. 

In taurus 4 the code is not triggered due to an exception in the push_event.

```
push_event generated the following python exception:
  File "/home/cfalcon/tools/taurus.git/lib/taurus/core/tango/tangoattribute.py", line 710, in push_event
    self.fireEvent(etype, evalue, listeners=listeners)
  File "/home/cfalcon/tools/taurus.git/lib/taurus/core/taurusmodel.py", line 291, in fireEvent
    l(self, event_type, event_value)
  File "/home/cfalcon/tools/sardana-sardana.git/src/sardana/taurus/core/tango/sardana/macroserver.py", line 573, in stateChanged
    self._old_door_state = self.getState()
  File "/home/cfalcon/tools/taurus.git/lib/taurus/core/util/log.py", line 1047, in new_func
    return func(*args, **kwargs)
  File "/home/cfalcon/tools/taurus.git/lib/taurus/core/tango/tangodevice.py", line 115, in getState
    stateAttrValue = self.stateObj.read(cache=cache)
  File "/home/cfalcon/tools/taurus.git/lib/taurus/core/tango/tangoattribute.py", line 503, in read
    raise self.__attr_err
PyTango.DevFailed: DevFailed[
DevError[
    desc = Event channel is not responding anymore, maybe the server or event system is down
  origin = EventConsumer::KeepAliveThread()
  reason = API_EventTimeout
severity = ERR]
]
```

IMO the workaround solve the problem and do not break nothing. So, it can be applied.

reszelaz:
I also agree with @mrosanes and @cmft. Just added an explanatory comment in the code. Merging it...

PR #560, opened 2017-08-25
Check motors limits before motor group movement

daneos:
Fix for bug #259 (SF#414).
Before starting group movement I check the limits for each motor and raise an exception if requested position is outside.

### DIFF
--- a/src/sardana/pool/poolmotorgroup.py
+++ b/src/sardana/pool/poolmotorgroup.py
@@ -267,12 +267,14 @@ class PoolMotorGroup(PoolGroupElement):
                 low = None
             if high is not None:
                 if float(new_position) > high:
-                    raise RuntimeError("%s: Requested movement above "
-                                       "upper limit." % element.name)
+                    msg = "requested movement of %s is above its upper limit"\
+                        % element.name
+                    raise RuntimeError(msg)
             if low is not None:
                 if float(new_position) < low:
-                    raise RuntimeError("%s: Requested movement below "
-                                       "lower limit." % element.name)
+                    msg = "requested movement of %s is below its lower limit"\
+                        % element.name
+                    raise RuntimeError(msg)
 
         for new_position, element in zip(new_positions, user_elements):
             element.calculate_motion(new_position, items=items,
--- a/src/sardana/pool/poolmotorgroup.py
+++ b/src/sardana/pool/poolmotorgroup.py
@@ -38,6 +38,8 @@ from sardana.sardanaattribute import SardanaAttribute
 from sardana.pool.poolgroupelement import PoolGroupElement
 from sardana.pool.poolmotion import PoolMotion
 
+from PyTango import AttributeProxy
+
 
 class Position(SardanaAttribute):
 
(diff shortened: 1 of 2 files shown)


daneos:
I'm calling for help here, since I run the test suite locally, and all tests are passing. I also can't find any relationship between my changes and tests that are failing.

cmft:
Hi @daneos ,
I don't think so travis is broken due to your code . The test seems to be broken in develop. I can reproduce it :
``` 
 python -m unittest sardana.macroserver.macros.test.test_scan.MeshTest
```
A similar patch was proposed some time ago and It was rejected. When @reszelaz come back next week we'll study the pros/cons of this patch.

reszelaz:
@daneos, I have tested your implementation and it works pretty well. However, I don't think that this will be the definitive solution for the limit problems as described in #259 and #36. The good thing is that it already covers many cases. I think that the right thing to do is to implement the verification on the server side but it is a bigger enhancement - currently there is nothing implemented apart from an empty class `SardanaAttributeConfiguration`.

Before we integrate this PR, I would add a TODO next to this implementation that says something like "limits verification should be done on the server side". Regarding the failing tests, could you please try to merge the latest develop to your branch so it will retrigger the tests with the most recent changes. If it does not help I suggest that you debug this problem by adding more traces to the test. Actually, I'm just doing something very similar in https://github.com/sardana-org/sardana/pull/612#issuecomment-332796331. So you can get some ideas from there. If you need further help, please let us know.

As the last thing, as @cmft said, there was a similar implementation that was proposed in https://github.com/sardana-org/sardana/issues/259#issuecomment-262976057. I have not done functional tests, but from the code review I think it won't work.

daneos:
@reszelaz, this PR aims only to fix #259 and not #36 (however I'd like to fix that too, when I'll have some spare time). I don't really know how it should be done the correct way, so I assumed that `MotorGroup` implementation is a good place to start.

I have tested the solution in [#259 (comment)](https://github.com/sardana-org/sardana/issues/259#issuecomment-262976057) and it does not work, it will set the limits from the motor that have the lowest limits set (for example, if you have a motor with a limit on 20mm and second one with a limit on 100mm then both of them will be limited to 20mm in group movement), and that's why I decided fix it myself.

I've added the TODO, and merged latest develop, but this did not fix my problems with tests, I'll do some more trobleshooting as you described.

reszelaz:
> @reszelaz, this PR aims only to fix #259 and not #36 (however I'd like to fix that too, when I'll have some spare time). I don't really know how it should be done the correct way, so I assumed that MotorGroup implementation is a good place to start.

IMHO the correct place to implement the protection in on the server side i.e. Pool. First of all, if we implement the limit protection logic in the taurus extension, another clients requesting the movement e.g. Jive or a python script using PyTango directly, would not be protected. Another point is that in case of pseudo motors we will need to accessing the hirerarchy of underneeth elements, these can be direclty physical motors but in some more advanced applications another pseudo motors, etc. I think that this should be easier to implement inside of the Pool where they are related by the publisher - subscriber chain. Until we properly implement it in the Pool, it woudl be great if you voluteer to implement a simplified version of this protection for the pseudo motors on the taurus extension level. I suggest to just treat the first level of underneeth elements. I imagine that this new implementation for pseudo motors would be very similar to the one for motor group

> I have tested the solution in #259 (comment) and it does not work, it will set the limits from the motor that have the lowest limits set (for example, if you have a motor with a limit on 20mm and second one with a limit on 100mm then both of them will be limited to 20mm in group movement), and that's why I decided fix it myself.

Thanks, then it confirms my suspicions.

> I've added the TODO, and merged latest develop, but this did not fix my problems with tests, I'll do some more trobleshooting as you described.

Thanks! 
If the affected test is the one running macro on the demo1 MacroServer, I suggest to extract the logs and print them in the test, something similar that was done in https://github.com/sardana-org/sardana/pull/612#issuecomment-332796331:

`taurus.warning("CommonBuffer: %s" % self.macro_executor.getCommonBuffer())`

teresanunez:
I have tested the new implementation and it works as expected.

daneos:
@teresanunez, that's great! I used your solution, since I think it's better than my old one.
Now we just need to get rid of the AttributeProxy.

reszelaz:
When comming back to this PR I just realized that it is *green* now. Maybe some of the last commits have fixed it :). Great!

reszelaz:
As agreed on the last follow-up meeting we will include this fix in the Jan18 release. Note that the proposed implementation, which uses the `AttributeProxy`, is more like a workaround. We should try to get rid of it as soon as possible, especially taking into accout this https://github.com/sardana-org/sardana/pull/663#issuecomment-363451003.

I'll merge this PR as soon as travis-ci finish its job and I will close #259 (a TODO about the workaround is already in the code). Many thanks to all involved in fixing this problem!

Issue #377, opened 2016-12-01
Bug on relmaclib


rhomspuron:
It you try to reload a library of macros and there is a macro with a wrong definition of the parameters, the macro relmaclib does not show the error, but on the MacroServer log you can see it. 

Macro code:
<pre>
class dynaGO(Macro):
       param_def = [
      ['temperature', Type.Float,None,],
      ['temperature_old', Type.Float,None,]
        ]

     def run(self, temperature, temperature_old):
       temp = temperature
       self.execMacro('dynRamp %f' %temp)
     ...
</pre>

MacroServer Log:
<pre>
SardanaTP.W005 DEBUG    2016-12-01 16:50:11,880 MS_BL22.MacroManager: Adding macro class dynaGO
SardanaTP.W005 ERROR    2016-12-01 16:50:11,882 MS_BL22.MacroManager: Error adding macro dynaGO
SardanaTP.W005 DEBUG    2016-12-01 16:50:11,882 MS_BL22.MacroManager: Details:
Traceback (most recent call last):
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmacromanager.py", line 502, in reloadMacroLib
    self.addMacro(macro_lib, macro)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmacromanager.py", line 514, in addMacro
    return add(macro_lib, macro)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmacromanager.py", line 523, in addMacroClass
    macro_class = MacroClass(**params)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmetamacro.py", line 238, in __init__
    Parameterizable.__init__(self)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmetamacro.py", line 118, in __init__
    self._parameter = self.build_parameter()
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmetamacro.py", line 137, in build_parameter
    return self._build_parameter(self.get_parameter_definition())
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/msmetamacro.py", line 168, in _build_parameter
    ret_p['description'] = p[3]
IndexError: list index out of range
SardanaTP.W005 DEBUG    2016-12-01 16:50:11,882 MS_BL22.MacroManager: Adding macro class dynaVtuneClaess
[...]

PR #577, opened 2017-08-30
Report error if macro parameter definition is wrong

mrosanes:
Report error in spock if macro could not be loaded
using relmaclib because of a wrong parameter definition
exists in the macro code.

In this case, the macro will not be available, and
the macro library shall be reloaded using
addmaclib after correcting the error in the parameter
definition of the affected macro.

### DIFF
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -493,6 +493,7 @@ class MacroManager(MacroServerManager):
                           macro_server=self.macro_server, exc_info=exc_info)
             return MacroLibrary(**params)
 
+
         # if there was previous Macro Library info remove it
         old_macro_lib = self._modules.pop(module_name, None)
         if old_macro_lib is not None:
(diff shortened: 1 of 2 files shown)


mrosanes:
This pull request deals with the issue https://github.com/sardana-org/sardana/issues/377
When a macro has an error in the parameter definition, spock was not reporting anything.
Now spock is reporting it. The macro affected by the wrong parameter definition is not available anymore.
The rest of macros of the same macro library are still available.

However, in the case of a wrong parameter definition, the macro library has to be loaded using addmaclib, once the parameter definition is corrected.

teresanunez:
Hi @mrosanes, I was trying this ... but I do not get any macro in spock working with your code. They are in the MacroServer but not in spock ... It could be that I am making something wrong ...

cmft:
Hi @mrosanes ,
The error are well reported in `spock` but I can not see the macros (old implementation) anymore, they are not in spock or MS.

cmft:
I didn't see any macro because in my case the wrong macro was the first to be loaded

mrosanes:
A new pull request is proposed to solve the error reported by @teresanunez and @cmft.
https://github.com/sardana-org/sardana/pull/642

Could you please test the new implementation?

mrosanes:
A new pull request #642 is proposed to solve issue https://github.com/sardana-org/sardana/issues/377.

PR #579, opened 2017-08-30
Fix error with case sensitivity

mrosanes:
Sequencer allows to load from xml and from txt, macros that
contains syntax errors related with the case sensitivity. The
macros are loaded into the sequencer, but they cannot be executed,
because macros are case sensitive.

Do not allow to load macros with syntax errors related with
the case sensitivity. Pop up a message if the syntax is not
correct.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/sardana.py
+++ b/src/sardana/taurus/core/tango/sardana/sardana.py
@@ -201,22 +201,20 @@ class BaseSardanaElementContainer:
         return self.getElement(elem_name) is not None
 
     def getElement(self, elem_name):
-        elem_name = elem_name.lower()
         for elems in self._type_elems_dict.values():
             elem = elems.get(elem_name)  # full_name?
             if elem is not None:
                 return elem
             for elem in elems.values():
-                if elem.name.lower() == elem_name:
+                if elem.name == elem_name:
                     return elem
 
     def getElementWithInterface(self, elem_name, interface):
-        elem_name = elem_name.lower()
         elems = self._interfaces_dict.get(interface, {})
         if elem_name in elems:
             return elems[elem_name]
         for elem in elems.values():
-            if elem.name.lower() == elem_name:
+            if elem.name == elem_name:
                 return elem
 
     def getElements(self):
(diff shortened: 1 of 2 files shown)


teresanunez:
I have reviewed this and it solves the case sensitivity problem. I can merge it if nobody is against.

reszelaz:
Thanks @mrosanes and @teresanunez!

The `BaseSardanaElementContainer` class is a base class used by the Pool and MacroServer taurus extensions. So, changing it may affect a bigger scope than just the macros. According to the [documentation](http://www.sardana-controls.org/en/latest/devel/howto_macros/macros_general.html#what-is-a-macro), macros are case sensitive in contrary to the other Sardana elements like for example, motors, experimental channels, etc.

Currently, case sensitivity of macros is not well treated in Sardana. For example, if you create two macros: `macro1` and `Macro1`,  the `lsm` macro output will show you only one of them but it is possible to edit both of them with `edmac`.

So, if you don't mind I would wait with integrating this PR until we clarify well where are the failing points that treats the macros case sensitivity.

teresanunez:
HI @reszelaz, thanks for the info. I agree with you, the treatment of the case sensitivity in the macros has to be consistent everywhere.

mrosanes:
I would propose to do the macros case insensitive, to be in accordance with the rest of elements of Sardana. We are using the same code in BaseSardanaElementContainer for macros and for other elements of Sardana (like motors, etc), but macros are case sensitive and other elements are not. Maybe generalizing the case insensitiveness in macros as well, would be a nice possibility.

cpascual:
Case insensitivity in motors expchans, etc is IMHO, an *unwanted* limitation due to these elements being mediated by Tango DSs. 
Therefore I think that we should *not* go into that direction for the elements that are not affected by this limitation.
Actually, to me the case insensitivity for other elements should be registered and documented as a known bug (even if we are probably never going to fix it).

mrosanes:
After thinking on it, I do not see the case insensitivity of motors and other elements as a bug. I would not expect 'Mot1' being different than 'MOT1' or 'mot1'. We have the Tango constraint, but in my opinion the case insensitiveness is well chosen for sardana purposes; even if Python is case sensitive. But this is only my opinion.

PR #580, opened 2017-08-30
[WIP] Issue578: Change nan value on spec file.

rhomspuron:
Implement the possibility to change the nan value for another value defined by the user
on spock. It uses the environment variable SpecNanValue.

### DIFF
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -279,6 +279,16 @@ class SPEC_FileRecorder(BaseFileRecorder):
             self.setFileName(filename)
         self._expectedlabels = pars.get('labels', None)
 
+        # Configure nan value
+        self.nan_value = float('nan')
+
+        if macro is None:
+            try:
+                self.nan_value = float(macro.getEnv('SpecNanValue'))
+            except Exception as e:
+                macro.debug('SpecRecorder: Error in reading SpecNanValue. %r'
+                            % e)
+
     def setLabels(self, labels):
         self._expectedlabels = labels
 
@@ -446,7 +456,7 @@ class SPEC_FileRecorder(BaseFileRecorder):
     def _writeRecord(self, record):
         if self.filename is None:
             return
-        nan, names, fd = float('nan'), self.names, self.fd
+        nan, names, fd = self.nan_value, self.names, self.fd
 
         d = []
         for oned_name in self.oned_names:


cpascual:
Just a suggestion:
Maybe it would make sense to allow setting this via `sardanacustomsettings.py` as well

src/sardana/macroserver/recorders/storage.py
```diff
@@ -279,6 +279,16 @@ def __init__(self, filename=None, macro=None, **pars):
             self.setFileName(filename)
         self._expectedlabels = pars.get('labels', None)
 
+        # Configure nan value
+        self.nan_value = float('nan')
+
+        if macro is None:
```
cpascual: You probably mean "is not None"

cpascual:
Given that the RecordList object now supports interpolation/extrapolation, is this feature still needed?

rhomspuron:
We should implement the same solution for 1D channels. In addition if we activate the interpolation/extrapolation this solution will not work. 

The users prefer to have interpolation/extrapolation for some channels and not for all of them.

src/sardana/macroserver/recorders/storage.py
rhomspuron: Yes, your right

rhomspuron:
Close by mistake. I removed my remote branch

reszelaz:
I just discussed this with @rhomspuron and below are our conlcussions.

We investigated a little bit how this is handled in SPEC, and there is no such configuration parameter. One may however implement a *macro hardware* plugin (something similar to a controller plugin in sardana) to return a value reserved for such cases e.g. `-1`, -`999`, `0`. The same possiblity exists in Sardana as well.

For the case when a channel is not available e.g. a hardware is disconnected, an attribute is not readable, it is recommended to disable this channel in the measurement group or exclude it from the measurement group.

We think that a global parameter that would apply to all the channels may be not enough in the case when the user would like to specify a different value for each channel. So implementing a generic solution much beyond the changes proposed in this PR.

We should not forget that sardana also allows to plugin custom recorders, and easily one could inherit from the Sardana default SPEC recorder and catch the cases when `NaN` arrives and handle them differently, depending on the channel, etc.

Having all this in mind we propose to close this PR. If anyone has different opinion about this feel free to comment here and we will get back to this topic.

dschick:
I justed asked at xrayutilities for handling Nan values during reading the spec files. I actually like to have nans. So I agree that one might write a custom recorder for changing that if needed.

PR #582, opened 2017-08-30
Issue581: Return a None object on wrong reading

rhomspuron:
Add a None object on the acquisition data when the reading is wrong
or fails.


src/sardana/pool/poolacquisition.py
```diff
@@ -716,7 +716,7 @@ def action_loop(self):
                 if is_value_error(value):
                     self.error("Loop final read value error for: %s" %
                                acquirable.name)
-                    acquirable.put_value(value)
+                    acquirable.extend_value_buffer(None, propagate=2)
```
reszelaz: Your proposal here won't work because the `extend_value_buffer` code would try `len(None)` which would fail.
Could you clarify why you set the None only at the end of the action and not in the  main loop of the action?

Regardless of the above doubts, I would leave the code without changes in case of the hardware acquisition action. Intermediate exceptions in the read methods during the acquisition action does not precisely mean that the acquisition is failed. We base this decision on the channel's state. I can imagine a situation when one of the intermediate read method raises an exception, but then the subsequent executions correctly retrieve the data. The `put_value` call is there to indicate the error via the `Value` attribute.

src/sardana/pool/poolacquisition.py
```diff
@@ -796,7 +796,7 @@ def action_loop(self):
                 if is_value_error(value):
                     self.error("Loop final read value error for: %s" %
                                acquirable.name)
-                    acquirable.put_value(value)
+                    acquirable.append_value_buffer(None, self.index)
```
reszelaz: Since the `append_value_buffer` apart of sending the new value via the `Data` attribute also sets the `Value` attribute then your proposal raises an exception in the `on_ct_changed` method of the CTExpChannel device. I propose to append the value itself in both cases: error and not error. This way the error message is transferred to the clients, as it was done previously with the `put_value` call, but this way we also fill the buffer with a placeholder - `SardanaValue` with the error information. Please test and change the implementation accordingly.

rhomspuron:
I tested your solution and it works on the beamline. You can change the code. 
Thanks

reszelaz:
Just to keep in mind, the information in the MacroServer about the channel's value: whether the corresponding trigger was skipped or the acquisition ended with an error is preserved. The recorders receive the following data: `NaN` if skiped and `None` if error. But then it depend on the recorder how this infomration is used. For example the output recorder prints `nan` or `<nodata>`, but the spec recorder always writes `nan` to the file.

Issue #585, opened 2017-08-31
Alphabetical order when presenting Pools, in new MacroServer creation


mrosanes:
When creating a new MacroServer, the options for choosing the Pool to which the MacroServer will be linked, are presented without order. From the user point of view, it is difficult to find a given Pool if a lot of Pools exists in the same Tango Host. 

Moreover, once the Pool has been chosen, the sentence ```Please select the Pool to connect to (return to finish)``` appears again, and a new again the ```return``` key has to be pressed.

Present the Pools in alphabetical order at the moment of a new MacroServer creation. Avoid to repeat the question again, once the Pool has already been chosen.

mrosanes:
After speaking with @reszelaz, the question is needed two times, because the MacroServer can be connected to two or more Pools. Because of this reason the question was posted more than once.

Thus, it is only needed to present the MacroServers in alphabetical order, but the question must be posted more than onces.

mrosanes:
Solution has been integrated to develop by @teresanunez.

PR #586, opened 2017-08-31
Present Pool Servers by alphabetical order

mrosanes:
When creating a MacroServer, the Pool names are presented in
order to allow the user to choose one of them.

Present them by alphabetical order.

### DIFF
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -682,12 +682,9 @@ def prepare_server(args, tango_args):
                         print("\nAvailable Pools:")
                         for pool in sorted_pools:
                             print pool
-                        msg = "\nPlease select the Pool to connect to " \
-                              "(return to finish): "
-                    else:
-                        msg = "\nIf you wish to connect to more than one " \
-                              "Pool, \nplease select the new Pool to  " \
-                              "connect to (return to finish): "
+                        print("")
+                    msg = "Please select the Pool to connect to " \
+                          "(return to finish): "
                     elem = raw_input(msg).strip()
                     if not len(elem) and not pool_names:
                         print("\nMacroServer %s has not been connected to any "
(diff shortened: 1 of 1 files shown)


mrosanes:
This pull request, should solve issue https://github.com/sardana-org/sardana/issues/585

teresanunez:
Hi, I have checked it and it solves both issues presented in #585. In my opinion it could be merged.
I don't know if somebody else want to review it ( @reszelaz  ?), if not I could merge it.

reszelaz:
@teresanunez thanks for reviewing it! I have also checked it and the changes in the while loop made that it is not possible to connect to more than one Pool. @mrosanes could you come back to the previous behavior?

teresanunez:
Hi @reszelaz, yes, I realized about that, but I thought that it was your decission to change that (allowing
 connecting to only one Pool), I remeber to read about that in some comments.

mrosanes:
In fact, the question of being able to choosing a Pool being presented more than once was a feature (not a bug). Reallowed connection to more than one Pool. Added information during the process of choosing the Pools to which a given MacroServer is connected.

teresanunez:
Hi, I have checked this again ... One point that could discussed: after selecting one pool the list of available Pools is shown again, complete, also with the already selected pool(s), it a bit confusing or even too much if the list is too long ... one can lose the overview of the pools that are already added.
Perhaps one could avoid to show again the list or, if shown, exclude the already selected. Well, it is only an opinion.

mrosanes:
Yes, you have reason; it is too much information. We will keep the previous behavior and only print the Pool names once. I have done a new commit to deal with it.

teresanunez:
Hi @mrosanes, thanks for the change. It is ok for me. Sorry, only as comment, for me it can
 stay like it is: perhaps it is a bit confusing to repeat the message:

 'If you wish to connect to more than one Pool,'

every time you get the possibility to connect to a new Pool ... in fact only the first time
would be 'more than one' ... But, well, it is fine.

teresanunez:
Hi @mrosanes, thanks, for me it is fine now. I can merge it if nobody has anything against.

reszelaz:
It looks nice now. Thanks to both!

PR #588, opened 2017-09-04
Add optional extrapolation of missing records in continuous scans

reszelaz:
The very first points in the continuous acquisition e.g. ascanct may be omitted, for example, if the motor updates to the software synchronizer were not precise/fast enough. In this case the zero order interpolation does not help if we want to get rid of the NaN values.

Introduce a new concept: extrapolation to avoid NaN in the very first points of measurements. Also, add ApplyExtrapolation environment variable to control this feature.

### DIFF
--- a/doc/source/users/scan.rst
+++ b/doc/source/users/scan.rst
@@ -185,6 +185,19 @@ Scans are highly configurable using the environment variables
 
 Following variables are supported:
 
+**ApplyExtrapolation**
+    Enable/disable the extrapolation method to fill the missing parts of the
+    very first scan records in case the software synchronized acquisition could
+    not follow the pace. Can be used only with the continuous acquisition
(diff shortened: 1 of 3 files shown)


reszelaz:
This PR is ready to be reviewed, any volunteers among @sardana-org/integrators ?

To reproduce the `NaN` at the beginning of the scan using dummy elements I was forced to change the Pool's property `MotionLoop_StatesPerPosition` to 20. Afterwards running `ascanct mot01 0 10 10 0.1` was enough. Then setting the `ApplyExtrapolation` environment variable one can get rid of the `NaN`.

teresanunez:
Hi @reszelaz , I can have a look in the next days.

teresanunez:
Hi @reszelaz, I have tested and it works as expected. I merge it.

PR #590, opened 2017-09-06
Continuous scans reads moveables positions

reszelaz:
Currently the continuous scans of `ct` type, `ascanct` & co. generates theoretical moveables positions and these are used in the scan records. In order to have a read position of moveables, it is necessary to add an extra experimental channel that will read it. This can be either hardware position capture counter timer or a tango attribute counter timer. In any case, it is not user-friendly.

The aim of this PR is to provide moveables positions in continuous scans by reading them on triggers of the software synchronizer. In order to implement it, the position buffers were added to the motor and the pseudo motors. The rest of the mechanism of the data transfer and data handling is exactly the same as in case of the experimental channels. 

There is one pending thing to decide, the measurement group attributes names. Now the master moveable - the one that is used as the base of the software synchronization is configured with the `moveable` attribute. Temporarily I have added `moveables` attribute to pass the list of moveables that has to be read during the acquisition. These attribute names are not very distinguishable. I will think of better names and add an extra commit just before the integration to develop. Meanwhile, it is ready for testing and review.

Just keep in mind, that the precision of the software synchronizer in the position domain depends on the frequency of the position readouts in the motion action. This can be controlled by two properties of the Pool device: `MotionLoop_SleepTime` and `MotionLoop_StatesPerPosition`.  In the future, we should add an extra property on the moveable level to be able to change it locally.


teresanunez:
Hi @reszelaz, thanks for that implementation, our users would really want this. I have tried
 it (with dummy motors) and I get the same positions for two or three consecutive points, I can
 change this playing with the properties you mention (I have still not managed to get different
positions for all points, but well, I could invest more time on that for a real case), the point 
is that I think that it would be better if this could be optimized in a more automatic way 
(in any case I think that the parameters to play with should not be properties, one should be able 
to tune them without stopping/starting sardana). Well, perhaps with real motor one does not
 get repeated positions ... Do you also see this?.

reszelaz:
@teresanunez thanks for the feedback! I totally agree that we would need a better way of adjusting the position readout frequency. But I would keep it out of this PR scope. 

Just as a quick hint: do you use any latency time? If your acquisition is skipped, due to the previous one still in progress, the position readout would be also skipped. Another possibility is that you are doing too fast scan together with too high resolution (too small intervals) and the software synchronizer is not able to follow - software synchronizer works in the position domain.

On Tuesday we plan to test it at the beamline and compare the results to the step scan. We will keep you informed.

teresanunez:
Hi @reszelaz, many thnaks for your tips ... In fact both cases apply, if I add a latency time big enough
I get different positions for all points, also without latency time but increasing the acquisition time the
positions are different. It works nice, thanks again for the implementation. Yes, please, inform about the results of the tests in the beamline.

reszelaz:
Yesterday we have performed some tests at the beamline and we have discovered one more thing that may affect the precision of the software synchronizer that we did not have in mind before.

We know the software synchronized acquisitions may be skipped due to the previous acquisitions still in progress. This can be easily worked around by correctly tunning the `latency_time` parameter of the scan.
Another reason of skipping the software synchronized acquisitions may be too slow updates of the motor's position that feed the software synchronizer. If the updates arrive slower than the required scan resolution then the software synchronized acquisition won't be even intended. 

Up to know we thought that be correctly adjusting the `MotionLoop` parameters was enough to work around this. By testing this approach in the real system, where several pseudo motors and motor groups are defined on top of the master moveable, we discovered that the time necessary to handle them can not be ignored. This is due to the fact that the listeners of motor's position are executed synchronously.

So for example a pseudo motor calculation that takes 100 ms means that during this time we won't update the motor's position, and we won't be able to update the software synchronizer, and as a consequence, we may loose a software synchronized acquisition.

In the following days, I will think how could we deal with that.

reszelaz:
After tests at the beamline and comparison of the step and continuous scan synchronized by software, we could say that the results are comparable with the error below 1%. So are pretty good. The total scan time was reduced from 17 min to 6 min - it is still quite long because of the very unpredictable time necessary to execute a single count on a particular experimental channel controller - to be improved.

We have discovered that a complicated hierarchy of pseudo motors and motor groups on top of the moveable being scanned has a big impact on the synchronization precision. That's why I propose to start these investigations/developments: #604, #608. The above results were achieved by adjusting the `MotionLoop_StatesPerPosition` to 2 and applying local changes to not update pseudo motors and motor groups while scanning (there were 26 pseudo motors and 12 motor groups on top of the moveable used as synchronization master). Meanwhile, I vote for integrating this feature in develop cause it may be already of profit for setups not affected by these type of overheads.

reszelaz:
Merging after positive voices from @teresanunez.

reszelaz:
While this feature is highly desired it may have some side effects. I propose to revert its integration until we properly deal with cases mentioned in #643.

PR #591, opened 2017-09-07
Fix bug in theoretical position in ascanct & co.

reszelaz:
Theoretical positions are not correclty calculated. Fix it by changing the number of numpy.linspace argument.


teresanunez:
Hi @reszelaz, I reviewed this. Should it be integrated?.

reszelaz:
Thanks! Yes, go ahead!

PR #592, opened 2017-09-08
New Stop and Abort algorithms (issue 157)

guifrecuni:
Hi, this PR should fix issue #157


reszelaz:
Together with @mrosanes we have evaluated this PR. Below our comments:

We started evaluating this PR with the assumption that Sardana did not support the multi axes stop/abort before. This PR adds this feature.

**Guidance when developing Stoppable/Abortable controllers**

This PR requires adding an explicit guidelines on how to develop a new Starable/Abortable controller.
When developing a new controller, first question that would need to be answered is, if the hardware supports multi axes stop. Four possibilities exists:
1. Your hardware supports only multi axes stop.
2. Your hardware supports only a single axis stop.
3. Your hardware supports both single axis and multi axes stop.
4. Your hardware supports only stop of all axes.

In the first case you develop:
* `PreStopAll` - clear list of axes to be stopped
* `PreStopOne` - add axis to the list of axes to be stopped
* `StopAll` - send a command to stop axes aggregated in the list

In this case we see that `StopOne` is not necessary to be developed. Now it raises `NotImplementedError` - so we should decide what to do, maybe require to override it and explicitly pass or remove the `NotImplementedError`?  We propose to use the solution with pass here (TODO).

In the second case you develop
* `StopOne` - stop a single axis

In the third case you do the same as in the first case, but in the `StopAll` method you have a choice to either send the multi axes command or the single axis command (depending on the number of elements aggregated in the list).

In the fourth case you develop:
* `StopAll` command - send a command to stop all axes.

Similarly to the first case, in this case we see that `StopOne` is not necessary to be developed. Now it raises `NotImplementedError` - so we should decide what to do, maybe require to override it and explicitly pass or remove the `NotImplementedError`? We propose to use the solution with pass here (TODO).

In case of abort applies the same rules as above stated.

**Possible scenarios when stopping/aborting elements in the Pool**

Each of the scenarios is followed by the sequence of calls being executed as it is now - before integrating this PR. Scenarios which require modifications due to the changes proposed in this PR are commented.

1. Stop a single motor in motion e.g. run mv macro and abort it with Ctrl+C in spock
* Motor.Stop Tango command
* `PoolElement.stop`
* `PoolController.stop_one`
* `PoolController.raw_stop_one`
* `StopOne`

This scenario still requires modifications in the code cause it won’t work correctly if the controller is developed with the guidance stated above in case of 1,3 and 4 type of hardware (TODO)

2. Stop a pseudomotor in motion e.g. with the mv macro and abort it with Ctrl+C in spock.
* PseudoMotor.Stop Tango command
* `PoolPseudoMotor.stop`
* `PoolBaseElement.stop` and `PoolBaseGroup.stop`
* For each controller involved in the pseudo call `PoolController.stop_elements` with the involved axes
`PreStopAll` -> `PreStopOne` & `StopOne` -> `StopAll`

3. Stop a motor group in motion e.g. with the move macro of a pair of physical motors with Ctrl+C in spock.
* MotorGroup.Stop Tango command
* `PoolMotorGroup.stop`
* `PoolGroupElement.stop`
* `PoolBaseElement.stop` and `PoolBaseGroup.stop`
* For each controller involved in the pseudo call `PoolController.stop_elements` with the involved axes
`PreStopAll` -> `PreStopOne` & `StopOne` -> `StopAll`

4. Stop a measurement group
* MeasurementGroup.Stop Tango command
* `PoolMeasurementGroup.stop`
* Software Synchronizer stop & `PoolGroupElement.stop`
* `PoolBaseElement.stop` and `PoolBaseGroup.stop`
* For each controller involved in the group call `PoolController.stop_elements` with the involved axes
`PreStopAll` -> `PreStopOne` & `StopOne` -> `StopAll`

5. Stop/Abort a pool
* Pool.Stop Tango command
* `Pool.stop`
* For each controller call `PoolController.stop_all`
* `StopAll`

This scenario still requires modifications in the code cause it won’t work correctly if the controller is developed with the guidance stated above in case of 1,3 and 4 type of hardware. We propose to directly use stop_elements without arguments. The same applies to abort (TODO).

6. Emergency break in the motion action which can be triggered by either errors in the starting sequence or by Fault or Alarm states during the motion.
* `PoolController.emergency_break`
There are two possibilities, if list of elements is passed:
* for each element call `PoolController._raw_stop_one`
* StopOne
* If `PoolController._raw_stop_one` fails, call `raw_abort_one` 
* `AbortOne`
If no elements are passed:
* `PoolController.raw_stop_all`
* `PoolController.raw_stop_all` fails, call `raw_abort_all`
* `AbortAll`

This scenario still requires modifications in the code cause it won’t work correctly if the controller is developed with the guidance stated above in case of 1,3 and 4 type of hardware. We propose to call PoolController.stop_elements with the list of elements to be stopped (TODO).


**Backwards compatibility**

* Before this PR the only available methods when developing controller plugins were: `StopOne` and `StopAll` or `AbortOne` and `AbortAll`.
* In the scenario 2, the actual methods being called were only `StopOne` for each physical axis of the corresponding physical motor controller involved in a pseudo motor. In this case we think that the backwards compatibility won’t be broken - the controllers already implementing the `StopOne` will still work as before.
* We believe that there are no controllers that were overriding the default behavior of `StopAll`.

**Other comment**
* `stop_axes` and `abort_axes` of PoolController could be removed - they are not used (TODO).

**Actions to be taken**
* We should review if the backwards compatibility is really maintained: @teresanunez, @amilan and @daneos could you comment on it from the point of view of the controllers developed at your sites. Other users are also welcome to comment:)
* Implement the pending changes - marked as TODO in the above text. @guifrecuni we can do it together with @mrosanes.
* Move the guidelines to the official documentation.

And finally, sorry about so long commnet ...if someone had reached this line:)

teresanunez:
Hi @reszelaz, I do not see that we could have problems with our controllers after this implementation.
Thanks for the detailed description.

reszelaz:
Continuing this disucssion we think that the `stop_elements` method should handle possible errors in the controller's code. The general strategy should be to try to stop as many elements as possible. We comment it in the following pseudo code of the group stop:

```
/FOR/ Each controller(s) implied in the motion

     - Call PreStopAll()  # if fails we pass to the next controller

    /FOR/ Each motor of the given controller implied in the motion
        - ret = PreStopOne(motor to stop)  # if fails we pass to the next motor
        - /IF/ ret is not true
            /RAISE/ Cannot stop. Motor PreStopOne returns False  # this should not raise but pass to the next motor
        - /END IF/
        - Call StopOne(motor to stop)  # if fails we pass to the next motor
    /END FOR/

    - Call StopAll()  # if fails we pass to the next controller

/END FOR/

# if any exeception was raised during the algorithm execution we raise an exception 
# with the list of controllers/motors that could not be stopped
```
Your feedback is more than welcome.

reszelaz:
> we think that the stop_elements method should handle possible errors in the controller's code.

Yes, I've checked that and in the develop branch it behaves correctly. If, for example, the `StopOne` method fails on one axis, the rest of the axes are tried to be stopped.

What is currently not optimal is that this failures are not notified to the user. So for example, if the user issued `Ctrl+C` in spock and one axis could not be stopped due to an exception in the `StopOne` method the user does not know about it. He just sees:

```
Door_zreszela_1 [9]: mv mot01 1000 mot02 1000
^C
Ctrl-C received: Stopping... Done!
```
but the motor continuous to move. But I think we could treat it in another feature.

reszelaz:
After thinking about it carefully I’m more in favor of leaving the methods removed in 795a325. This is because the `PoolController` API is oriented to the axis numbers and not the elements. Actually only the `stop_elements` and  `abort_elements` use element objects as arguments. So these removed methods should stay but their implementation should change. Since we are in rush preparing the release I think it is better to mark it as a todo. 

In order to close this feature, the only missing thing is to handle the exceptions possibly coming from controllers so that we try to stop as many elements as possible in all scenarios.

reszelaz:
Flake8 error F403 was avoided by marking "import *" in `sardana.pool.test` with `#NOQA` - more details in a1579cb.

reszelaz:
Thanks @guifrecuni and @mrosanes for the implementation. I've just corrected minor things and now the PR is ready so I'm merging it.

Below the summary of changes:
* The new algorithm of stopping/aborting elements was implemented and is now documented in [How to write a motor controller](http://www.sardana-controls.org/en/latest/devel/howto_controllers/howto_motorcontroller.html#multiple-motion-synchronization) (actually will be available when RTD finishes building the new doc).
* Guidelines on how to program the `Stoppable` interface methods are similar to those of the `Startable` interface which are also documented in the same chapter of the official documentation. More specific guidelines are present in https://github.com/sardana-org/sardana/pull/592#issuecomment-362026115 with one detail: it is not explicitly said which method should be used in order to add axis to the list of axis to be stop. This can be either `PreStopOne` or `StopOne` and it is up to the controller developer which one to use.
* Now, in case of a failed operation of Stopping in spock the user is informed that something went wrong.
* The following, unused and internal methods were removed from the `PoolController` calss: `raw_stop_all`, `raw_stop_one`, `_raw_stop_all`, `_raw_stop_one`, `stop_all`, `stop_one`, `raw_abort_all`, `raw_abort_one`, `_raw_abort_all`, `_raw_abort_one`, `abort_all`, `abort_one`

PR #594, opened 2017-09-08
Fix repetitions for software acquisition

reszelaz:
Software synchronized acquisition receives wrong number of repetitions
(the global one, that should be passed only to the hardware synchronized 
acquisition). Fix it by setting 1 as number of repetitions - single software
synchronized acquisitions will be repeated while the software synchronizer
triggers them.


reszelaz:
I'm also merging this one, cause it was a clearly copy&paste bug.

PR #595, opened 2017-09-08
Fix validating motor positions with limits in ascanct & co.

reszelaz:
Final position is wrongly check against its limits. Fix it.
Also fix the motor name extraction in order to correctly format the error message.


teresanunez:
I have checked it and it corrects both bugs. I can merge it if nobody against.

reszelaz:
I think it is safe to merge it cause this was clearly a bug. Thanks for the review.

PR #597, opened 2017-09-11
Start hardware TriggerGates as last ones

reszelaz:
Software synchronizer is started as the last one in the synchronization
action when the hardware TriggerGate elements are already working. This is
not a problem if the hardware TriggetGate elements work in the position
domain but if they work in the time domain they may start triggering too
fast with respect to the motor's move - the motors are started to move
by the GScan when the MeasurementGroup Start has finsihed. Make the hardware
TriggerGate elements start as the last ones in the synchronization action
what aslo means the last ones in the MeasurementGroup Start command. This
change does not affect the software synchronization which works in the
position domain.

### DIFF
--- a/src/sardana/pool/poolsynchronization.py
+++ b/src/sardana/pool/poolsynchronization.py
@@ -138,6 +138,11 @@ class PoolSynchronization(PoolAction):
                                               self._synch_soft)
                 self.add_finish_hook(remove_pos_listener, False)
 
+            # start software synchronizer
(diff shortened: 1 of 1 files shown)


droldan:
I have tested the changes of start procedure and it works fine, we can see the improvement clearly.
Thanks.

reszelaz:
If there are no other comments, I will integrate it within 24h.

PR #601, opened 2017-09-15
Implement timeout/watchdog in continuous scans (issue-136)

### LINKED ISSUE #136: Timeout/Watchdog in Sardana useful for continuous scans (among others) (SF#245)
Sardana could have a timeout/watchdog feature, especially interesting in the continuous scans, when hardware problems (missed triggers) could leave the acquisition waiting infinitely.

Reported by: sagiss ( http://sf.net/u/mrosanes )

Original Ticket: [sardana/tickets/245](https://sourceforge.net/p/sardana/tickets/245)

reszelaz:
Hardware problems (missed triggers or positioning problems) could leave the
acquisition waiting infinitely. Add a timeout/watchdog mechanism that will
stop the measurement group if the estimated acquisition time elapses and
the measurement group does not finish. Add an extra time (15 s) to be
sure that the acquisition will not finish. Fix #136.

The extra time is hardcoded now. But I imagine we will make it configurable in the future when we get experience with this kind of problems.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -2182,7 +2182,7 @@ class CTScan(CScan, CAcquisition):
             finally:
                 # wait extra 15 s to for the acquisition to finish
                 # if it does not finish, abort the measurement group
-                # (this could be due to missed hardware triggers or 
+                # (this could be due to missed hardware triggers or
                 # positioning problems)
                 # TODO: allow parametrizing timeout
                 timeout = 15
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -2180,10 +2180,12 @@ class CTScan(CScan, CAcquisition):
                     "Moving to waypoint position: %s" % repr(final_pos))
                 motion.move(final_pos)
             finally:
-                # for the moment just adds an extra 15 s to wait for the
-                # acquisition to finish TODO: allow parametrizing extra time
-                EXTRA_TIME = 10
-                timeout = delay_time + total_time * repeats + EXTRA_TIME
+                # wait extra 15 s to for the acquisition to finish
+                # if it does not finish, abort the measurement group
+                # (this could be due to missed hardware triggers or 
+                # positioning problems)
+                # TODO: allow parametrizing timeout
+                timeout = 15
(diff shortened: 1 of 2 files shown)


teresanunez:
HI @reszelaz, I could not test it with hardware, but reviewing the code I think it could be merged.
IMO this functionality is needed.

teresanunez:
@reszelaz, perhaps I miss it because I am simulating ... but the 'waitFinish' is done in 'finally:' ...
 so it is not called during the scan. Any issue blocking the scan will go to the 'finally' statement?.
 Trying to simulate it I do not get it,  I am still trying.

reszelaz:
Many thanks @teresanunez for giving it a try. The `waitFinish` must be called in case of a normal finish and in case of an exception. This is how the asynchronous operations (move, acquisition) must be used. That's why it is placed in the `finally`.  When implementing it, I have simulated this behavior with a dummy counter synchronized by hardware, previously changing the code of the controller. After a start of the counter I was always returning Moving state. This test shows how the measurement group is stopped after the timeout.

teresanunez:
Hi @reszelaz, many thanks, I have tested it now as you wrote above and it works, but I have a doubt ...
why do you want to set timeout as:

timeout = delay_time + total_time * repeats + EXTRA_TIME

I think it is not need it to wait:  delay_time + total_time * repeats, with only EXTRA_TIME could be
enough ... If the scan has already finish, like it is in the simulated case, you don't want to wait again
 the whole time of the scan, that could be quite long, and if the scan has a problem and you arrive
 to 'finally' due to a exception, why do you want to wait the time that the scan would take plus
 the EXTRA_TIME. Would not be enough with the EXTRA_TIME?. In my simulated case, for
 example, I have to wait minutes after the motor reaches the end position before the MG is
 stopped and I get the exception in spock, and during that time nobody knows what is going on.

reszelaz:
You are totally right! I forgot that the move is synchronous in the `try..finally` clause, so it will already wait the amount of time corresponding to `delay_time + total_time * repeats + deceleration_time`. The EXTRA_TIME is enough. Thanks a lot! I will push a commit fixing it right now.

teresanunez:
Thanks to you, Zibi.

teresanunez:
Somebody against merging this?, if not I will proceed to merge it.

PR #602, opened 2017-09-19
Fix bug in generate timestamps (ascanct & co.)

reszelaz:
Too many timestamps are generated. The initial delay already indicates
timestamp of the first point so the loop over repeats must iterate
one time less.  Fix it.


teresanunez:
Hi, I have reviewed this, it solves the bug, it is harmless to merge it. Should I make it?

reszelaz:
Yes, please. Thanks for the review!

PR #603, opened 2017-09-19
Fix documentation

cmft:
Fix the reference of tango motor property "Sleep_bef_last_read".
It is bad writen in the sardana documentation.

Replace Sleep_before_last_read by Sleep_bef_last_read


reszelaz:
Thanks for this fix. I will push it as soon as CI gives us green light.

PR #604, opened 2017-09-19
Reduce motion overheads

reszelaz:
Core objects related to moveables are related by the publisher-subscriber pattern. For example:
* dial position is a publisher and motor’s position is a subscriber
* motor’s position is a publisher and motor is a subscriber
* motor’s position is a publisher and pseudo motor’s position is a subscriber
* pseudo motor’s position is a publisher and pseudo motor is a subscriber
* motor is a publisher and pseudo motor is a subscriber

The attached sketch tries to demonstrate it (please forgive me its poor quality...)

motion-overheads

Since a pseudo motor is subscribed to motor’s events with the `on_element_changed` callback it processes all the dial position events propagated via the following chain: `dial position -> position -> motor -> pseudo motor`. After looking at the callback code, I think that this is not necessary, cause the only thing that it does is to recalculate the state and status. State and status information are already well calculated because at the beginning and at the end of the motion these are notified with the corresponding events. What is even worse is that the `on_element_changed` callback is called twice because there is an another event propagation route: `dial position -> motor -> pseudo motor`. I propose to filter out events different than state and status. A similar change is proposed for motor groups.

### DIFF
--- a/src/sardana/pool/poolmotorgroup.py
+++ b/src/sardana/pool/poolmotorgroup.py
@@ -139,12 +139,9 @@ class PoolMotorGroup(PoolGroupElement):
 
     def on_element_changed(self, evt_src, evt_type, evt_value):
         name = evt_type.name.lower()
-        if name in ('state', 'position'):
+        if name == 'state':
             state, status = self._calculate_states()
-            if name == 'state':
-                propagate_state = evt_type.priority
-            else:
-                propagate_state = 0
+            propagate_state = evt_type.priority
             self.set_state(state, propagate=propagate_state)
             self.set_status(status, propagate=propagate_state)
 
--- a/src/sardana/pool/poolpseudomotor.py
+++ b/src/sardana/pool/poolpseudomotor.py
@@ -303,17 +303,17 @@ class PoolPseudoMotor(PoolBaseGroup, PoolElement):
 
     def on_element_changed(self, evt_src, evt_type, evt_value):
         name = evt_type.name.lower()
-        # always calculate state.
-        status_info = self._calculate_states()
-        state, status = self.calculate_state_info(status_info=status_info)
-        state_propagate = 0
-        status_propagate = 0
-        if name == 'state':
-            state_propagate = evt_type.priority
-        elif name == 'status':
-            status_propagate = evt_type.priority
-        self.set_state(state, propagate=state_propagate)
-        self.set_status(status, propagate=status_propagate)
+        if name in ("state", "status"):
+            status_info = self._calculate_states()
+            state, status = self.calculate_state_info(status_info=status_info)
+            state_propagate = 0
+            status_propagate = 0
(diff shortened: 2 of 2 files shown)


teresanunez:
Hi @reszelaz, I have had a look at this ... It is not affecting at the behaviour, but I have observed that with this new implementation, if I move a motor, ex. mot1, and a pseudomotor depending on it, ex. pm1. I have GUIs opened showing the position of mot1 and pm1, and I move mot1 ... the position of pm1 changes accordingly, but it stays always with blue color in background ... until I read the pm1 positon or reopen de GUI, I mean, it is missing the event changing the corresponding property. I think ...

reszelaz:
Thanks for the review. I confirm that. There must be some inconsistency with the attribute quality when emitting the position event. I will investigate it.

teresanunez:
Hi @reszelaz, thanks for the update. The problem I reported above is not there any more. IMO it is fine
now.

reszelaz:
Indeed the problem is the wrong Position quality in the group elements (pseudo motors and motor groups). Since in the Sardana core there is no quality metadata assigned to a value the Tango layer uses the element’s state to determine the quality. As it is now, each position change of the motion group elements recalculates state. This is not needed, and should be done only when the state has changed. So for that need, before updating the position for the last time when the motion finishes, we update the state. This update however can not be pushed with the Tango event, because too early state transition will allow another motion to start before the last position event is sent. Of course the position can be read before emitting the state event, so it will have the correct value, but anyway I think that this is not correct. If you think that it is not critical, just let me know, because then everything becomes much simpler:).

Anyway, I have managed to implement it by adding 2 more levels of propagating events (only internal propagation) and cheating a little bit. At the end of the motion, I switch the state internally for the need of Position quality, and then I put it back to Moving (again internally), before setting the final state internally and externally. This way the Tango layer can filter out events that are not supposed to be propagated to the Tango world in form of events and when the Position quality is determined it is already updated. Well,  the code could be improved, by adding the enum for the propagation levels that are now pure integers...

Now thinking, we could avoid this hack by simply adding quality metadata to Sardana values and always putting the new position with the corresponding quality from within the motion action. I would need to explore how difficult it is to implement it this way. So for the moment, I just encourage you to test it but please don’t integrate it yet.

Just for your knowledge I have profiled this implementation and in a move of two dummy motors (velocity 100, acceleration and deceleration 2) on range of 100 with MotionLoop_StatesPerPosition set to 1 and a gap and offset defined on top we have reduced, for example, a number of calls to PoolBaseGroup._calculate_element_state from ~3500 times to ~50 times (I attach cProfile outputs). This optimization may allow higher frequency of MotionLoop in systems with multiple pseudo motors or motor groups defined on top of moveables being moved so desired in software synchronized continuous scans. In a real system at one beamline we have 26 pseudo motors and approx 10 motor group on top of the moveable being scanned.

Issue #445, opened 2017-04-06
expconf does not allow to change the measurement group’s master timer nor the master monitor


jkotan:
Hi,

It looks like expconf does not allow to change the measurement group’s master timer nor the master monitor. It seems that them master timer is defined by the first experimental channel.

Cheers,
Jan

PR #607, opened 2017-09-22
Respect enabled flag on controllers

reszelaz:
When acquisition starts, the enabled flag is not correctly respected on controllers.
The controllers that are not enabled (none of their channels is enabled) are loaded, started, etc.
This bug was probably introduced while merging SEP6 into develop because this feature
used to worked before. Fix it.

### DIFF
--- a/src/sardana/pool/poolacquisition.py
+++ b/src/sardana/pool/poolacquisition.py
@@ -545,14 +545,6 @@ class PoolAcquisitionBase(PoolAction):
                 if ElementType.CTExpChannel in ctrl.get_ctrl_types():
                     _pool_ctrl_dict_loop[ctrl] = pool_ctrl_data
 
-        pool_ctrls = []
-        self._pool_ctrl_dict_loop = _pool_ctrl_dict_loop = {}
-        for ctrl, v in pool_ctrls_dict.items():
-            if ctrl.is_timerable():
-                pool_ctrls.append(ctrl)
-            if ElementType.CTExpChannel in ctrl.get_ctrl_types():
-                _pool_ctrl_dict_loop[ctrl] = v
-
         # make sure the controller which has the master channel is the last to
         # be called
         pool_ctrls.remove(master_ctrl)


teresanunez:
Hi @reszelaz, I have probably not understood this correctly, but I do not see what you describe.
If I have for ex. ein dummyCounterTimer controller and none of its channels are in the MG the load/start
functions of the controller are not called ... In any case I think the lines you remove should be removed.

reszelaz:
Thanks @teresanunez for the review! This bug appears if you use the enable/disable channel feature of the measurement group. You can try it with the `expconf` widget. To trigger it you add one controller and some of its channels to the measurement group and then disable all of these channels (the second column called "enabled" in the expconf). In this case the said controller will still be loaded... 

This was not the case before merging SEP6 - see last commit before the merge 919789b.

teresanunez:
Hi @reszelaz, thanks, now I go it. I have tested it now  and the bug you mention above is solve. What I have observed is that if you have a controller with channels in the MG but all these channels are disabled, the Pool through following error at each scan point:

SardanaTP.W007 ERROR    2017-09-25 11:32:57,240 SardanaTP.SardanaTP.W007: Uncaught exception running job 'run' called from thread SardanaTP.W008:
  File "/usr/lib/python2.7/threading.py", line 783, in __bootstrap
    self.__bootstrap_inner()
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/dist-packages/taurus/core/util/threadpool.py", line 134, in run
    cmd(*args, **kw)
  File "/home/tnunez/sardana_patch_4/src/sardana/util/funcgenerator.py", line 167, in run
    self.fire_active()
  File "/home/tnunez/sardana_patch_4/src/sardana/util/funcgenerator.py", line 223, in fire_active
    self.fire_event(EventType("active"), self._id)
  File "/home/tnunez/sardana_patch_4/src/sardana/sardanaevent.py", line 110, in fire_event
  File "/home/tnunez/sardana_patch_4/src/sardana/sardanaevent.py", line 130, in _fire_event
    real_listener.event_received(self, event_type, event_value)
  File "/home/tnunez/sardana_patch_4/src/sardana/pool/poolacquisition.py", line 253, in event_received
    get_thread_pool().add(self._sw_acq.run, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/taurus/core/util/threadpool.py", line 134, in run
    cmd(*args, **kw)
  File "/home/tnunez/sardana_patch_4/src/sardana/pool/poolaction.py", line 344, in run
    self.start_action(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/taurus/core/util/log.py", line 158, in wrapper
    ret = f(*args, **kwargs)
  File "/home/tnunez/sardana_patch_4/src/sardana/pool/poolacquisition.py", line 751, in start_action
    PoolAcquisitionBase.start_action(self, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/taurus/core/util/log.py", line 158, in wrapper
    ret = f(*args, **kwargs)
  File "/home/tnunez/sardana_patch_4/src/sardana/pool/poolacquisition.py", line 550, in start_action
    pool_ctrls.remove(master_ctrl)
ValueError: list.remove(x): x not in list


The scans ends without problems, in spock one sees no errors. I have taurus brach support3.x.

                        Regards,
                                       Teresa

teresanunez:
Sorry, I know now where is the error. The Synchronization of the two controllers was set different.
 One was set to hardware and the other to software. If both are set to the same value the error
 from the Pool does not appear. Perhaps this could be a real scenario ... but in any case the printout
 of the error does not affect the scan.

reszelaz:
I think that the problem is related to the master timer/monitor. If you disable all channels from the controller that one of its channel is the master timer or monitor you will get the error that you describe. But if you do that for any other controller you don't get this error.

Currently it is not possible to change the master timer/monitor of the measurement group due to #445.
I think that at some point we should implement a protection, or at least a warning to the user, in case he tries to disable the master timer/monitor.

teresanunez:
Could it be that in my case there were two master timers?. One controller was synchronized with
 dummy hardware trigger and the other with software one. The master timer was actually enabled ...
 but perhaps the other controller, due to the different synchronization, had a different one. I don't know
 if a case like this is foreseen ... but expconf allows it. In any case it would be a good idea to put some
warnings.

reszelaz:
Currently, in the measurement group there is only one master timer and one master monitor. If you look into the measurement group's configuration attribute, these are the ones set at the top most level of the dictionary, next to the label and description. Then, each of the controllers has its own timer and its own monitor. These are located deeper in the configuration dictionary, on the controller level, next to the channels, synchronization and synchronizer keys.

After the changes introduced with SEP6, now we have two acquisition actions: hardware and software. In case of the hardware synchronization, the master timer/monitor does not have sese, cause anyway these channels will be synchronized by trigger/gates. In case of software acquisition, it still has sense to have the master timer and monitor - the one that will be started as the last one.

From your traces, it looks like the software acquisition had the problem. But it could be that the master timer of the measurement group was actually a channel synchronized by the hardware so it was assigned to the hardware acquisition action, hence the problem you encountered.

Anyway, we should review the possible scenarios and document them and finally implement the proper protections/warnings to the user. I will create an issue for that.

teresanunez:
Hi @reszelaz, many thanks for the explanation.

droldan:
Hi, 
I have sent Pull Request [#615](https://github.com/sardana-org/sardana/pull/615) related with this problem, also contains the changes suggested in this pull request.

reszelaz:
PR #615 was integrated in develop and it already deals with this issue. Closing this PR.

Issue #568, opened 2017-08-28
defelem doesn't automatically increment axis number


daneos:
When you try to add an element to a controller that already has elements and you don't specify the axis number, you might get an error. It looks like the automatic discovery of the lowest non-ocuppied axis number doesn't work (it always returns 2).

PR #609, opened 2017-09-25
Fix defelem macro when using default axis number

reszelaz:
This PR provides a fix for #568


daneos:
I've tested this PR, but it seems to always return `1` instead of `2`.
```
Door_test-issue-568_1 [1]: defctrl DummyMotorController issue568_motctrl01
Created issue568_motctrl01

Door_test-issue-568_1 [2]: defelem issue568_mot01 issue568_motctrl01
Created issue568_mot01

Door_test-issue-568_1 [3]: defelem issue568_mot02 issue568_motctrl01
An error occurred while running Macro 'defelem(issue568_mot02, issue568_motctrl01, -1) -> 62857dec-a466-11e7-b84c-080027b23b22':
Exception: Controller already contains axis 1 (issue568_mot01)
```

daneos:
I did another test, added second motor manually, and then tried to add another one using axis auto discovery, and it still tried to use axis `1`.
```
Door_test-issue-568_1 [11]: defelem issue568_mot02 issue568_motctrl01 2
Created issue568_mot02

Door_test-issue-568_1 [12]: defelem issue568_mot03 issue568_motctrl01
An error occurred while running Macro 'defelem(issue568_mot03, issue568_motctrl01, -1) -> 38416a80-a46d-11e7-b84c-080027b23b22':
Exception: Controller already contains axis 1 (issue568_mot01)
```

reszelaz:
Thanks for the review @daneos !
I can not reproduce what you describe in your first comment. This is how it looks in my case:
```
Door_zreszela_1 [14]: defctrl DummyMotorController motctrl05
Created motctrl05

Door_zreszela_1 [15]: defelem mot501 motctrl05
Created mot501

Door_zreszela_1 [16]: mot501.get_property("axis")
        Result [16]: {'axis': ['1']}

Door_zreszela_1 [17]: defelem mot502 motctrl05
Created mot502

Door_zreszela_1 [18]: mot502.get_property("axis")
        Result [18]: {'axis': ['2']}

Door_zreszela_1 [19]: defelem mot503 motctrl05
Created mot503

Door_zreszela_1 [20]: mot503.get_property("axis")
        Result [20]: {'axis': ['3']}
```

daneos:
My previous tests were done on CentOS 7, so I tried this on Debian to see if it will make any difference.\
Unfortunately, running any macro results only in this error:
```
DevFailed: DevFailed[
DevError[
    desc = Received malformed data for event tango://10.81.0.101:10000/door/test-issue-568/1/state.change
  origin = ZmqEventConsumer::push_zmq_event()
  reason = API_WrongEventData
severity = ERR]
]
```
Full traceback in the attachment.
[issue-568-traceback.txt](https://github.com/sardana-org/sardana/files/1348830/issue-568-traceback.txt)

reszelaz:
That's strange. Since you report that it happens on any macro execution, it makes me think that it could be a more general issue, maybe Tango?. Could you do a simple test and try to subscribe and receive state events coming from other devices e.g. a motor in the Pool.
In my case I also use Debian 9 with the following packages:
* python-tango 9.2.0-2
* libtango9 9.2.5a+dfsg1-2 with the attached [patch](https://github.com/sardana-org/sardana/files/1348996/0004-Fix-tango-controls-cppTango-354.patch.txt) - this patch should make a difference here

daneos:
I'm running Debian 9 with Tango 9.2.5a installed from sources (without a patch) and PyTango 9.2.2.
I'll apply this patch and see what happens.

reszelaz:
Sorry, I wanted to say: should NOT make a difference - a typo from my side. I think it is more interesting to perform the test of the state events.

daneos:
Before you wrote about testing subscriptions I decided to reinstall my Debian VM to have a clean environment and surprisingly this was enough to get Sardana working.
I still, however, experience the problem with `1` always being used as axis number.
It also doesn't matter if I run single Sardana instance or separate Pool and MacroServer.

I tested this on the following configurations with the same result:
- CentOS 7, Tango 9.2.5a, PyTango 9.2.2
- Debian 9, Tango 9.2.5a (without patch), PyTango 9.2.2
- Debian 9, Tango 9.2.5a (with patch), PyTango 9.2.2
- Debian 9, Tango 9.2.5a (without patch), PyTango 9.2.0
- Debian 9, Tango 9.2.5a (with patch), PyTango 9.2.0

For all these configurations I used zeromq 4.2.2 and omniORB 4.2.2, both compiled from sources.

daneos:
I think I have found the problem. In the `Controller.getUsedAxes` method, in the `for` loop `elem.controller` contains the controller name without the `tango://` scheme, while `self.getFullName()` returns the name with the scheme. The comparison is always false, so last used axis is always `None`.

reszelaz:
Great! It sounds like a typical Taurus3/Taurus4 comparison problem. It may perfectly be that I've tested it on Taurus 3 only, sorry!

Thinking about a solution for that... Since the Pool does not use the scheme in the internal registers yet, we would need to strip the scheme before the comparison and mark it as a TODO related to the Taurus3/Taurus4 compatibility. I suspect that there may be still a problem with the FQDN in the authority name - similar to what was found in #623 and #627. Do you think that this could be a problem here as well?

If you already found a solution to that, don't hesitate to push directly to my branch. Thanks a lot!

daneos:
OK, it looks like fixed for Taurus4. You can try to run this also with Taurus3 to see if I didn't break anything.
I've tested this for FQDN compability, and the FQDN is not being used neither in loop `elem` nor `getFullName`, but I'm not sure if it should.

reszelaz:
Thanks! I have tried it on Taurus3 and there are no problems. But I still have problems on Taurus4. I use the latest taurus develop branch, which version do you use? The FQDN in taurus URIs were introduced after the Jul17 release.

daneos:
I just realised that I changed the hostname of my VM to include domain, but forgot to change TANGO_HOST...
But with TANGO_HOST changed it also works, FQDN is used everywhere and comparison passes.
I use Taurus 4.1.1, I will try with the `develop` branch.

daneos:
Just tested with latest Taurus `develop` and I didn't find any problems, with or without using FQDN.

reszelaz:
I have also checked it when using FQDN in the TANGO_HOST and there are no problems. The problem appear if I change the TANGO_HOST to PQDN, as I used to work before.

daneos:
Could you describe the problem you're experiencing? I'm running this with PQDN right now and everything seems to be OK.

reszelaz:
```
Door_zreszela_1 [1]: defctrl DummyMotorController motctrl10
Created motctrl10

Door_zreszela_1 [2]: defelem mot1001 motctrl10
Created mot1001

Door_zreszela_1 [3]: mot1001.get_property("axis")
        Result [3]: {'axis': ['1']}

Door_zreszela_1 [4]: defelem mot1002 motctrl10
An error occurred while running Macro 'defelem(mot1002, motctrl10, -1) -> 1cc91530-a857-11e7-b8f5-f8b156a48f79':
Exception: Controller already contains axis 1 (mot1001)
```
The problem is in the comparison that you describe in https://github.com/sardana-org/sardana/pull/609#issuecomment-333845721. In my case I have:
```
elem.controller = pc255:10000/controller/dummymotorcontroller/motctrl11
name = pc255.cells.es:10000/controller/dummymotorcontroller/motctrl11
```
What does not match.

daneos:
That's weird, I have both = `v-deb.new.solaris.x.daneos.com:10000/controller/dummymotorcontroller/issue568_motctrl01` when using FQDN and both = `v-deb:10000/controller/dummymotorcontroller/issue568_motctrl01` when using PQDN.

I'm testing on Debian 9 with Taurus `develop`, so it shouldn't differ between my and your setup.
On the other hand, `hostname` on my system (even without `-f` flag) gives the FQDN, while `domainname` gives just `(none)`, so maybe my config is broken.

reszelaz:
@daneos sorry for not replying so long, the ICALEPCS conference in the last days consumed all my time..
I will make a study on how to migrate Sardana to use FQDN, so we could avoid this kind of problems. I will come back to you soon.

daneos:
@reszelaz my configuration you asked for during follow-up meeting:
Debian 9, Tango 9.2.5a (with omniORB 4.2.2 and zeromq 4.2.2), PyTango 9.2.2, Taurus 4.1.1.
I have also tested this with Taurus `develop`.
I also managed to fix the problem with domain names, but still, the solution works for me.

By the way, I have found another problem with FQDN/PQDN compability, described here #667 .

reszelaz:
Since Jan18 release we started using URI as Sardana element references.
I have removed the taurus4 compatibility layer from this PR, which BTW was not working 100 % correctly, which is no more necessary.

I have quickly tested this PR with taurus4 and taurus3. The #568 seems to be solved now.
Can any of @sardana-org/integrators review it?

teresanunez:
Hi @reszelaz, I have tested this with taurus3 and taurus4 and it works properly, it solves also the issue #568. I can proceed to merge it if you want.

reszelaz:
Great! Go ahead then. Many thanks!

Issue #610, opened 2017-09-25
Measurement group master timer monitor role after SEP6


reszelaz:
Master timer and master monitor are the last channels to be started by the acquisition action. These are the first timer and monitor channels in the measurement group and should be configurable by changing the order of channels in the expconf - currntly buggy #445.

SEP6 introduces new concepts to the measurement process e.g. synchronizers, software and hardware acquisition, etc. With this changes, the master timer and monitor roles gets more complicated. We should document all the possible scenarios of configurations. Furthermore, the current implementation seems to be not well protected agains some more advanced configurations - see #607.

PR #614, opened 2017-09-27
Find Timer on MeasurementGroup set_configuration method

droldan:
It is related with [#445](https://github.com/sardana-org/sardana/issues/445) and [#610](https://github.com/sardana-org/sardana/issues/610)
On set Measurement Configuration it will detect the timer/monitor.
It will search the first channel enabled, the timer and monitor will be set with the channel found.
Its allows to configure the measurement group timer/monitor by changing the order of channels in the expconf.
Also, works if the configuration is passed programmatically.


teresanunez:
I have tested this with expconf and it solves the issue #445. It also unifies the master timer with the timer of the controller the master timer belongs to, what I think it is important. IMO it could be merged.

droldan:
Thanks @teresanunez !

droldan:
Yesterday we spoke with Zibi about this pull request, and we found an issue, on try to find the timer/monitor is necessary to check only if the channel has a Software Synchronizer selected, not Hardware Synchronizer.
It is pending to fix it.

reszelaz:
Thanks @droldan and @teresanunez for investigating and reviewing this PR.
As described in #610, after the integration of the SEP6, the measurement group configuration got really complicated. As it was before, when only software synchronization was in used, the master timer/monitor roles in the measurement group configuration were enough. The order of channels (more precisely channels configured as timer/monitor for a given controller) in the measurement group was used to determine the master timer/monitor. Well at least this was the theory, because the expconf widget is actually buggy and it never updates the master timer/monitor - #445.
At the beginning of the SEP6 development, we took a technical debt and did not pay too much attention to the master timer/monitors for each of the acquisition actions (software and hardware). This was actually documented as a TODO in the code but it was unnoticed when we were finishing the SEP6 developments.
In the acquisition preparation moment, the complete measurement group configuration gets splitted into sub-configurations for each of the actions. And the master timer/monitor is assigned randomly. We have just discovered this and fixed it by assigning a master timer/monitor for each of the actions according to the rule used up to now, so basically respecting the order of channels in the measurement group. I have just pushed new commits with the necessary implementation and I have made some testes. But it would be nice if you could test it as well before the final integration. Just keep in mind that I have overridden the history of the branch, eliminating the commits that @droldan proposed before.

So in summary:
* Now the master timer/monitor is selected on the server side
* The information about the master timer/monitor passed via the configuration attribute of the measurement group is completely ignored.

We should come back to this issue at some point and I propose to do it when we will refurbish the expconf widget. At least at ALBA we have plenty ideas on how to make it more user friendly. IMHO I would make the assignment of the master/timer monitor more explicit e.g. with a different background color of the channel-timer cell and use the information coming from the client via the configuration attribute.

droldan:
I have tested with expconf and it is working well.
Note: Is selected the first channel ignoring if it is enabled or disabled (expected behavior).
IMO it needs to be integrated, without this propose we can found some problems with randomly timer/monitor.

reszelaz:
Thanks @droldan for the tests. I completelly agree with the urgency of this issue. I proceed with the integration.

Issue #587, opened 2017-09-01
Exceptions from controllers are not propagated to the user


daneos:
Exceptions raised and not caught inside controllers should be passed to the user, as they may contain useful information.

When a controller raises a standard Python `Exception`, the only thing that's displayed in Spock is
```
door/bl04bm/devel [39]: ct 0.0001
Fri Sep 01 15:15:53 2017

An error occurred while running Macro 'ct(0.0001) -> ba578b82-8f17-11e7-9dcc-b8ca3aadf63e':
Measurement group ended acquisition with Fault state
```
while the exception itself contains a description of what happened (`raise Exception("Keithley6482CTController: Acquisition time should be between 0.01 and 10 PLC")`).

If the controller raises `PoolException` it's even worse, because the user doesn't get any error message; with my example:
```
door/bl04bm/devel [40]: ct 0.0001
Fri Sep 01 15:31:48 2017

  kc1  =    <nodata>
  kc2  =    <nodata>
```

I have tested this only for counter/timer controllers.

reszelaz:
Due to the changes in introduced in the SEP6, the software synchronized acquision is no more started in the measurement group's Start command but in the worker threads. Then it is not so easy to pass this information. As a workaround we interpret that a problem happened based on the state - Fault.

I fully agree that it is not acceptable in the long term, but to correctly implement it it will require bigger changes and I won't have time to do it in the short term.

PR #615, opened 2017-09-27
Fix use of mutable channels object

droldan:
Improve the start_action method, solve problems on manage mutable
objects enabling/disabling measurement group channels.

This patch solves conflicts on save the measurement Group configurations when it has disabled channels.
This Pull Request is related with [PR-607](https://github.com/sardana-org/sardana/pull/607)

### DIFF
--- a/src/sardana/pool/poolacquisition.py
+++ b/src/sardana/pool/poolacquisition.py
@@ -545,6 +545,23 @@ class PoolAcquisitionBase(PoolAction):
                 # only CT will be read in the loop, 1D and 2D not
                 if ElementType.CTExpChannel in ctrl.get_ctrl_types():
                     _pool_ctrl_dict_loop[ctrl] = pool_ctrl_data
+            # ctrl that contains the master timer/monitor can not be disabled
+            elif ctrl is master_ctrl:
+                self.main_element.set_state(State.Fault, propagate=2)
+                msg = "master timer/monitor ({0}) is disabled".format(
+                    master.name)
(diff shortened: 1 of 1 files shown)


reszelaz:
I have checked your proposal and it works well. Apart of that the bug fix it also simplifies the start algorithm. I'm mergin it. Thanks!

reszelaz:
Actually, I took a profit of this PR and implemented what was said in https://github.com/sardana-org/sardana/pull/607#issuecomment-331850643. Now, in case a timer/monitor is disabled the acquisition will not start. Due to the same constraint that is described in https://github.com/sardana-org/sardana/issues/587#issuecomment-327902646 we can not show to the end user the ideal error message.

If the timer/monitor channel belongs is synchronized by hardware the user will see the following error message:

```
Door_zreszela_1 [14]: ct 0.1
Fri Sep 29 15:21:42 2017

An error occurred while running Macro 'ct(0.1) -> 22783268-a519-11e7-91f1-f8b156a48f79':
ValueError: timer/monitor (ct11) of ctctrl04 controller is disabled)
```

If the timer/monitor channel belongs is synchronized by software the user will see the following error message:

```
Door_zreszela_1 [12]: ct 0.1
Fri Sep 29 15:21:13 2017

An error occurred while running Macro 'ct(0.1) -> 10e4615c-a519-11e7-91f1-f8b156a48f79':
Measurement group ended acquisition with Fault state
```

The correct messages are always logged by the server.

PR #616, opened 2017-09-28
Use QIcon.fromTheme to get the Icons

mrosanes:
Some icons are not visible in macroexecutor.

Use QIcon.fromTheme in order to load the icons correctly.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/favouriteseditor/favouriteseditor.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/favouriteseditor/favouriteseditor.py
@@ -29,6 +29,7 @@ favouriteseditor.py:
 import copy
 
 from taurus.external.qt import Qt
+from taurus.qt.qtgui.resource import getIcon
 from taurus.qt.qtgui.container import TaurusWidget
 from taurus.qt.qtcore.configuration import BaseConfigurableClass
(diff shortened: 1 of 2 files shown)


cpascual:
Your implementation has 2 problems:

- It wont't work with the oldest PyQts supported by taurus-3.x : It uses Qt.QIcon.fromTheme, which is not available in early PyQt versions still supported by taurus 3.x. For Taurus 4 it is ok since we only support theme-aware PyQt versions
- You are changing a resource icon by a theme icon. The original one would not change if the theme changes, while in your fix, the icon may look different on different systems. Although **in the case of the icons changed by this PR**, I do not see it as a big problem (it could even be considered better) because the actions seem to match well with the theme description. 

The most conservative approach would be to use the taurus3.x API (which is supported by taurus4 with deprecations) until sardana is ready to drop support for Taurus 3. For example, use:

```python
from taurus.qt.qtgui.resource import getIcon
self.removeAction = Qt.QAction(getIcon(":/actions/list-remove.svg"), ...)
```

mrosanes:
Applied modifications to allow Taurus3 compatibility while maintaining the compatibility with Taurus4, following the snippet provided by @cpascual.

reszelaz:
I've checked it on taurus3 and taurus4 and now in both cases. Merging it...
Thanks to both!

PR #618, opened 2017-09-28
PoolChannel shows "dev_alias" as label

### LINKED ISSUE #617: PoolChannel shows "dev_alias" as label
PoolChannel shows "dev_alias" as label instead of the device alias with Taurus4

cmft:
PoolChannel shows "dev_alias" as label instead of the device alias
with Taurus4

Fix #617


reszelaz:
I just revert the change for PoolChannel class. The bug described in #617 is related to PoolChannelTV class and not PoolChannel. Apart of that it works nicelly in taurus 3 and tauru4.

PR #612, opened 2017-09-26
Do not serialize instrument information

cmft:
The PoolElement serialize method include instrument information,
the instrument are dynamically created and the serialization
in the PoolElement causes troubles.

Avoid to pass it.

PR #620, opened 2017-09-29
Skip MeasTest.test_meas until #57 is fixed

reszelaz:
MeasTest.test_meas fails on travis due to issue #57. Execution of
the testsuite in the same docker container that uses travis locally
does not fail. Skip it until we fix this issue.


reszelaz:
What was agreed in https://github.com/sardana-org/sardana/pull/612#issuecomment-332853206 also requires skipping the test, otherwise the builds on develop branch will fail in travis. I merge it directly.

PR #623, opened 2017-10-01
Adapt continuous scans to FQDN in Taurus 4 model names

reszelaz:
taurus-org/taurus#488 introduced FQDN in model names. The channel's
column description is composed from the full_name that is composed in
the Pool and transferred with the measurement group configuration and it
does not contain the FQDN. Adapt CAcquisition.buffer_changed callback
to be compatible both with Taurus3 and Taurus4.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1836,13 +1836,33 @@ class CAcquisition(object):
             full_name = element.getName()
         else:
             full_name = element.getFullName()
-        # TODO: for Taurus4 compatibility
+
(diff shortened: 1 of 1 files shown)


teresanunez:
I have reviewed this and IMO it is fine. But I could only check that I do not get problems with
 taurus3 (I can not use taurus4 i my computer).

cpascual:
I reviewed the code (but did not test). All seems ok for me.

reszelaz:
Many thanks to both for the review. I've developed it on Taurus4 so it works well. I'm integrating...

PR #624, opened 2017-10-01
Fix continuous scans when used with Taurus4

reszelaz:
Moveables can not contain scheme in their names cause the internal Pool
registers do not use schemes. Strip scheme from the moveables full names
so they are understandable by the Pool.


teresanunez:
I have reviewed this and IMO it is fine. But I could only check that I do not get problems with
 taurus3 (I can not use taurus4 i my computer).

reszelaz:
I confirm that it does not cause problems on Taurus 4. I'm merging it. Thanks!

PR #625, opened 2017-10-01
Re-introduce intermediate event of counters

reszelaz:
Intermediate events of counters were temporarily removed with the acceptance
of SEP6. Introduce them again for software synchronized acquisition.


teresanunez:
I have reviewed and it works as expected. I merge it.

PR #626, opened 2017-10-01
Fix dummy counter value after abort

reszelaz:
Dummy counters involved in an acquisition that was aborted always
returns the value corresponing to the integration time even
if the abort happend earlier. Fix it and always return the real value
of the counter. This has a side effect on the noramlly finished
acquisition which from now on does not show the round number.

If you don't like the side effect we should work on how to return the round
number in case of normal end of the acquisition.

### DIFF
--- a/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py
+++ b/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py
@@ -132,8 +132,6 @@ class DummyCounterTimerController(CounterTimerController):
         if channel.mode == AcqSynch.SoftwareTrigger:
             if self.integ_time is not None:
                 t = elapsed_time
-                if not channel.is_counting:
-                    t = self.integ_time
                 if ind == self._timer:
                     channel.value = t
                 else:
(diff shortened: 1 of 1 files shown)


teresanunez:
I have reviewed this. IMO it is not a problem to see float instead of integer numbers as
 values of the dummy counters. In case of Abort it behaves as described by Zibi, but only
 for one of the counters ... the other ones shown zero, since they are not read after the
 abort is produced, at least in my case ... I make ctrl-C to interrupt the scan and only the first
 counter show a value different from zero in the last point. @reszelaz, is this also the behaviour
you have?.

reszelaz:
Thanks for the review. I didn't realize about it but you are right. I have just added one more commit to update all the channels for the last time when the AbortOne is called.

teresanunez:
Hi @reszelaz, thanks for the update. I have reviewed it now and IMO it can be merged. If nothing against I can make it.

reszelaz:
I found an easy way to improve this and return a round number when the acquisition ends normally. So now, in case of the aborted acquisition we have the partial result, and in case of normal end we have the integration time. I'll merge it then.

PR #627, opened 2017-10-03
Fix uct macro

reszelaz:
Fix bug #319


teresanunez:
I have reviewed and the macro uct works now. I merge it.

PR #628, opened 2017-10-05
Implement Taurus Counter Timer Controller

rhomspuron:


reszelaz:
Thanks for this contribution! This controller goes in pair with those requested in #181. The difference between these two is that the Tango one will be able to make optimizations that are possible only when using the pure PyTango API. This one opens a possibility to access different data sources thanks to the taurus schemes, for example to access the EPICS control system. I'm merging it...

PR #630, opened 2017-10-16
Issue 471: Set the macro data in a ct

mrosanes:
When executing a 'ct' macro, allow to set the macro
data with the channels data, in order to be accessed
from other macrow which executes a 'ct' macro from within
them.

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -658,19 +658,22 @@ class ct(Macro):
         self.outputDate()
         self.output('')
         self.flushOutput()
-
         state, data = self.mnt_grp.count(integ_time)
-
+        macro_data = {}
         names, counts = [], []
         for ch_info in self.mnt_grp.getChannelsEnabledInfo():
             names.append('  %s' % ch_info.label)
-            ch_data = data.get(ch_info.full_name)
+            ch_name = ch_info.full_name
+            ch_data = data.get(ch_name)
             if ch_data is None:
                 counts.append("<nodata>")
             elif ch_info.shape > [1]:
                 counts.append(list(ch_data.shape))
             else:
                 counts.append(ch_data)
+
+            macro_data[ch_name] = ch_data
+        self.setData(macro_data)
 
         table = Table([counts], row_head_str=names, row_head_fmt='%*s',
                       col_sep='  =  ')


mrosanes:
This pull request is dealing with the issue #471, and is a shorter alternative to the pull request #472. It only allows to set the data of the macro 'ct' with the values of the channels data. For the references to the channels the full names are used.

droldan:
Nice @mrosanes, will be useful!

teresanunez:
I have reviewed and it works as described. Solves issue #471.

reszelaz:
Since we are implementing a new feature I would already start using the proper URIs, that means including the tango scheme part.
Now I get this:
```
Door_zreszela_1 [4]: ct 0.001
Thu Oct 19 16:49:25 2017

  ct01  =       0.001
  ct21  =       0.001
  ct11  =       0.001

Door_zreszela_1 [5]: macrodata
        Result [5]: 
{'pc255:10000/expchan/ctctrl01/1': 0.001,
 'pc255:10000/expchan/ctctrl04/1': 0.001,
 'pc255:10000/expchan/ctctrl06/1': 0.001}
```
And what we should get is:
```
{'tango://pc255:10000/expchan/ctctrl01/1': 0.001,
 'tango://pc255:10000/expchan/ctctrl04/1': 0.001,
 'tango://pc255:10000/expchan/ctctrl06/1': 0.001}
```
or even:
```
{'tango://pc255.cells.es:10000/expchan/ctctrl01/1': 0.001,
 'tango://pc255.cells.es:10000/expchan/ctctrl04/1': 0.001,
 'tango://pc255.cells.es:10000/expchan/ctctrl06/1': 0.001}
```
Then we should also prepare recepies on how to access to this data from the wrapper macro assuming that the user knows the short names e.g. ct01 or ct02. This should be doable in taurus3 and taurus4. @mrosanes would you like to continue with this?

Then, what do you think to add some information to the macro description (docstring) about the data format, something similar to what we have for parameters and results. In this case not automatic.

mrosanes:
Yes, why not. I'll try to follow with this development next week, at least to complete the fullname.

reszelaz:
@mrosanes, @droldan I wonder if it wouldn't be easier now to implement this feature with the the
https://github.com/sardana-org/sardana/blob/d8a7f7e95dbda543802d1cc551a401c70c33ea1f/src/sardana/macroserver/scan/scandata.py#L168 class recently enhanced in the #644.

reszelaz:
PR #682 deals with the same issue.

PR #631, opened 2017-10-17
Only allow to count the deprecation warnings if using Taurus>3

mrosanes:
The deprecation warnings have to be counted only in the case of
using Taurus4. Do not count the deprecation warnings in presence
of Taurus3.


cpascual:
The changes seem ok to me,  but the testsuite is failing (not sure if it is related). Maybe taurus4 features are used somewhere else?
 
Note: I fixed a minor PEP8 issue

Issue #500, opened 2017-07-14
Unify data references in scan macros


reszelaz:
In the case of the scans macro data the data references are inconsistent. The moveables use names e.g. “mot01”, the internal channels use the full names without the scheme part e.g. “pc255:10000/expchan/ctctrl01/1” and the external channels use the proper URIs e.g. “tango://pc255:10000/sys/tg_test/1/short_scalar”. I think that this is a bug and we should agree and unify the data references in the scan data. Following this decision, I would use the same for #471 and #472.

Knowing that a scan may involve motors coming from different pools, defined in different databases, I think that we need proper URIs as references. If we agree on this change we should think if the backwards compatibility layer is necessary.

LJBD:
👍 for using full names for every element!

cpascual:
IMHO, as a general rule, all **internal** references to data sources should be *fullname* URIs. 
*normalname*s are acceptable for user I/O and *shortnames* only for display and only in very specific contexts where ambiguities are not an issue.

teresanunez:
I think this is solved. Somebody knows?

reszelaz:
THis issue is still there:

```
Door/zreszela/1 [24]: ascan mot01 0 1 1 0.1
Operation will be saved in /tmp/foo.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #6114 started at Sun Jun 27 09:42:54 2021. It will take at least 0:00:02.202000
 #Pt No    mot01      ct01      ct02      ct03      ct04       dt   
   0         0        0.1       0.2       0.3       0.4     3.03211 
   1         1        0.1       0.2       0.3       0.4     4.15906 
Operation saved in /tmp/foo.h5 (HDF5::NXscan)
Scan #6114 ended at Sun Jun 27 09:42:59 2021, taking 0:00:04.274254. Dead time 95.3% (motion dead time 94.4%)

Door/zreszela/1 [25]: d = %macrodata

Door/zreszela/1 [26]: d[0].data
        Result [26]: 
{'tango://pt222.cells.es:10000/expchan/ctctrl01/1': 0.1,
 'tango://pt222.cells.es:10000/expchan/ctctrl01/2': 0.2,
 'tango://pt222.cells.es:10000/expchan/ctctrl01/3': 0.30000000000000004,
 'tango://pt222.cells.es:10000/expchan/ctctrl01/4': 0.4,
 'point_nb': 0,
 'timestamp': 3.032108783721924,
[...]

PR #633, opened 2017-10-20
Document OutputCols and OutputBlock environement variables

reszelaz:
Add missing documentation of OUtputCols and OutputBlock environment variables.

### DIFF
--- a/doc/source/users/scan.rst
+++ b/doc/source/users/scan.rst
@@ -220,6 +220,15 @@ Following variables are supported:
     Its value is a dictionary with keys pointing to the server side directory
     and values to the client side directory/ies (string or list of strings).
 
+**OutputCols**
+    Select a subset of the scan output columns to be printed. Two different
+    ways of specifying the selected colums exists. First, by specifying an
+    integer number e.g. 2 means print just the very first two colums.
+    Second, by specifying a list of string identifiers e.g. scan elements URIs
+    or the scan record metadata identifiers: "point_nb", "timestamp".
+    The original order of colums is not changed. This feature is similar to
+    the *expconf* enable/disable output.
+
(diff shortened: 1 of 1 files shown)


doc/source/users/scan.rst
```diff
@@ -220,6 +220,22 @@ Following variables are supported:
...
+    of the console output and facilitate an easy visual access to the scan
+    header printed at the beginning of the scan - especially usefull for very
+    long scans.
+
+**OutputCols**
+    Select a subset of the scan output columns to be printed. Two different
+    ways of specifying the selected colums exists. First, by specifying an
+    integer number e.g. 2 means print just the very first two colums.
+    Second, by specifying a list of string identifiers e.g. scan elements URIs
+    or the scan record metadata identifiers: "point_nb", "timestamp".
```
izadworny: It could be usefull to place example of setting OutputCols e.g. 
`senv OutputCols "['tango_host:port/expchan/zerodctrl01/1','motor1','timestamp']"`
It took me quite long time to discover proper syntax.

reszelaz: Very good point! I was also thinking about it. Actually what I wrote in the documentation, about the scan elements URIs, is not 100% right at the moment but what we aim to. This is exactly the same as explained in: #500. I hope that we will get to a unified references ASAP and what we have now is just a buggy behavior.

Then, the `senv` macro syntax e.g. how to pass a list of strings in spock, could be explained in the senv macro documentation.

reszelaz:
The OutputBlock is not an environment variable but a ViewOption. This PR should be corrected.

amilan:
Hi, 

I have added this two variables to the environment variables catalog. 
It's done in this PR #849

PR #635, opened 2017-10-23
Add Flake8 configuration and ignore E129

reszelaz:
As commented by @mrosanes in https://github.com/sardana-org/sardana/pull/551#issuecomment-338572275  our current code style check on the PR is more strict than PEP8. Relax the E129 error so it is coherent with what PEP8 says. Please keep in mind that when we accept this PR the checking commit should be removed from the history.


reszelaz:
Now it works as expected. The b09082d commit should be reverted before the integration.

cpascual:
Merged (after solving a trivial conflict)

PR #638, opened 2017-10-27
Removed uselles setting of label config in PMTV

reszelaz:
Setting of label config in PMTV has no effect - the widget works exactly the same
if the line is removed. This applies both to Taurus3 and Taurus4.


cmft:
reviewed the code. All seems ok for me.

reszelaz:
Thanks for the review! Did you also try it? It would be interesting to test it with taurus 3 and taurus 4.

cmft:
I tested it with both.

reszelaz:
Great! I will proceed then with the integration.

PR #639, opened 2017-10-27
Avoid "already in motion" error

rhomspuron:


reszelaz:
Thanks @rhomspuron for this contribution. In order to understand the problem one would need to read this https://github.com/sardana-org/sardana/issues/70#issuecomment-262973232 which refers to a similar issue but on the experimental channels. In case of motors the situation is exactly the same. In Sardana there exists two levels of protection from abusive actions: Tango state machine and the Sardana operation context. As it is now, the state changes MOVING -> ON before the operation context is released, allowing the user to move the motor again. Very fast execution of successive movements may reach a point where the state has already changed to ON but operation context is still locked. Making the requested motion to be rejected. The proposed change releases the operation context before setting new state.

Now, before merging this commit, I will amend the proper commit message and rewrite the history but for the future please provide some more information on the proposed change.

PR #642, opened 2017-11-06
Report error message if macros cannot be reloaded

mrosanes:
Report error message in spock if macros cannot be reloaded

### DIFF
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -507,6 +507,12 @@ class MacroManager(MacroServerManager):
 
         params = dict(module=m, name=module_name,
                       macro_server=self.macro_server, exc_info=exc_info)
+
+        # Dictionary for gathering macros with errors
+        macro_errors = {}
+        count_correct_macros = 0
+        count_incorrect_macros = 0
+
         if m is None:
             file_name = self._findMacroLibName(module_name)
             if file_name is None:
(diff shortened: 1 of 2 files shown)


teresanunez:
Hi @mrosanes, I have reviewed this. It works as expected: an error is reported if the macro parameters are not defined correctly, but only a comment about the different behaviour in case an error in a macro is due to wrong parameters (case 1) or to a python syntax (case 2).
In both cases the macro is in the MacroServer, we make changes and we add an error:
Case 1. relmac and relmaclib will report the error, and the macro is not more in the MS, for having it again, after correcting the error,  one needs to run relmaclib with the name of the library (relmac doesn't work because the macro is not there any more)
Case 2.   relmac and relmaclib will report the error, but the macro is still in the MS with the code of the last working version. And after correcting the error one can use relmaclib or relmac (since the macro is still there) for loading the new version.
Perhaps it is fine to have this two different behaviours ... but I would add to the error coming from wrong parameters, that the macro has to be added again using relmaclib (and not relmac).

mrosanes:
The main difference is that in Case 2, the error, is a Python error; and the full module contains this error; so it cannot be loaded, and the old module is kept. If modifications were done in other macros of the same library at the same time, they would neither be available.

However, in the case of an error in the parameter definition description, the module can be loaded because it does not contain python errors, but the macro contains an 'error' because it does not follow the pattern of a Sardana macros (e.g. missing parameter description, etc). In this case the correct macros are loaded correctly (even if they were modified), because the module as a whole does not contain Python errors; but, the macros containing errors in the parameter description etc, cannot be loaded. Thus, it is indicated that some macros could not be loaded correctly. 

I agree with specifying that the macros shall be added again using relmaclib. I will propose a new commit. Thanks for testing.

teresanunez:
Hi @mrosanes, thanks for the update. IMO it can be merged now.

PR #644, opened 2017-11-08
Add __getitem__ method to Record class

rhomspuron:
Allow to get the data from the record as a dictionary. It is possible to give the alias of the channel.

### DIFF
--- a/src/sardana/macroserver/scan/scandata.py
+++ b/src/sardana/macroserver/scan/scandata.py
@@ -30,8 +30,11 @@ __all__ = ["ColumnDesc", "MoveableDesc", "Record", "RecordEnvironment",
 
 import copy
 import math
+import socket
 
 from taurus.core.util.singleton import Singleton
+from taurus import Device, Attribute
+from taurus.core.taurusexception import TaurusException
 
 from sardana.macroserver.scan.recorder import DataHandler
 from threading import RLock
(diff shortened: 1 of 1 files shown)


rhomspuron:
One way to use on a macro is:
```
class getData(Macro):
    def run(self):
        ch_name = 'ct13'
        scan, _ = self.createMacro('ascan mot13 0 1 10 0.1')
        self.runMacro(scan)
        data = scan.data
        for idx, rc in data.items():
            self.info(rc[ch_name])
```

teresanunez:
Hi, I have tested this and it woks as expected.

rhomspuron:
Hi Teresa, 

Thanks for test it

src/sardana/macroserver/scan/scandata.py
```diff
@@ -186,6 +189,67 @@ def setComplete(self):
...
+        # check if the item is a device name or an attribute name
+        try:
+            chn_proxy = Device(item)
+        except TaurusException:
+            try:
+                chn_proxy = Attribute(item)
+            except TaurusException:
+                raise KeyError(item)
+
+        # TODO: Refactor this code when the data dictionary uses only taurus
```
cpascual: This TODO should be written before. It should cover also the check of whether the "item" is an attr or a dev name (because in Taurus4 we can know this without having to instantiate the object at all)

src/sardana/macroserver/scan/scandata.py
```diff
@@ -186,6 +189,67 @@ def setComplete(self):
...
+        # 4 full name
+
+        try:
```
cpascual: My suggestion is to refactor the whole thing as follows:
- remove  `__get_fullnames`
- create `__get_t3_name`  that does the t3-specific part of `__get_fullnames`
- create `__get_t4_name` that finds (without instantiating anything) whether the item is an attr or dev, and  then uses `validator.getNames(item)[0]` for fnding the full name
- create `__get_tango_name` (to support the special case of tango-urls without scheme name) and implement it as 

  ```
  n = __get_t4_name()
  if n startswith('tango://'):
      n = n[8:]
  return n
  ```

- Finally refactor this try block of ` __getitem__` to sequentially construct names with the above functions and try to get them from `data`

cpascual:
Merged, thanks!

PR #646, opened 2017-11-13
[WIP] Add support to GeneralHooks (#200)

### LINKED ISSUE #200: Hooks by default in scans (SF#354)
Hi,
       I copy here an e-mail that I sent some time ago, it has
 become now a request from our users.

   our beamline scientists want that all the scans run by default
 with hooks functions, not like it is now that it is necessary to define
 a new macro running an existing scan macro after adding hooks to it.
 They want to keep the original names for the current scan macros and
 simply define a function that will be run always like a hook.
 I think that the current implementation for the hooks works quite
 well, and offers flexibility for example for the arguments that have
 to be passed. A fixed function could have some problem there.
 So in my opinion one should keep what is implemented and try to
 add the requested feature that could be activated by an additional
 variable (set for example in the spock environment).
 What do you think about something like this?. 

                            Regards,
                                       Teresa

Original Ticket: [sardana/tickets/354](https://sourceforge.net/p/sardana/tickets/354)

reszelaz:
This PR is for development of the GeneralHooks feature as requested in #200.

TODOs:

- [x] Report and fix bug in Hookable.hooks.set (now it only resets the list with hooks and not the dictionary with hints) - @reszelaz
- [x] Before appending a hook to the macro object it would be nice to verify if this macro support a given hook place. This could be consulted with the `Hookable.getAllowedHookHints` method - @teresanunez 
- [x] Agree and implement the commodity macros e.g. `lsgenhook`, `addgenhook`, etc. - @teresanunez 
- [x] Test other use cases e.g. scan macros and hook macros with parameters - @teresanunez 
- [x] Fix setting of GeneralHooks environment in generalHooks.py testing module - @teresanunez 

Open questions:

- [x] Agree on GeneralHooks environment variable syntax (see #200 for discussion) - now it uses the programmatic API e.g. `senv GeneralHooks "[('MyHook1', ['my-hook-place'])]"`?
- [x] Extract GeneralHooks configuration at the very start of the macro/sequence execution (this may be a more general issue)?
- [x] Agree on what to do in case of an error during the general hook execution? Currently an error during the normal hook execution aborts the scan macro.
- [x] Add support to callables?
 
For testing I use this https://github.com/reszelaz/generalHooks. Ideally this should be converted into tests...

Fixes #200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ This file follows the formats and conventions from [keepachangelog.com]
   (`ApplyExtrapolation` environment variable) (#588)
 - `--file` option to sequencer - it allows to load a sequence file
   directly on the application startup moment (#283, #551)
+- Report error line number when loading a sequence from a txt file
+  fails (#114, #552)
 - Present available pools at the macroserver creation moment in the
   alphabetical order (#585, #586)
 - Present available doors at the spock profile creation moment in the
@@ -40,6 +42,8 @@ This file follows the formats and conventions from [keepachangelog.com]
   #615)
 - Selection of the master timer/monitor for each of the acquisition
   sub-actions (hardware and software) (#614)
+- Avoid "already involved in motion" errors due to wrong handling of
+  operation context and Tango state machine (#639)
 - Make the information about the element's instrument fully dynamic and
   remove it from the serialized information (#122, #619)
 - uct macro (#319, #627)
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -212,15 +212,21 @@ class Hookable(Logger):
         self._getHooks().append(hook_info)
         hook = hook_info[0]
         hints = hook_info[1]
-        self._getHookHintsDict()['_ALL_'].append(hook)
+        allowed_hookhints = self.getAllowedHookHints()
         if len(hints) == 0:
+            self._getHookHintsDict()['_ALL_'].append(hook)
             self._hookHintsDict['_NOHINTS_'].append(hook)
             return
         for hint in hints:
-            try:
-                self._hookHintsDict[hint].append(hook)
-            except KeyError:
-                self._hookHintsDict[hint] = [hook]
+            if hint in allowed_hookhints:
+                self._getHookHintsDict()['_ALL_'].append(hook)
+                break
+        for hint in hints:
+            if hint in allowed_hookhints:
+                try:
+                    self._hookHintsDict[hint].append(hook)
+                except KeyError:
+                    self._hookHintsDict[hint] = [hook]
 
     @propertx
     def hooks():
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -202,6 +202,26 @@ class Hookable(Logger):
         else:
             return self._getHookHintsDict().get(hint, [])
 
+    def appendHook(self, hook_info):
+        """Append a hook according to the hook information
+
+        :param hook_info: sequence of two elements, the first one is the hook
+            and its optional parameters/arguments, the second one is the list
+            of hints e.g. hook places
+        """
+        self._getHooks().append(hook_info)
+        hook = hook_info[0]
+        hints = hook_info[1]
+        self._getHookHintsDict()['_ALL_'].append(hook)
+        if len(hints) == 0:
+            self._hookHintsDict['_NOHINTS_'].append(hook)
+            return
+        for hint in hints:
+            try:
+                self._hookHintsDict[hint].append(hook)
+            except KeyError:
+                self._hookHintsDict[hint] = [hook]
+
     @propertx
     def hooks():
         def get(self):
--- a/src/sardana/macroserver/macros/env.py
+++ b/src/sardana/macroserver/macros/env.py
@@ -312,28 +312,17 @@ class defgh(Macro):
     """Define general hook. Without arguments add default ones """
     
     param_def = [
-        ['macro_name', Type.String, "default", 'Macro name with parameters. Ex.: "mv exp_dmy01 10"'],
-        ['hook_pos', Type.String, "default", 'Position or positions where the hook has to be executed. Ex.: "pre-scan", several positions "pre-scan,post-scan"'],
+        ['macro_name', Type.String, None, 'Macro name with parameters. Ex.: "mv exp_dmy01 10"'],
+        ['hook_pos', Type.String, None, 'Position or positions where the hook has to be executed. Ex.: "pre-scan", several positions "pre-scan,post-scan"'],
     ]
     
     def run(self, macro_name, hook_pos):
-        default_list = [('gh_pre_scan',['pre-scan']),
-                        ('gh_pre_move',['pre-move']),
-                        ('gh_pre_acq',['pre-acq']),
-                        ('gh_post_acq',['post-acq']),
-                        ('gh_post_move',['post-move']),
-                        ('gh_post_step',['post-step']),
-                        ('gh_post_scan',['post-scan'])]
-        if hook_pos == "default":
-            self.info("Defining general hooks with default names")
-            self.debug(default_list)
-            self.setEnv("_GeneralHooks", default_list)
-        else:
-            self.info("Defining general hook")
-            try:
-                macros_list = self.getEnv("_GeneralHooks")
-            except:
-                macros_list = []
+        
+        self.info("Defining general hook")
+        try:
+            macros_list = self.getEnv("_GeneralHooks")
+        except:
+            macros_list = []
             positions_split = hook_pos.split(",")
             positions = []
             for pos in positions_split:
--- a/src/sardana/macroserver/macros/env.py
+++ b/src/sardana/macroserver/macros/env.py
@@ -302,9 +302,29 @@ class lsgh(Macro):
     def run(self): 
         try:
             general_hooks = self.getEnv("_GeneralHooks")
-            self.output("General Hooks:")
+            default_dict = {'pre-scan': [],
+                            'pre-move': [],
+                            'pre-acq': [],
+                            'post-acq': [],
+                            'post-move': [],
+                            'post-step': [],
+                            'post-scan': []}
+            out = List(['Hook place', 'Hook(s)'])
             for hook in general_hooks:
-                self.output(hook)
+                name = hook[0]
+                places = hook[1]
+                for place in places:
+                    default_dict[place].append(name)
+            for pos in default_dict.keys():
+                pos_set = 0
+                for hook in default_dict[pos]:
+                    if pos_set:
+                        out.appendRow(["", hook])
+                    else:
+                        out.appendRow([pos, hook])
+                    pos_set = 1
+            for line in out.genOutput():
+                self.output(line)
         except:
             self.output("No general hooks")
 
@@ -323,6 +343,7 @@ class defgh(Macro):
             macros_list = self.getEnv("_GeneralHooks")
         except:
             macros_list = []
+            
         positions_split = hook_pos.split(",")
         positions = []
         for pos in positions_split:
--- a/src/sardana/macroserver/macros/env.py
+++ b/src/sardana/macroserver/macros/env.py
@@ -31,6 +31,7 @@ __docformat__ = 'restructuredtext'
 
 from taurus.console.list import List
 from sardana.macroserver.macro import Macro, Type, ParamRepeat
+from sardana.macroserver.msexception import UnknownEnv
 
 ##########################################################################
 #
@@ -303,29 +304,30 @@ class lsgh(Macro):
     def run(self):
         try:
             general_hooks = self.getEnv("_GeneralHooks")
-
-            out = List(['Hook place', 'Hook(s)'])
-            default_dict = {}
-            for hook in general_hooks:
-                name = hook[0]
-                places = hook[1]
-                for place in places:
-                    if place not in default_dict.keys():
-                        default_dict[place] = []
-                    if name not in default_dict[place]:
-                        default_dict[place].append(name)
-            for pos in default_dict.keys():
-                pos_set = 0
-                for hook in default_dict[pos]:
-                    if pos_set:
-                        out.appendRow(["", hook])
-                    else:
-                        out.appendRow([pos, hook])
-                    pos_set = 1
-            for line in out.genOutput():
-                self.output(line)
-        except:
+        except UnknownEnv:
             self.output("No general hooks")
+            return
+
+        out = List(['Hook place', 'Hook(s)'])
+        default_dict = {}
+        for hook in general_hooks:
+            name = hook[0]
+            places = hook[1]
+            for place in places:
+                if place not in default_dict.keys():
+                    default_dict[place] = []
+                if name not in default_dict[place]:
+                    default_dict[place].append(name)
+        for pos in default_dict.keys():
+            pos_set = 0
+            for hook in default_dict[pos]:
+                if pos_set:
+                    out.appendRow(["", hook])
+                else:
+                    out.appendRow([pos, hook])
+                pos_set = 1
+        for line in out.genOutput():
+            self.output(line)
 
 
 class defgh(Macro):
@@ -353,7 +355,7 @@ class defgh(Macro):
         self.output(macro_name)
         try:
             macros_list = self.getEnv("_GeneralHooks")
-        except:
+        except UnknownEnv:
             macros_list = []
 
         hook_tuple = (macro_name, position)
@@ -376,7 +378,7 @@ class udefgh(Macro):
     def run(self, macro_name, hook_pos):
         try:
             gh_macros_list = self.getEnv("_GeneralHooks")
-        except:
+        except UnknownEnv:
             return
 
         if macro_name == "all":
--- a/src/sardana/macroserver/macros/env.py
+++ b/src/sardana/macroserver/macros/env.py
@@ -30,7 +30,7 @@ __all__ = ["dumpenv", "load_env", "lsenv", "senv", "usenv",
 __docformat__ = 'restructuredtext'
 
 from taurus.console.list import List
-from sardana.macroserver.macro import *
+from sardana.macroserver.macro import Macro, Type, ParamRepeat
 
 ##########################################################################
 #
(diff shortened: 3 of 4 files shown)


teresanunez:
Hi @reszelaz, thanks a lot for starting this. 
About the macros, they already exist. Our users don't need to know anything about the environment
variables, they set the general hooks and get all the information about what is set via macros.

teresanunez:
I have tried to test it with your generalHooks.py file, but I get the error:

[START] runMacro Macro 'MyMacroWrapper() -> 45f19320-c930-11e7-93de-901b0e40a03c'
In MyMacroWrapper
Sending exception event
[ENDEX] (MacroServerException) runMacro MyMacroWrapper
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/msmacromanager.py", line 1314, in runMacro
    for step in macro_obj.exec_():
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/macro.py", line 2290, in exec_
    res = self.run(*self._in_pars)
  File "/home/tnunez/generalHooks_zibi/generalHooks.py", line 43, in run
    my_macro, _ = self.createMacro("MyMacro")
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/macro.py", line 333, in new_fn
    ret = fn(*args, **kwargs)
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/macro.py", line 1128, in createMacro
    return self.prepareMacro(*pars)
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/macro.py", line 333, in new_fn
    ret = fn(*args, **kwargs)
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/macro.py", line 1201, in prepareMacro
    return self.executor.prepareMacro(args, init_opts, kwargs)
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/msmacromanager.py", line 1114, in prepareMacro
    self._prepareGeneralHooks(macro_obj)
  File "/home/tnunez/sardana_zibi_gh/src/sardana/macroserver/msmacromanager.py", line 964, in _prepareGeneralHooks
    for hook_info, hook_places in general_hooks:
ValueError: too many values to unpack

An error occurred while running Macro 'MyMacroWrapper() -> 45f19320-c930-11e7-93de-901b0e40a03c':
too many values to unpack
Job ended (stopped=False, aborted=False)


It could be that I am setting something wrong ...

teresanunez:
I change in msmacromanager.py the line:

for hook_info, hook_places in general_hooks:

to:
for hook_info, hook_places in general_hooks.items():

and now it works.

teresanunez:
I see now the result of running the generalHooks.py script and the hook that should be added as generalHook is not called. The outuput is:

In MyMacro
In MyHook2
In MyMacroWrapper
In MyMacro
In my_hook2

It never comes In MyHook1. Should it already work?

teresanunez:
The environment variable generalHooks is not set correctly in the generalHooks.py script, this is the reason of the errors reported above: 

- ValueError: too many values to unpack
- general hook not executed

Setting the environment correctly, ex.:

senv GeneralHooks "[('MyHook1', ['my-hook-place'])]"

the code in msmacromanager.py is correct like it is and the general hooks are executed.

teresanunez:
Hi @reszelaz. I have tried with scans and parameters in the hook and it works. You told me about this yesterday, but I didn't realize it until today ... what is different with respect to my implementation, is that
the GeneralHooks environment is now a list, and in my case is was a dictionary. So the macros that I have doesn't work. They have the same problem than your generalHooks.py example. I will try to modify both things, but only to be sure: do we want to keep the environment variable as a list instead of a dictionary?.

teresanunez:
For the tests of hooks with parameters I have directly set the environment from
spock with the syntax:

senv GeneralHooks "[('umv mot14 1', ['pre-acq'])]"

and it works. As I said above, the macros I had for setting this don't work because I set
 a dictionary and not a list.

reszelaz:
> senv GeneralHooks "[('umv mot14 1', ['pre-acq'])]"

Great! I'm positivelly surprised that this works. The explanation is that the `ExecMacroHook` uses internally the `Macro.execMacro` method which supports multiple input syntaxis

Could you check as well the following two other forms:
`senv GeneralHooks "[('umv', ['mot14', 1], ['pre-acq'])]"`
and 
`senv GeneralHooks "[('umv [[mot14 1]]', ['pre-acq'])]"`

reszelaz:
> do we want to keep the environment variable as a list instead of a dictionary?

The pros and cons of both possibilites are listed along the https://github.com/sardana-org/sardana/issues/200 thread. @cpascual who originally proposed the dictinary format seems to be also ok with the list format (the list format resembles more the programmatic API): https://github.com/sardana-org/sardana/issues/200#issuecomment-344173056. If no one speaks against the list, I would go into this direction.

teresanunez:
Hi @reszelaz, the other two forms do not work:

Door_gh_1 [1]: senv GeneralHooks "[('umv [[mot14 1]]', ['pre-acq'])]"
GeneralHooks = [('umv [[mot14 1]]', ['pre-acq'])]

Door_gh_1 [2]: %ascan mot13 1 10 3 1
ScanDir is not defined. This operation will not be stored persistently. Use Use "expconf" (or "senv ScanDir <abs directory>") to enable it
Scan #28 started at Wed Nov 15 11:47:53 2017. It will take at least 0:00:05.469694
 #Pt No    mot13      ct13      ct14      ct15      ct16       dt   
Scan #28 ended at Wed Nov 15 11:47:53 2017, taking 0:00:00.011643. Dead time 100.0% (motion dead time 85.6%)
An error occurred while running Macro 'ascan(mot13, 1.0, 10.0, 3, 1.0) -> 6e764a12-c9f2-11e7-9e7c-901b0e40a03c':
Moveable with name [[mot14 does not exist

Door_gh_1 [3]: senv GeneralHooks "[('umv', ['mot14', 1], ['pre-acq'])]"
GeneralHooks = [('umv', ['mot14', 1], ['pre-acq'])]

Door_gh_1 [4]: %ascan mot13 1 10 3 1
Error: too many values to unpack

Door_gh_1 [5]:

reszelaz:
I have just rewritten the history of this branch cause previously the comits had wrong author.

teresanunez:
HI @reszelaz, I have created three macros for handling the general hooks:

- add general hook in one or more positions
- remove general hook from all or one position
- list general hooks

Before sending the file we should I agree in the name of the library and the name of the macros.
Here the possibilities that I propose:

Library name:
1. Add it to the standard.py library file
2. Add it  to the scan.py library file
3. Create a new library file, ex. generalhooks.py

I think I would go for 1. since the idea is use then not only for scans but for any future Hookable macro, so one can consider them as general. Creating a new file I don't think that is necessary.

Macro names:
1. lsgh, addgh, delgh
2. gh_list, gh_add, gh_delete

(using 'gh' as short form of 'general hooks' could also be discussed, but I think it is the easiest and
most straight forward way)
Solution 1 follows the 'standard' sardana nomeclature, the solution 2 has the advantage that typing
gh_ one can see the macros available for handling the general hooks.

Please give your opinion about the proposed solutions or add new ones. Many thanks.

teresanunez:
Hi @reszelaz, I have fixed the problem with the generalHooks.py module. The line:

door.runMacro("senv GeneralHooks {'my-hook-place':'MyHook1'}",
                  synch=True)

has to be replaced by:


door.runMacro("senv GeneralHooks [(\"MyHook1\"  [\"my-hook-place\"])]",
                  synch=True)

If you agree, could you correct it or should I make a PR in your repo?

reszelaz:
Great! Done, it is pushed.

I think that by mistake you wrote double quotes extenrally and internally what gives a SyntaxError. I've changed it to used single quotes internally and it works.

> door.runMacro("senv GeneralHooks [("MyHook1" ["my-hook-place"])]",
synch=True)

reszelaz:
Regarding the macros and module names.

Personally I would use the macro names format as we have used up to now, so names: lsgh, addgh, delgh. For adding and deleting we could also consider defgh and udefgh...

Another alternative for the module location: the env module. It already stores macros for the ViewOptions which implementation is somehow similar to the GeneralHooks.

Now thinking, I would also rename the environment variable to _GeneralHooks (similar to the _ViewOptions). This way we give a message to the user that this variable somehow forms a part of the internal configuration (changing its format in the future would be eaiser for developers). At the end I think that the user should work with the commodity macros.

teresanunez:
Hi @reszelaz, the problem with the internal double quotes was due to the formatting of these messages in github. I didn't realize about the change that was automatically done. It works with single quote or with double quote preceded by slash (this was like I put it). Thanks for taking care of that.

teresanunez:
About the naming of macros and module. I agree with you. I would then make following:

- Add the macros to env.py
- Use _GeneralHooks instead of GeneralHooks
- Use defgh, udefgh, lsgh as macro names

Should I make a PR to your repo or send you the modified file and you added to this PR?

teresanunez:
I have commited the file env.py with the new macros to this PR.

teresanunez:
Hi @reszelaz, I have added the checks of the allowed hints. As for the macros, it is open to discussion.

teresanunez:
I do not solve the pep8 errors until the changes are accepted ... It is really a pain,
they are never solved automatically and I can not find the reason of the fail.

reszelaz:
Great! Thanks for the macros. I already have some first feedback. When trying to print the help of the `defgh` macro, by mistake I typed (instead of `?` I typed `>`):

```
Door_zreszela_1 [2]: %defgh>
Defining general hooks with default names
```

What already lead to the definition of some default general hooks. Since I do not have this macros in my Sardana system then the subsequent execution of the scan will fail. I think that this may cause some trouble for the users. I understand that the default general hooks comes from your internal convention. Do you think that it would be possible to move it to your internal macro and leave the generic macros without the default general hooks?

Now some cosmetic comments:
* What do you think to list the general hooks by hook place? I mean that the first column is the hook place and then the macros that would be executed in that hook place?
* What do you think to use the `List` class (`taurus.console.list.List`) for pretty print of general hooks. We already use it on other list macros.
* `defgh` could use repeat parameter for the hook places?

Tomorrow I will continue with the review. And I will also reply to your feedback about the flake8 error. Sorry about them.

teresanunez:
Hi @reszelaz, many thanks for your comments. I will work on that. About the default names, yes, I will
 take it out from the sardana macros and keep it in a private desy one.

reszelaz:
Just thinking about the best output of the lsgh macro. In case we would have multiple macros attached to the same hook place, having a single line with in the table could become unreadable. Don't you think so? What about having a table like this:

```
Door_zreszela_1 [18]: lsgh
    Hook place   Hook(s)
 -------------- -------
      pre-acq    open_shutter
                 ct 10
     post-acq    clos_shutter
```
I don't know if it is possible to achieve something like this with the `List` class from taurus.

teresanunez:
Hi @reszelaz , thanks, yes, I agree with this output. I will test with the taurus list class, let's see.

teresanunez:
Hi @reszelaz, only one comment ... setting the general hooks and also the environment _GeneralHooks are based on a function name and the positions in which it will be used, butnow
 for listing the general hooks we use the other way around. I think it is better to list it like  this, but
 it could be a bit confusing to set it in one way and get it listed in another way. Well, I don't know, only
to have it in mind.

teresanunez:
Hi @reszelaz, I have implemented the issues you pointed out. You can have a look. Another comment, 
before I was only allowing the hook positions defined for the scans ... now I allow any position, so the
users can define not existing hook positions and they will be accepted, but they will never run ... Like this we have the flexibility of using this in any defined Hookable macro, but it has the disadvantage that if somebody define a position thinking that is already existing (ex. somebody misstype pre-mv as pre-move) it could be hard to know why the hook macro is not running in the 'defined' position.

reszelaz:
Regarding to the flake8 errors, the ones that make the [check fail](https://travis-ci.org/sardana-org/sardana/jobs/305205929) are:
- W293 blank line contains whitespace
- E302 expected 2 blank lines, found 1
- W291 trailing whitespace
- E501 line too long (101 > 79 characters)
- E231 missing whitespace after ','
- W391 blank line at end of file

To see the statictics of the errors in this file one could do:
```
zreszela@pc255:~/workspace/sardana$ flake8 -qq --count --statistics src/sardana/macroserver/macros/env.py
2     E302 expected 2 blank lines, found 1
3     E402 module level import not at top of file
6     E501 line too long (85 > 79 characters)
1     F403 'from sardana.macroserver.macro import *' used; unable to detect undefined names
26    F405 Macro may be undefined, or defined from star imports: sardana.macroserver.macro
1     W291 trailing whitespace
10    W293 blank line contains whitespace
1     W391 blank line at end of file
50
```
And to fix them automatically one could do:
```
autopep8 -a -a -i --select=W293,E302,W291,E231,W391 src/sardana/macroserver/macros/env.py
```
This could eliminated all the errors:
```
zreszela@pc255:~/workspace/sardana$ flake8 -qq --count --statistics src/sardana/macroserver/macros/env.py
3     E402 module level import not at top of file
6     E501 line too long (85 > 79 characters)
1     F403 'from sardana.macroserver.macro import *' used; unable to detect undefined names
26    F405 Macro may be undefined, or defined from star imports: sardana.macroserver.macro
36
```
Note that the `E501 line too long (101 > 79 characters)` was excluded from the automatic correciton on purpose. The `autopep8` is not perfect when correcting this error. Similarly as we have decided to not apply it automatically in #54 I do not recommend to do it here. The result may be less readable than the original code. It is better to correct this errors manually.

Be aware that before pushing to the PR branch in order to check if the flake8 will pass, one could test it locally using the ci/flake8_diff.sh script.

If you have problems with all this I can give you a hand or even take care of this corrections. And I think that it fine to do it as the last thing before the integration.

reszelaz:
> setting the general hooks and also the environment _GeneralHooks are based on a function name and the positions in which it will be used, butnow for listing the general hooks we use the other way around.

I would treat the case of the `_GeneralHook` environment variable as a separate issue. It is an internal data structure and its syntax was chosen like this to faciliatate the implementation.

Regarding the `defgh` macro. First of all, its syntax is similar to the programatic one - first the hook and then the hook places. Second of all, this way the same hook could be attached to multiple hook places. We could achieve the second point when inverting the order of parameters (first hook places and then the hook) by using the extended spock syntax for the repeat parameters e.g. `defgh [pre-acq post-scan] "mv mot01 0`, but then for the spock users we would force them to use the extended spock syntax. I think we could get some complains about it. So I think that what we have now, first hook and then hook places, is the best choice.

Regaring the `lsgh` output, I have a feeling that listing first the hook place and then the hooks will be more intuitive to the users when seeing the overall configuration of general hooks. But this is just my impression.

reszelaz:
> before I was only allowing the hook positions defined for the scans ... now I allow any position, so the
users can define not existing hook positions and they will be accepted, but they will never run ... Like this we have the flexibility of using this in any defined Hookable macro, but it has the disadvantage that if somebody define a position thinking that is already existing (ex. somebody misstype pre-mv as pre-move) it could be hard to know why the hook macro is not running in the 'defined' position.

This is true. But I think that we can not do too much about it. The feature of general hooks is very generic and without knowing the macro where we want to attach the hook it is impossible to validate if the hook place exist. As a solution to this problem one could define a user macro for defining general hooks for scans that would do the validation before calling `defgh`.

reszelaz:
In commit 399c2c0 I have used ParamParser (used in spock) to parse the hook information (macro name and its parameters). Now defining hooks containing the following parameters works as well:
- string parameters containing spaces encapsulated in quotes
- extended repeat parameter syntax with square brackets

I have tested the following configuration:
```
Door_zreszela_1 [6]: lsgh
   Hook place          Hook(s)
 ------------ --------------------
     pre-scan       mv [[mot02 9]]
                            ct 0.1
                               lsm
                       mv mot03 10
               Print 'Hello world'
```

That could be achieved by:
```
defgh "mv [[mot02 9]]" pre-scan
defgh "ct 0.1" pre-scan
defgh lsm pre-scan
defgh "mv mot03 10" pre-scan
defgh "Print 'Hello world'" pre-scan
```
All of them works correctly.

The only thing that still does not work in to have in the `_GeneralHooks` variable hook information stored in a more pythonic way - instead of the spock syntax string, the python objects.

teresanunez:
Hi @reszelaz, perfect, many thanks, I have applied the commit and checked the different ways of setting the macro hooks and it works nicely.

teresanunez:
I think the information you wrote above about how defgh can be used could be added to the description of the macro. I will add it.

reszelaz:
Solution for first point from the TODOs list is proposed in #655.

reszelaz:
I've just fixed the PEP8 errors. Steps that I've done:
* `autopep8 -a -a -i --select=W293,E302,W291,E231,W391 src/sardana/macroserver/macros/env.py`
* Manually corrected two E501:
```
src/sardana/macroserver/macros/env.py:343:80: E501 line too long (96 > 79 characters)
        ['macro_name', Type.String, None, 'Macro name with parameters. Ex.: "mv exp_dmy01 10"'],
                                                                               ^
src/sardana/macroserver/macros/env.py:371:80: E501 line too long (87 > 79 characters)
        ['hook_pos', Type.String, "all", 'Position to undefine the general hook from'],
```
* Manually corrected three E722:
```
src/sardana/macroserver/macros/env.py:327:9: E722 do not use bare except'
        except:
        ^
src/sardana/macroserver/macros/env.py:356:9: E722 do not use bare except'
        except:
        ^
src/sardana/macroserver/macros/env.py:379:9: E722 do not use bare except'
        except:
        ^
```
Now it is clean.

reszelaz:
Regarding the open questions from the initial comment:

>  Agree on GeneralHooks environment variable syntax (see #200 for discussion) - now it uses the programmatic API e.g. senv GeneralHooks "[('MyHook1', ['my-hook-place'])]"?

I think that this one we could consider as closed. As it is now, the environment variable is not foreseen for manual manipulation - its name was changed to the *private* scope (starts with an underscore). A set of macros was provided for dealing with the general hooks configuration. So it's syntax is not so important anymore. But anyway, from the programming point of view I think that we have made an optimal selection for its syntax.

>  Extract GeneralHooks configuration at the very start of the macro/sequence execution (this may be a more general issue)?

Since the configuration is stored in the environment and as on today the environment is totally dynamic. This is the opposite to the bash environment where the starting process makes a snapshot of the current state of the environment and its child processes inherits it. So if we want to change or extend the environment behavior I think that we could make an extra feature request for that and deal it there.

>  Agree on what to do in case of an error during the general hook execution? Currently an error during the normal hook execution aborts the scan macro.

Now the macro basically fails with an error. I think that it is fine, but please comment if you have another opinion.

>  Add support to callables?

If really needed, this could be implemented a posteriori with another feature request.

teresanunez:
Hi @reszelaz, thanks for your comments. I agree with you, we can accept this like it is and further improvements can be done in new feature requests.

reszelaz:
Merge this PR as agreed on the last [follow-up meeting](https://github.com/sardana-org/sardana-followup/blob/master/20180125/MINUTES.md) after marking the new API as _experimental_.

PR #648, opened 2017-11-15
On find master, check if the timer enabled

droldan:
Additional protection to Pull Request [#614)[https://github.com/sardana-org/sardana/pull/614]

On find the master timer, it should to take into account that the element must be enabled.


reszelaz:
With this change we have relaxed the process of selection of the master timer/monitor. If the first timer/monitor is disabled it looks for next one (and so on...) and tries to select them for the master timer/monitor.

But we should start thinking on how to improve the expconf widget cause now two features:
* slection of the master timer/monitor
* changing the order of columns in the scan record
are based on the order of channels in the measurement group. Conflicts between these two may arrise.

Issue #427, opened 2017-02-16
Default values in a param repeat for a whole repetition are not functional


sagiss:
Default values in a param repeat for a whole repetition are not functional (they are not taken into account).

Given a parameter definition, with default values specified; as:
</pre>
param_def = [
    {'name': 'motor_pos_list',
     'default_value': ["mot01", 50],
     'type': [
        {'name': 'motor',
         'default_value': "mot99"
         },
        {'name': 'pos',
         'default_value': 100
         }
        ]}
    ]
</pre>

The default values for the whole repetition are never used:  ["mot01", 50]
Instead the default values for the single elements are always used: "mot99" and 100.

If the default values for the single elements are not defined, the default values for the whole repetition are not used neither. They seem to be always ignored.

PR #649, opened 2017-11-15
Add example of macro as a hook

reszelaz:


teresanunez:
Hi @reszelaz, I have tried your example macro, but I only get it running if the macro to be run as a hook does not admit any argument. For example, it works for 'ct' but not for 'lsm'. It also does not work for macros that need arguments, but well, this could be said in the example.

reszelaz:
I have also observed that. Just to be more precise I think tha it does not work with macros with repeat parameters. `ct` macro has one parameter - integ_time. The lsm has one parameter - filters which is a repeat parameter.

teresanunez:
Yes, you are right. It works with macros with no parameters or parameters with default values if they are not repeat parameters.

reszelaz:
Great! I will work on improving it. I think the bug also causes problems in https://github.com/sardana-org/sardana/pull/646#issuecomment-344556082.

reszelaz:
> Yes, you are right. It works with macros with no parameters or parameters with default values if they are not repeat parameters.

This is related to #427. Workaround proposed in #654 fixes this issue.

I have also added comments in the macro on how to pass parameters to the hook macro. 

@teresanunez  what do you think about it now?

teresanunez:
Hi @reszelaz, the problem running with macros with parameters reported above is still not solved,
isn't it?. Sorry, I do not see any difference affecting the performance or the explanation how to use.
I guess I miss what you were trying to do.

reszelaz:
Have you tried with #654 applied?

teresanunez:
HI @reszelaz, ok, so this was where you wanted to solve: macros with default parameters when they were ParamRepeat ... yes, this is solved. I was checking the case of giving parameters to macros. You add examples about that for ExecMacroHook ... but I do not see how to use the hooked_scan_with_macro setting macros with parameters. I have tried several ways but none work. Should it work?, in case no, it is not a problem for now ... but then I think that the information you add now to the file is not relevant or needed there ... it could be added where you define ExecMacroHook instead.

reszelaz:
> but I do not see how to use the hooked_scan_with_macro setting macros with parameters. I have tried several ways but none work. Should it work? 

You are right. Using the `hooked_scan_with_macro` macro example it is not possible to pass parameters to the hook macro. I don't see any easy way of adding an extra parameter to this macro that would serve to pass any arbitrary parameter to the hook macro. In the macro description it already says that the hook macro does not accept parameters.

> but then I think that the information you add now to the file is not relevant or needed there ... it could be added where you define ExecMacroHook instead.

Just to clarify, you mean that these examples:
```
# In order to attach macro with parameters pass all of them in form of
# a list (repeat parameters are allowed) e.g.
# "ExecMacroHook(self, ["ct", 0.1])"
# "ExecMacroHook(self, ["mv", [["mot01", 0], ["mot02", 0]]])"
```
could be moved to the `ExecMacroHook` class docstring. I like it! Let me know and I will move them there.

teresanunez:
Hi @reszelaz, perfect, for me this change would be nice and after it I would say that the PR could be merged.

Issue #643, opened 2017-11-06
Optionally disable readout of motors positions in continuous scans


reszelaz:
When in the following description we talk about the continuous scan this refers to the “ascanct & co.” scans.

In some cases, having the read position in continuous scans is not optimal. The read is executed by software when a software trigger arrives. If the synchronization description is too demanding, and the software trigger can not follow it, this would imply scan records without the positions. Of course interpolation of this missing values is possible.

This may affect the online plot tool, more precisely its x axis, in the following ways:

- If the interpolation is disabled, the NaN reported in the position column will make the ScanTrend show curves with jumps to positions like “-Inf”
- If the interpolation is enabled, the “zero order” interpolation will plot multiple channels results on the same x value.

While having the read values is desired in most of the cases, especially when the software synchronization is in use, the theoretical ones are still interesting, for example, for the very fast scans. I propose to, by default, execute the position read and add an environment variable to disable the read and provide theoretical values instead.

mrosanes:
The proposed option seems right to me.

guifrecuni:
Of course, in case of hardware trigger by position, I'd use the already computed position list used to configure the hardware.

In case of hardware trigger by time, we could use theoretical values always, or read as much as possible and differentiate somehow readed and theoretical positions. I'd skip repeated position values for different triggered measurements.

reszelaz:
After implementing this feature we should revert the 2f04614 in order to reintroduce motor position readouts introduced with #590. More information about reverting merges and reverting reverted merges can be found in this [document](https://git-scm.com/blog/2010/03/02/undoing-merges.html)

PR #650, opened 2017-11-16
Revert PR #590: read motor positions in continuous scans

reszelaz:
This PR temporarily reverts integration of #590 until #643 gets properly implemented.


reszelaz:
Since I'm the one who merged the previous PR request I would like to now rever this decission. I think we should wait for #643 to be integrated as well.

PR #651, opened 2017-11-16
Force the writing of the file on the disk.

rhomspuron:
The scientists have reported a delay between acquisition and the data available in SPEC file. The data is saved on folder which are mounted on NFS. The Pool and the MacroServer are running on a industrial PC and the spock and PyMCA on a workstation.  

PyMca refresh button the points data of the file is not updated correctly while the scan is writing the data. if we  run a 'watch tail -1 scanfile.dat' on the industrial PC, we can see how the scan file is written at each scan point.

Initially we tried to replicate the problem with bash:
```
while true; do date && sleep 1 ; done >file.txt
```
...but this worked fine (delay <3 seconds). We could do "tail -f file.txt" from a separate PC and it was refreshing properly. So we figured the problem was something that could be seen from the python code:
```
#!/usr/bin/env python

import time
import os

file = '/beamlines/bl22/controls/tmp/test.txt'

if _name=='main_':
    with open(file, 'w') as f:
        for i in range (1000):
           f.write('{0}\n'.format(time.time()))
           print 'sleeping'
           f.flush()
           os.fsync(f.fileno())
           time.sleep(1)
```
Without the call to os.fsync, the buffer is not written to disk (NFS in this case) and the client running on a separate machine does not see the new data. Even with this code there is a delay of 3 seconds caused by NFS cache semantics, but this is well withing tolerable limits for the user.

We later did a similar test directly using the sardana code and the proposed fix indeed solves the problem and PyMCA refreshes the data immediately.

Finally, here are some interesting links with miscellaneous info about the problem of buffering when writing data in python along with the subtleties of file buffering at glibc and OS levels:

https://stackoverflow.com/questions/7127075/what-exactly-the-pythons-file-flush-is-doing

https://blog.gocept.com/2013/07/15/reliable-file-updates-with-python/

https://stackoverflow.com/questions/9824806/how-come-a-file-doesnt-get-written-until-i-stop-the-program

https://docs.python.org/2/library/os.html  (check os.fsync)

### DIFF
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -222,6 +222,7 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
         self._createPreScanSnapshot(env)
 
         self.fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _compression(self, shape, compfilter='gzip'):
         """
@@ -303,6 +304,7 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
             else:
                 self.debug('missing data for label %r', dd.label)
         self.fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _endRecordList(self, recordlist):
 
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -222,7 +222,6 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
         self._createPreScanSnapshot(env)
 
         self.fd.flush()
-        os.fsync(self.fd.fileno())
 
     def _compression(self, shape, compfilter='gzip'):
         """
@@ -304,7 +303,6 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
             else:
                 self.debug('missing data for label %r', dd.label)
         self.fd.flush()
-        os.fsync(self.fd.fileno())
 
     def _endRecordList(self, recordlist):
 
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -169,6 +169,7 @@ class FIO_FileRecorder(BaseFileRecorder):
         self.fd.write(outLine)
 
         self.fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _writeRecord(self, record):
         if self.filename is None:
@@ -187,6 +188,7 @@ class FIO_FileRecorder(BaseFileRecorder):
 
         fd.write(outstr)
         fd.flush()
+        os.fsync(self.fd.fileno())
 
         if len(self.mcaNames) > 0:
             self._writeMcaFile(record)
@@ -387,6 +389,7 @@ class SPEC_FileRecorder(BaseFileRecorder):
         self.fd = open(self.filename, 'a')
         self.fd.write(header % data)
         self.fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _prepareMultiLines(self, character, sep, items_list):
         '''Translate list of lists of items into multiple line string
@@ -476,6 +479,7 @@ class SPEC_FileRecorder(BaseFileRecorder):
         fd.write(outstr)
 
         fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _endRecordList(self, recordlist):
         if self.filename is None:
@@ -650,6 +654,7 @@ class NXscan_FileRecorder(BaseNAPI_FileRecorder):
         self._createPreScanSnapshot(env)
 
         self.fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _createPreScanSnapshot(self, env):
         # write the pre-scan snapshot in the
@@ -722,6 +727,7 @@ class NXscan_FileRecorder(BaseNAPI_FileRecorder):
             else:
                 debug("missing data for label '%s'", dd.label)
         fd.flush()
+        os.fsync(self.fd.fileno())
 
     def _endRecordList(self, recordlist):
 
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -654,7 +654,6 @@ class NXscan_FileRecorder(BaseNAPI_FileRecorder):
         self._createPreScanSnapshot(env)
 
         self.fd.flush()
-        os.fsync(self.fd.fileno())
 
     def _createPreScanSnapshot(self, env):
         # write the pre-scan snapshot in the
@@ -727,7 +726,6 @@ class NXscan_FileRecorder(BaseNAPI_FileRecorder):
             else:
                 debug("missing data for label '%s'", dd.label)
         fd.flush()
-        os.fsync(self.fd.fileno())
 
     def _endRecordList(self, recordlist):
 


reszelaz:
Thanks @rhomspuron for this extensive investigation!

I remember that in the past a simple flush was enough to make the file written on the NFS and the scientists were able to monitor the scan results with the PyMCA on the workstation. Could it be that the behavior is somehow sensitive to the NFS configuration?

The change in the code is somehow simple. I don't see problems with its integration. I just wonder if the "os.fsync" is a blocking call i.e. will it block until all the data of the record is written on the disk? Have you tried if it has any impact ot the scan performance?

I also see that you were trying to modify the H5 recorders but finally you removed this change? Are these recorders also affected? Or maybe the h5 libraries are making the equivalent to os.fsync intenrally?

I would also like to raise a question to the Community. Do you have similar setusp as we do: two PCs and ScanFile on the NFS (see the PR message)? Have you experience this behavior? Did your users complain about it?

daneos:
At Solaris we also had some problems with scan files on NFS. Our setup involved Sardana server, a workstation, and a separate NFS server.
With such setup the problem was Sardana was freezing if the NFS was unavailable (due to network error, or anything else), and was frozen until NFS access was restored.
We fixed this problem by storing scan files locally on the Sardana server and sharing them via NFS from there.

I don't know about the delays in PyMCA since we use a dedicated solution for live plotting, that reads from Door output rather than scan files.

teresanunez:
At DESY we don't suffer from that issue, since we do not read ascii files for an 'online' display.
We also have an own tool for displaying the data read from the Door.

rhomspuron:
Hi @reszelaz, 

Answering to your questions:

> Could it be that the behavior is somehow sensitive to the NFS configuration? 

Yes, when the program executes the flush command it copies the internal buffer of the file to the OS buffer, but it not mean that the file will be saved, for that reason it does not solve the problem. The NFS has a mode to synchronize periodically the files, but this it is not the default mode and it does not work on all OS versions. 

> I just wonder if the "os.fsync" is a blocking call i.e. will it block until all the data of the record is written on the disk? 

Yes it is a blocking call, it will block until the buffer of the system is written on the disk. 

> Have you tried if it has any impact on the scan performance? 

We are using at CLAESS (3 months ago) and it does not impact on the scans (step and continuous)

> I also see that you were trying to modify the H5 recorders but finally you removed this change? Are these recorders also affected? Or maybe the h5 libraries are making the equivalent to os.fsync internally? 

I changed this because the H5 file object does not give the file_id in the same way to the normal file object, we should investigate how to get the file_id. The NXscan_FileRecorder has the same problem it does not update the file. The CLAESS users do not use the H5 file with PyMCA, for that reason I did not investigate how to do it.

reszelaz:
Great! Thanks for the feedback.

>>Have you tried if it has any impact on the scan performance?

>We are using at CLAESS (3 months ago) and it does not impact on the scans (step and continuous)

I'm worrying about the step scan performance and more exactly about the dead time. Since in the step scan the record is added in the `SScan.stepUp` method, and all recorders are handled sequentially, the blocking `os.fsync` could add some extra dead time.

Since it seems that not all users are interested in having the data inmediatelly in the file, we could think on making it configurable. This way these users would not be affected by this change. Of course I wound not bother about this if the extra time is negligable.

Since you are already using this solution, could you please provide some comparison of the step scan time performance with the `os.fsync` and without?

rhomspuron:
Hi @reszelaz,

I did some scans with/without the patch. 

With the patch
```
timeit -n 10 %ascan mot13 0 10 10 0.1
Operation will be saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1056 started at Mon Dec 18 10:16:47 2017. It will take at least 0:00:01.120000
 #Pt No    mot13      ct13      ct14      ct15      ct16       dt   
   0         0        0.1       0.2       0.3       0.4     0.0132408
   1         1        0.1       0.2       0.3       0.4     0.375578
   2         2        0.1       0.2       0.3       0.4     0.528867
   3         3        0.1       0.2       0.3       0.4     0.656514
   4         4        0.1       0.2       0.3       0.4     0.786724
   5         5        0.1       0.2       0.3       0.4     0.905468
   6         6        0.1       0.2       0.3       0.4     1.02351 
   7         7        0.1       0.2       0.3       0.4     1.15227 
   8         8        0.1       0.2       0.3       0.4     1.28063 
   9         9        0.1       0.2       0.3       0.4     1.40432 
   10        10       0.1       0.2       0.3       0.4      1.5334 
Operation saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1056 ended at Mon Dec 18 10:16:49 2017, taking 0:00:01.655184. Dead time 33.5% (motion dead time 4.6%)
....
10 loops, best of 3: 1.49 s per loop

timeit -n 10 %ascanct mot13 0 10 10 0.1
Operation will be saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1119 started at Mon Dec 18 10:24:18 2017. It will take at least 0:00:00
 #Pt No    mot13      ct13      ct14      ct15      ct16       dt   
Motor positions and relative timestamp (dt) columns contains theoretical values
   0         0        0.1       0.2       0.3       0.4       0.01  
   1         1        0.1       0.2       0.3       0.4       0.11  
   2         2        0.1       0.2       0.3       0.4       0.21  
   3         3        0.1       0.2       0.3       0.4       0.31  
   4         4        0.1       0.2       0.3       0.4       0.41  
   5         5        0.1       0.2       0.3       0.4       0.51  
   6         6        0.1       0.2       0.3       0.4       0.61  
   7         7        0.1       0.2       0.3       0.4       0.71  
   8         8        0.1       0.2       0.3       0.4       0.81  
   9         9        0.1       0.2       0.3       0.4       0.91  
   10        10       0.1       0.2       0.3       0.4       1.01  
Correcting overshoot...
Operation saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1119 ended at Mon Dec 18 10:24:20 2017, taking 0:00:01.253341. Dead time 100.0% (motion dead time 100.0%)
Operation will be saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
10 loops, best of 3: 1.35 s per loop
```
Without the patch
```
timeit -n 10 %ascan mot13 0 10 10 0.1
Operation will be saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1088 started at Mon Dec 18 10:21:54 2017. It will take at least 0:00:01.106325
 #Pt No    mot13      ct13      ct14      ct15      ct16       dt   
   0         0        0.1       0.2       0.3       0.4     0.012306
   1         1        0.1       0.2       0.3       0.4     0.341635
   2         2        0.1       0.2       0.3       0.4     0.49388 
   3         3        0.1       0.2       0.3       0.4     0.610897
   4         4        0.1       0.2       0.3       0.4     0.738575
   5         5        0.1       0.2       0.3       0.4     0.86619 
   6         6        0.1       0.2       0.3       0.4     0.984056
   7         7        0.1       0.2       0.3       0.4     1.10308 
   8         8        0.1       0.2       0.3       0.4     1.22073 
   9         9        0.1       0.2       0.3       0.4     1.33891 
   10        10       0.1       0.2       0.3       0.4     1.46657 
Operation saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1088 ended at Mon Dec 18 10:21:56 2017, taking 0:00:01.578044. Dead time 30.3% (motion dead time 4.5%)
10 loops, best of 3: 1.44 s per loop


timeit -n 10 %ascanct mot13 0 10 10 0.1
Operation will be saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1119 started at Mon Dec 18 10:24:18 2017. It will take at least 0:00:00
 #Pt No    mot13      ct13      ct14      ct15      ct16       dt   
Motor positions and relative timestamp (dt) columns contains theoretical values
   0         0        0.1       0.2       0.3       0.4       0.01  
   1         1        0.1       0.2       0.3       0.4       0.11  
   2         2        0.1       0.2       0.3       0.4       0.21  
   3         3        0.1       0.2       0.3       0.4       0.31  
   4         4        0.1       0.2       0.3       0.4       0.41  
   5         5        0.1       0.2       0.3       0.4       0.51  
   6         6        0.1       0.2       0.3       0.4       0.61  
   7         7        0.1       0.2       0.3       0.4       0.71  
   8         8        0.1       0.2       0.3       0.4       0.81  
   9         9        0.1       0.2       0.3       0.4       0.91  
   10        10       0.1       0.2       0.3       0.4       1.01  
Correcting overshoot...
Operation saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
Scan #1119 ended at Mon Dec 18 10:24:20 2017, taking 0:00:01.188341. Dead time 100.0% (motion dead time 100.0%)
Operation will be saved in /siciliarep/tmp/rhoms/scans/20171218.dat (Spec)
10 loops, best of 3: 1.28 s per loop
```

There is not a big different between them, but the value with the patch depends of the storage state.

rhomspuron:
I did more test (20 repetitions per scan) on the beamline with a Ni6602 card and a dummy motor.

With the PR:
```
%ascanct dmot1 0 10 100 0.1
mean time: 11.40690016746521
max time: 11.624068975448608
min time: 11.28506088256836

%ascan dmot1 0 10 100 0.1
mean time: 38.223894381523131
max time: 38.63029909133911
min time: 38.02476716041565

```

Withou the PR:

```
%ascanct dmot1 0 10 100 0.1
mean time: 11.389265537261963
max time: 11.625422954559326
min time: 11.323078870773315

%ascan dmot1 0 10 100 0.1
mean time: 38.116396427154541
max time: 38.462743043899536
min time: 37.86162805557251
```

reszelaz:
Thanks for the tests. I see that we are talking about 100 ms on a 100 points step scan. I would assume this overhead for the moment. If you disagree we would add an environment variable to make it optional.

@amilan I know that you wanted to give it a try. I merge this PR as it is but I think it will be still possible to comment on it.

PR #653, opened 2017-11-22
Eliminate ElementList Door attribute

cmft:
The Elements attributes are foreseen for exposing all elements
to the clients in form of encoded stream of data

e.g. Pool's elements, MacroServer elements.

Eliminate ElementList Door attribute since it is accessible from
the MacroServer.


reszelaz:
I agree on this PR. I'm aware that this way we are eliminating an attribute for the Tango device which in normal cases is considered as breaking the API. However in this case I'm almost sure that this attribute was never working correctly so no one could use this API. I do not plan to bump the major version due to that. Please shout if you think the opposite:)

PR #654, opened 2017-11-22
Workaround for bug #427 - default values of repeat parameters

reszelaz:
It is not well defined how to use default values of repeat parameters.
Now there exist two possibilities of definition: either on the repeat
parameter level or on the member level. For repeat parameters of only
one member defining the default value on the member level is not enough
cause the execution of the macro with the Macro.execMacro call fails.
The failure is related with the ParamDecoder.decodeRepeat implementation.
Duplicating the default value on the repeat parameter level workarounds
this. Apply this workaround until #427 is properly fixed.


reszelaz:
I've just added TODOs in the code with more information.

teresanunez:
I have reviewed this and it helps as workaround. I would merge it.

PR #657, opened 2017-11-24
Avoid exceptions in spock and widgets when MS is down

reszelaz:
Taurus3 used to return None for the getState query when the device server
was down. Taurus4 raises exceptions which are currently not handled.
In this case provide the same behavior as Taurus3 - assign None to the old
state.

This exceptions appear in spock or any macroserver related widget e.g. macrobutton.


cpascual:
Reproduced and confirmed that the PR fixes it. Merged

PR #659, opened 2017-11-27
Meshct macro

rhomspuron:
Adapt the CT Scan class to implement multiple waypoints and implement the meshct macro

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -85,6 +85,40 @@ def getCallable(repr):
         return lambda: None
 
 
+# TODO: remove starts
+def _calculate_positions(moveable_node, start, end):
+    '''Function to calculate starting and ending positions on the physical
+    motors level.
+    :param moveable_node: (BaseNode) node representing a moveable.
+                          Can be a BranchNode representing a PseudoMotor,
+                          or a LeafNode representing a PhysicalMotor).
+    :param start: (float) starting position of the moveable
+    :param end: (float) ending position of the moveable
+
+    :return: (list<(float,float)>) a list of tuples comprising starting
+             and ending positions. List order is important and preserved.'''
+    start_positions = []
+    end_positions = []
+    if isinstance(moveable_node, BranchNode):
+        pseudo_node = moveable_node
+        moveable = pseudo_node.data
+        moveable_nodes = moveable_node.children
+        starts = moveable.calcPhysical(start)
+        ends = moveable.calcPhysical(end)
+        for moveable_node, start, end in zip(moveable_nodes, starts,
+                                             ends):
+            _start_positions, _end_positions = _calculate_positions(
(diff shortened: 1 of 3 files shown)


reszelaz:
@rhomspuron thanks for this contribution, it looks promising! I have reviewed the code and made some cosmetic changes. The most relevant is about the use of the `waypoint_id` field in the step generated by the waypoint generator. Could you test if it still works as expected for you.

Apart of that I see the following two problems:

1. The waypoint generator also fills positions for the motor that is not intended to be moved. In the code you make some conditional skips for this motor e.g. to not participate in the acceleration and deceleration time selection or in the motion parameters configuration. But, at the end, this motor is being moved. Yes, it is true that in theory you move it to the same position as the starting position. But in practice, for motors with not precise positioning system, this may actually trigger a slight move of the motor. That, in this case would happen during the scan of another motor. 
2. You modify the `_index_offset` on the fly. I’m afraid that if the events gets queued for processing, and the next waypoint is already in progress, this may cause inconsistency of data i.e. wrong indexes being assigned and maybe some other consequences which I don’t know now.

While the first one, IMO, sounds like acceptable for the moment (knowing the consequences), I consider integration of this PR with the second one as problematic. Would you mind correcting it?

Finally, I took a profit of this PR and documented the current API of the waypoint generator. Which IMO requires some changes - marked as TODOs.

rhomspuron:
@reszelaz thanks for the review! 

I checked the code and the first problem is worst, the second motor will move to the final position (user position + acceleration  displacement) and it is not the user position. I will implement a solution for that.

The second problem should not pass because, the code is waiting for the finish on the measurement group. For that reason is not possible to start another acquisition without handling all data events. 

The problem could be to have not ordered list of events, thats mean a new event arrived before a old event, in this case the event receiver will take first the event of state previous to process all the data, but I do not know if it is possible. 

The best solution is to generate on the Pool the correct index, but for that we should change the MeasurementGroup API. I did the change to have that solution, but I will not integrate to this branch. I have it on: 
https://github.com/rhomspuron/sardana/tree/meshct_index_pool

reszelaz:
Great!

I have reviewed the changes for the motion path calculation and they are ok.

Regarding the second problem I completelly agree that the measurement group state change is automatically a synchronization mechanism. On the Pool side all the Data (ValueBuffer) events are emitted before the final State event gets emitted. Then on the client we have them buffered by the zmq and consumed sequentially. I would ask on the Tango forum if the events could get shuffled on the way to the client.

reszelaz:
Regarding the test failure it was due to the problems with travis-ci. It is workaround with #669. Could you please merge the develop branch to your PR. This should fix the tests.

reszelaz:
We have the first [confirmation](http://www.tango-controls.org/community/forum/c/general/development/is-events-order-maintained/) from the Tango experts about the events order. Then I'm merging it... Thanks!

PR #661, opened 2017-11-28
Missing  value of the total_scan_intervals. 

rhomspuron:
The total_scan_intervals key of the ScanDataEnvironment, does not have value for all aNscan types.


teresanunez:
Hi, I have reviewed this and IMO it can be integrated.

reszelaz:
Mergin it after @teresanunez approval.

PR #663, opened 2017-11-30
Check physical motor limits before moving PseudoMotors

teresanunez:
A very preliminary version for starting a discussion.
The PyTango AttributeProxy should not be used for getting the configuration, but
an specific Sardana class.

### DIFF
--- a/src/sardana/pool/poolpseudomotor.py
+++ b/src/sardana/pool/poolpseudomotor.py
@@ -45,7 +45,7 @@ from sardana.pool.poolbasegroup import PoolBaseGroup
 from sardana.pool.poolmotion import PoolMotion
 from sardana.pool.poolexception import PoolException
 
-from PyTango import AttributeProxy
+from sardana.sardanaattribute import SardanaAttributeConfiguration
 
 class Position(SardanaAttribute):
 
@@ -555,16 +555,8 @@ class PoolPseudoMotor(PoolBaseGroup, PoolElement):
                 raise PoolException("Cannot calculate motion: %s reports "
                                     "position to be None" % element.name)
             # TODO: get the configuration for an specific sardana class and
-            # get rid of AttributeProxy.
-            config = AttributeProxy(element.name + '/position').get_config()
-            try:
-                high = float(config.max_value)
-            except ValueError:
-                high = None
-            try:
-                low = float(config.min_value)
-            except ValueError:
-                low = None
+
+            low, high = SardanaAttributeConfiguration(element.name + '/position').range
             if high is not None:
                 if float(new_position) > high:
                     raise RuntimeError("%s: Requested movement above "
--- a/src/sardana/pool/poolpseudomotor.py
+++ b/src/sardana/pool/poolpseudomotor.py
@@ -557,23 +557,22 @@ class PoolPseudoMotor(PoolBaseGroup, PoolElement):
             # TODO: get the configuration for an specific sardana class and get rid of AttributeProxy.
             config = AttributeProxy(element.name + '/position').get_config()
             try:
-                high =  str(config.max_value)
-            except:
+                high = float(config.max_value)
+            except ValueError:
                 high = None
             try:
-                low =  config.min_value
-            except:
+                low = float(config.min_value)
+            except ValueError:
                 low = None
             if high is not None:
-                if float(new_position) > float(high):
+                if float(new_position) > high:
                     raise RuntimeError("%s: Requested movement above upper limit."
                                        % element.name)
             if low is not None:
-                if float(new_position) < float(low):
+                if float(new_position) < low:
                     raise RuntimeError("%s: Requested movement below lower limit."
                                        % element.name)
-                    
-            
+
             element.calculate_motion(new_position, items=items,
                                      calculated=calculated)
         return items
--- a/src/sardana/pool/poolpseudomotor.py
+++ b/src/sardana/pool/poolpseudomotor.py
@@ -554,7 +554,8 @@ class PoolPseudoMotor(PoolBaseGroup, PoolElement):
             if new_position is None:
                 raise PoolException("Cannot calculate motion: %s reports "
                                     "position to be None" % element.name)
-            # TODO: get the configuration for an specific sardana class and get rid of AttributeProxy.
+            # TODO: get the configuration for an specific sardana class and
+            # get rid of AttributeProxy.
             config = AttributeProxy(element.name + '/position').get_config()
             try:
                 high = float(config.max_value)
@@ -566,12 +567,12 @@ class PoolPseudoMotor(PoolBaseGroup, PoolElement):
                 low = None
             if high is not None:
                 if float(new_position) > high:
-                    raise RuntimeError("%s: Requested movement above upper limit."
-                                       % element.name)
+                    raise RuntimeError("%s: Requested movement above "
+                                       " upper limit." % element.name)
             if low is not None:
                 if float(new_position) < low:
-                    raise RuntimeError("%s: Requested movement below lower limit."
-                                       % element.name)
+                    raise RuntimeError("%s: Requested movement below "
+                                       " lower limit." % element.name)
 
             element.calculate_motion(new_position, items=items,
                                      calculated=calculated)
--- a/src/sardana/pool/poolpseudomotor.py
+++ b/src/sardana/pool/poolpseudomotor.py
@@ -45,6 +45,7 @@ from sardana.pool.poolbasegroup import PoolBaseGroup
 from sardana.pool.poolmotion import PoolMotion
 from sardana.pool.poolexception import PoolException
 
+from PyTango import AttributeProxy
 
 class Position(SardanaAttribute):
 
@@ -553,6 +554,26 @@ class PoolPseudoMotor(PoolBaseGroup, PoolElement):
             if new_position is None:
                 raise PoolException("Cannot calculate motion: %s reports "
                                     "position to be None" % element.name)
+            # TODO: get the configuration for an specific sardana class and get rid of AttributeProxy.
+            config = AttributeProxy(element.name + '/position').get_config()
+            try:
+                high =  str(config.max_value)
+            except:
+                high = None
+            try:
+                low =  config.min_value
+            except:
+                low = None
+            if high is not None:
+                if float(new_position) > float(high):
+                    raise RuntimeError("%s: Requested movement above upper limit."
+                                       % element.name)
+            if low is not None:
+                if float(new_position) < float(low):
+                    raise RuntimeError("%s: Requested movement below lower limit."
+                                       % element.name)
+                    
+            
             element.calculate_motion(new_position, items=items,
                                      calculated=calculated)
         return items
--- a/src/sardana/sardanaattribute.py
+++ b/src/sardana/sardanaattribute.py
@@ -358,12 +358,25 @@ class ScalarNumberAttribute(SardanaAttribute):
         SardanaAttribute.__init__(self, *args, **kwargs)
         self.filter = ScalarNumberFilter()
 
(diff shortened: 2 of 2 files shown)


daneos:
OK, looks great for me.
I modified the conversion while reading limits, it converts now to float inside the `try` block, since I've run into this problem when using motors with no limits configured:
```
An error occurred while running Macro 'umv([['tgap1', '1500.0']]) -> 874e4442-d5df-11e7-a40c-0242687759b6':
ValueError: could not convert string to float: Not specified
```
I have tested this using two physical motors with limits, a Slit using these motors with no limits on gap and offset, and another Slit using gap and offset from previous Slit as the physical roles.
I can see that moving any of these silts will check limits on physical motors.

teresanunez:
I also think that this could be integrated like this. The other solution would requires major changes.
@reszelaz, what do you think?.

reszelaz:
As @teresanunez says the proper implementation would require more effort.

Just brainstorming about it...
I think that we would need to get the attributes configuration in the Tango layer and create the `SardanaAttributeConfiguration` objects and set them to the Sardana attributes. The common implementation could be somewhere in the `PoolElementDevice` or `PoolDevice` class so it would work transparently for any attribute. What would be also necessary is to update the Sardana configuration every time someone changes it in Tango (callback?).

But coming back to this PR I don’t think that we would be able to get rid of the AttibuteProxy with what we have now. What we could try to do is to move the check to the Tango layer. In the Tango layer we have the callbacks executed on changes like e.g. position update or position write. These are “position” or “w_position” events. Maybe we could do the check there?

Let’s talk about it on the call...

teresanunez:
Hi @reszelaz, thanks for your comments. I have just now finished a new implementation, where I move the call to the ProxyAttribute to the SardanaAttributeConfiguration class. Like this could be easier the further implementation. I wanted to commit it now, but I will wait until the phone discussion.

reszelaz:
On the call we have agreed to:

- Check if in the current implementation, in case of a complex pseudo motors hierarchy e.g. physical motors -> pseudo motors level 1 -> pseudo motors level 2, the limits of pseudo motors level 1 are also verified.
- We will try if implementing this in the right way - verification in the core that would apply to any attributes, is relatively easy:
  - We will start assuming that the `SardanaAttributeConfiguration` is using the Tango AttributeProxy object.
  - In the first order we will check if it is possible to use the configuration from within the `SardanaAttribute.set_write_value` or `SardanaAttribute._set_write_value`. This would require to change the current implementation and place the limits verification only in the base class (`SardanaAttribute`) and call the super class method in all the sub-classes overriding it e.g. Position.
  - In the second order, If the previous point is possible, we could think of having a specific attribute configuration classes e.g. `TangoAttributeConfiguration` inheriting from `SardanaAttributeConfiguration`.

teresanunez:
> Check if in the current implementation, in case of a complex pseudo motors hierarchy e.g. physical motors -> pseudo motors level 1 -> pseudo motors level 2, the limits of pseudo motors level 1 are also verified.

I have checked this, and yes, all the motors of the different levels are checked.

teresanunez:
Hi, 
   I have checked this, here my comments:

> We will try if implementing this in the right way - verification in the core that would apply to any attributes, is relatively easy:
> 
>     We will start assuming that the SardanaAttributeConfiguration is using the Tango AttributeProxy object.

This is done and it works fine.

>     In the first order we will check if it is possible to use the configuration from within the SardanaAttribute.set_write_value or SardanaAttribute._set_write_value. This would require to change the current implementation and place the limits verification only in the base class (SardanaAttribute) and call the super class method in all the sub-classes overriding it e.g. Position.

This works but only if the pseudomotors use only real motors as physical motors. If they use pseudomotors does not work, because in that case only the function calculate_motion is used for
those motors, but neather set_write_position nor set_position are  called for those motors any
 more. I would then keep the implementation how it is: calling SardanaAttributeConfiguration from the
calculate_motion function for the attribute Position of each of the physical motors.

>     In the second order, If the previous point is possible, we could think of having a specific attribute configuration classes e.g. TangoAttributeConfiguration inheriting from SardanaAttributeConfiguration

This could be done in any case, because the first point can (is already implemented).

teresanunez:
I have also observed that when the PseudoMotors have another PseudoMotors as PhysicalMotors depending on the start positions the positions set to the PseudoMotors are not always reached. This is due to the fact that the real motors are moved only at the end of the hierarchy ... so the calculation of the PseudoMotors positions can be done with the original position and not with the one that should have after a first movement.

reszelaz:
@teresanunez I don't get this last comment:
> I have also observed that when the PseudoMotors have another PseudoMotors as PhysicalMotors depending on the start positions the positions set to the PseudoMotors are not always reached. This is due to the fact that the real motors are moved only at the end of the hierarchy ... so the calculation of the PseudoMotors positions can be done with the original position and not with the one that should have after a first movement.

Could you provide an example?

reszelaz:
I've checked it and as @teresanunez says, with what is currently implemented, an arbitrary hierarchy of the pseudomotors would be correctly checked. This is thanks to the recursive call to the `calculate_motion` method of the `PseudoMotor` and `Motor` classes.

I know that this is not the scope of this PR but we should also think on how to check limits in the opposite direction i.e. when one tries to move a physical motor to a place where the pseudo motor would be out of the limits. This may give us another insight on the whole problem. I think that the only way would be with the chain of events between the SardanaAttribute, but it is not so trivial…

I found a way on how to avoid the use of `AttributeProxy`, please follow this [thread](http://www.tango-controls.org/community/forum/c/general/development/how-to-retrieve-the-attribute-range-limits-within-the-server/) in the Tango forum. When it gets completely clarified we should think on how to prepare a self consistent `SardanaAttributeConfiguration` (or its specialization) object. This will also affect the #560.

I’m also in favor of extending the `SardanaAttribute` class with a capability to verify the limits. This way we could transfer all the comparison logic to this class. What do you think?

teresanunez:
> @teresanunez I don't get this last comment:
> 
>     I have also observed that when the PseudoMotors have another PseudoMotors as PhysicalMotors depending on the start positions the positions set to the PseudoMotors are not always reached. This is due to the fact that the real motors are moved only at the end of the hierarchy ... so the calculation of the PseudoMotors positions can be done with the original position and not with the one that should have after a first movement.
> 
> Could you provide an example?

You can reproduce it if you create two PseudoMotors where the Physical motors are also PseudoMotors, for example:

          gap05    gap_new      mot17      mot18      mot19      mot20
User    1.37500    1.87500    1.18750    0.18750    0.00000    0.00000
Dial        nan        nan    1.18750    0.18750    0.00000    0.00000
                                                                      
       offset05 offset_new
User    0.50000   -0.43750
Dial        nan        nan

mot17 and mot18 are the physical motors for gap05 and offset05 and these last both are the physical
motors for gap_new and offset_new.
Look what happens if I move gap_new to 3:

Door_issue102_1 [10]: mv gap_new 3
Door_issue102_1 [11]: wa
Current positions (user, dial) on 2018-02-02 12:53:32.272573

          gap05    gap_new      mot17      mot18      mot19      mot20
User    1.93750    2.43750    1.46875    0.46875    0.00000    0.00000
Dial        nan        nan    1.46875    0.46875    0.00000    0.00000
                                                                      
       offset05 offset_new
User    0.50000   -0.71875
Dial        nan        nan

gap_new is not going to 3 but to 2.43750

teresanunez:
Here with some numbers so that it is easy to see:

Current positions:
Door_issue102_1 [3]: wa
Current positions (user, dial) on 2018-02-02 13:25:25.617922

          gap05    gap_new      mot17      mot18      mot19      mot20
User    2.48242    2.98242    1.74121    0.74121    0.00000    0.00000
Dial        nan        nan    1.74121    0.74121    0.00000    0.00000
                                                                      
       offset05 offset_new
User    0.50000   -0.99121
Dial        nan        nan

Now we send:

Door_issue102_1 [4]: mv gap_new 2

These are the values in CalcPhysical (index, pseudomotor positions, calculated physical position)

Slit.CalcPhysical 1 [2.0, -0.9912109375] ->0.0087890625
Slit.CalcPhysical 2 [2.0, -0.9912109375] ->1.9912109375
Slit.CalcPhysical 1 [2.482421875, 0.0087890625] ->1.25
Slit.CalcPhysical 2 [2.482421875, 0.0087890625] ->1.232421875
Slit.CalcPhysical 1 [1.9912109375, 0.5] ->1.49560546875
Slit.CalcPhysical 2 [1.9912109375, 0.5] ->0.49560546875

And here the final positions:

          gap05    gap_new      mot17      mot18      mot19      mot20
User    1.99121    2.49121    1.49561    0.49561    0.00000    0.00000
Dial        nan        nan    1.49561    0.49561    0.00000    0.00000
                                                                      
       offset05 offset_new
User    0.50000   -0.74561
Dial        nan        nan

mrosanes:
Hi @teresanunez and @daneos. I have been reviewing PRs https://github.com/sardana-org/sardana/pull/663 and https://github.com/sardana-org/sardana/pull/560 from the point of view of performance referring to the number of accesses to the Tango DB.

With these pull requests, the number of accesses to the Tango Database is going from 1 DbImportDevice access per scan point, to 3 DbImportDevice accesses per scan point. This increase is due to the AttributeProxy creation in SardanaAttributeConfiguration.

--

Apparently, there is no problem; but in BL09-Mistral Beamline, we are facing a problem caused by the number of accesses per second to the TangoDB. It is related to a GUI (XMController) of the TXM Microscope detector designed by the external company Xradia-Zeiss, which interfaces with some Tango motors of the beamline and it is performing too much polling to the Device Attributes, by creating an attribute proxy at each of the accesses to the attributes. This has increased a lot the number of accesses to the TangoDB of Mistral, and has increased the CPU to around 70%, with a load of 2-3 (for a CPU at 2600MHz) for only 4 Tango motors interfaced from the GUI XMController. 

It is not the same case, but it is only to indicate that increasing the number of accesses to the TangoDB by creating attributes or device proxies can have some not expected consequences.

reszelaz:
Regarding the examples from https://github.com/sardana-org/sardana/pull/663#issuecomment-362567070 and https://github.com/sardana-org/sardana/pull/663#issuecomment-362573815 I'm not sure if they should work...

What I suspect is:
1. The move of `gap_new` calls its `calculate_motion` which retrieves the current `offset_new` position and performs the `calc_all_physical` of the Slit controller - this is fine.
2. The calculated positions are then used when iterating over the physical elements (in this case these are pseudos `gap05` and `offset05`) and their `calculate_motion` methods are called.
3. However these calculations are executed sequentially i.e. first the `gap05` is calculated using the current position of `offset05` and then the other way around the `offset05` is calculated using the current position of `gap05`.
4. But at the end we are calculating positions of the same physical motors :`mot17` and `mot18`. So the second calculation is overriding the results of the first one. Furthermore none of the results are correct, cause instead of the current position of the sibling, the new position of the sibling should be used - we are changing at the same time `gap05` and `offset05`.

But, I repeat, these are just my suspicions - I did not have time to debug it.

Anyway, I think that what @teresanunez observed is not related to the limits. So I propose to open a new issue for that where we could discuss it.

reszelaz:
As agreed on the last follow-up meeting we will include this fix in the Jan18 release. Note that the proposed implementation, which uses the `AttributeProxy`, is more like a workaround. We should try to get rid of it as soon as possible, especially taking into accout this https://github.com/sardana-org/sardana/pull/663#issuecomment-363451003.

I've created #704 cause the last 2 commits from @teresanunez, which try to use the `SardanaAttributeConfiguration`, try to add a new API which I'm not sure that should stay like this. So in order to not influence us when we will try to properly solve the problem I prefer to not include them. But many thanks for the intention to improve the current workaround!

I'll merge #704 tomorrow if travis-ci allows me:) and I'm closing now this PR. Many thanks to all involved in fixing this problem!

Issue #662, opened 2017-11-28
Is PoolUtil thread-safe?


jkotan:
Hi, 

I would like to ask if access to `PoolUtil` object (defined on the `poolutil` module level) via its `get_device(...)` is thread safe? 
Since `get_device(...)` is non-atomic operation on `CaselessDict()` wouldn't be better to protect it via mutex or lock?

I ask this question because we encountered problems with `TangoAttrCTController`.
In one of our pools we have almost a hundred  pool `TangoAttrCTController` devices where many of them are from other `TANGO_HOST`'s. Accessing them simultaneously we get randomly corba TIME_OUT errors, e.g.

    DevFailed[
    DevError[
	desc = TRANSIENT CORBA system exception: TRANSIENT_CallTimedout
      origin = DeviceProxy:read_attribute
      reason = API_CorbaException
    severity = ERR]

    DevError[
	desc = Timeout (3000 mS) exceeded on device expchan/tangoattributectctrl/33
      origin = DeviceProxy:read_attribute
      reason = API_DeviceTimedOut
    severity = ERR]
    ]

Best regards,
Jan

jkotan:
When I serialize my code, i.e. read attribute values of TangoAttrCTController one after another in one thread I do not get the above error. Moreover, It happens only for TangoAttrCTController devices. So it looks like it is thread-safety problem.

reszelaz:
@jkotan thanks for the report! I would like to ask you some questions about your setup in order to undestand it better. Maybe even I would be able to reproduce it on my environment.

From the error message I see that the timeout is on the experimental channel device proxy. When does it happen? In a step scan or a ct macro? 

How many of these experimental channels point to Tango attributes comming from the same device? Is reading of these attributes slow? Do you have an average time of readout.

> When I serialize my code, i.e. read attribute values of TangoAttrCTController one after another in one thread I do not get the above error.

You mean that you have prepared a script which reads the `Value` attributes of these experimental channels one by one?

jkotan:
Hi Zibi,

in out NXSRecSelector server I've a thread  worker
https://github.com/nexdatas/recselector/blob/master/nxsrecconfig/CheckerThread.py
which checks attributes of tango servers if they are readable.
Basically it goes through all import attributes to store and checks it they are readable by 

1) creating related `DeviceProxy` devices 
2) checking device states
3) ping devices
4) read the attribute values

I use to set about 20 workers to speed up the process of checking the attributes.
[...]

PR #665, opened 2017-11-30
add lock to PoolUtil

jkotan:
It fixes point 3) mentioned in #662


reszelaz:
Looks ok to me. Tested with [`TangoAttrCTCtrl`](https://sourceforge.net/u/zreszela/sardana-controllers/ci/sep6/tree/python/countertimer/TangoAttrCTCtrl.py).

Issue #667, opened 2017-12-11
Memorized attributes can't be restored if TANGO_HOST has changed


daneos:
Recently I switched my virtual machines to use PQDN instead of FQDN in TANGO_HOST and while starting Sardana I noticed this problem:
```
Tango NamedDevFailedList exception
   Exception for object Moveable
   Index of object in call (starting at 0) = 0
       Severity = ERROR 
       Error reason = PyDs_PythonError
       Desc : KeyError: "There is no element with full name 'v-deb.new.solaris.x.daneos.com:10000/motor/motctrl01/1'"

       Origin :   File "/usr/local/lib/python2.7/dist-packages/sardana-2.3.3a0-py2.7.egg/sardana/tango/pool/MeasurementGroup.py", line 235, in write_Moveable
    self.measurement_group.moveable = attr.get_write_value()
  File "/usr/local/lib/python2.7/dist-packages/sardana-2.3.3a0-py2.7.egg/sardana/pool/poolmeasurementgroup.py", line 624, in set_moveable
    self._moveable_obj = self.pool.get_element_by_full_name(moveable)
  File "/usr/local/lib/python2.7/dist-packages/sardana-2.3.3a0-py2.7.egg/sardana/sardanacontainer.py", line 176, in get_element_by_full_name
    "There is no element with full name '%s'" % full_name)

   Summary exception
       Severity = ERROR 
       Error reason = API_AttributeFailed
       Desc : Failed to execute DeviceClass::set_memorized_values() on device mntgrp/pool_test_1/dummy, object(s) Moveable
       Origin : DeviceClass::set_memorized_values()
```
The `v-deb.new.solaris.x.daneos.com:10000` is my old TANGO_HOST, new one is `v-deb:10000`.
It looks like memorized values are saved with element full name, together with TANGO_HOST.

All the machines and database are the same, only a way Tango references the database changed.

PR #668, opened 2017-12-13
[WIP] Adapt Sardana to use Taurus 4 compatible names (URI with FQDN)

reszelaz:
This PR aims to adapt Sardana to internally use the Taurus 4 compatible names (URI and FQDN). When designing the solution we will assume that:
- Client and Server sides use the same Sardana version
- Backwards compatibility will be maintained but only in one direction i.e. rollbacks will not be supported

The following parts of the code will require changes:

- [x] elements full names (Pool and most probably the MacroServer) containers
- [x] measurement group configuration attribute (controllers and elements names)
- [x] measurement group moveable and moveables attribute (moveables will be introduced after re-reverting #590)
- [x] Sardana model names in taurusgui XML configuration
- [x] all custom modifications for Taurus3 and Taurus4 compatibility
- [x] door and macroserver model names in spock profile files

Fixes issues: #667, #315, #500, ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ This file follows the formats and conventions from [keepachangelog.com]
   `Macro.execMacro` (#654)
 
 ### Changed
+- Internally use URI (FQDN) references to be fully compatible with Taurus4
 - Rename edctrl to edctrlcls macro (#541)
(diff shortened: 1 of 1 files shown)


reszelaz:
I think everything is done from the checklist. I'm closing this issue then.

PR #669, opened 2017-12-14
Use old Ubuntu image for travis-ci to fix build errors

reszelaz:
Travis builds started to fail from the Ubuntu Trusty roll-out
from 12.12.2017. Use the old image. More info in:
https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch

Another PR will be open to investigate where exactly is the problem.
From the logs we can see that neither Pool nor MacroServer can be started on docker:
```
$ if [ $TEST == "testsuite" ]; then docker exec sardana-test supervisorctl start Pool; fi
Pool: ERROR (abnormal termination)
$ if [ $TEST == "testsuite" ]; then docker exec sardana-test supervisorctl start MacroServer; fi
MacroServer: ERROR (abnormal termination)
```

PR #670, opened 2017-12-14
[WIP] Investigate Pool and MS crash by printing logs

reszelaz:
Please ignore this PR. It is just to debug travis problems described in #669

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,13 +31,15 @@ install:
   - if [ $TEST == "testsuite" ]; then docker exec sardana-test bash -c "cd /sardana && python setup.py install"; fi
 
   # start Pool and MacroServer necessary for macro tests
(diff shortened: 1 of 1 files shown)


reszelaz:
The first conclusions are that there are problems with mysql:

```
0.05s$ if [ $TEST == "testsuite" ]; then docker exec sardana-test bash -c "ls /var/run/mysqld/mysqld.sock"; fi
ls: cannot access /var/run/mysqld/mysqld.sock: No such file or directory
```

More details in `/var/log/mysql/error.log` of this [build](https://travis-ci.org/sardana-org/sardana/jobs/319292127#L1860)

reszelaz:
We won't use travis-ci anymore - see #1433. There is no more interest in this investigation.

PR #671, opened 2017-12-15
Synchronization Parameters

rhomspuron:
The IcePAP and the Pmac controllers allow to reuse the same output pin to generate trigger according any register of their motors. For that reason a general trigger controller for these hardware need to know which motor is using on the acquisition to configure the proper register to generate the triggers. 

On the other hand, IMHO the synchronization object should have the information for the Position Domain for all motors involved on the acquisition, the current implementation generates the information using the master motor (the first one). For a hardware trigger generator which uses Position Domain, this implementation can produce wrong behaviors (and results) if the motor which is configured is not the first one, and to debug the problem could be difficult.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -2114,13 +2114,23 @@ class CTScan(CScan, CAcquisition):
             initial_position = start
             total_time = abs(total_position) / path.max_vel
             delay_time = path.max_vel_time
-            synch = [{SynchParam.Delay: {SynchDomain.Time: delay_time},
-                      SynchParam.Initial: {SynchDomain.Position: initial_position},
-                      SynchParam.Active: {SynchDomain.Position: active_position,
-                                          SynchDomain.Time: active_time},
-                      SynchParam.Total: {SynchDomain.Position: total_position,
-                                         SynchDomain.Time: total_time},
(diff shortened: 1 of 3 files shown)


reszelaz:
@13bscsaamjad today I was mentioning you this PR where @rhomspuron already thought about adding a moveable to the synchronization description.

13bscsaamjad:
@reszelaz this looks good to me.

rhomspuron:
Hi all, 

Could you integrate this PR ASAP? 

Thanks in advance

reszelaz:
Many thanks, @rhomspuron for proposing this enhancement and @13bscsaamjad for the review! And sorry for waiting so long for my comments.
I think it would be a very nice feature to allow the multiplexor mode of operation for the trigger/gates. While the proposed solution, to reach the `TriggerGateController` plugin with the moveable full name, works for motion controllers supporting the multiplexor mode (see [IcePAPTriggerGateController](https://github.com/ALBA-Synchrotron/sardana-icepap/blob/master/sardana_icepap/ctrl/IcePAPTriggerController.py) example), I have some doubts it will work for non-motion controllers e.g. Pandabox, PiLC, etc. I have listed some ideas on how to cover both types of hardware in #1678. I hope after the holidays period we could meet and discuss which direction to follow, and soon we could include this enhancement in Sardana.

PR #675, opened 2018-01-04
Make MacroStatus read return the last macro status

reszelaz:
MacroStatus attribute of the Door returns an empty value. Make it return
a valid value - the last macro status update. If no macro was run yet it
returns `null`.

### DIFF
--- a/src/sardana/tango/macroserver/Door.py
+++ b/src/sardana/tango/macroserver/Door.py
@@ -350,7 +350,10 @@ class Door(SardanaDevice):
         self.__buf_data = data
 
     def read_MacroStatus(self, attr):
-        attr.set_value('', '')
+        macro_status = self.door.macro_status
+        codec = CodecFactory().getCodec('json')
+        status = codec.encode(('', macro_status))
+        attr.set_value(*status)
 
     def Abort(self):
         self.debug("Abort is deprecated. Use StopMacro instead")


teresanunez:
Hi @reszelaz, I have tested this. If a macro has been run it works properly, but if not one get an the error message:

    desc = TypeError: object of type 'NoneType' has no len()

when trying to read the attribute. I think one should try to avoid this error.

reszelaz:
Thanks for testing! It also failed on travis-ci. I will look into that.

reszelaz:
Could you give me a hint on how did you reproduce it? I try this after the MacroServer startup:
```
In [1]: import tango

In [2]: tango.DeviceProxy("Door_zreszela_1").macrostatus
Out[2]: ('json', 'null')
```
And there is no error.

teresanunez:
I get it directly with jive reading the attribute and with a python script.
But I get the same error with the example you give:

>>> tango.DeviceProxy("door/read_macrostatus/1").macrostatus
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/tango/device_proxy.py", line 264, in __DeviceProxy__getattr
    return self.read_attribute(name).value
  File "/usr/lib/python2.7/dist-packages/tango/green.py", line 264, in greener
    ret = submitter(fn, self, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/tango/green.py", line 100, in submit
    return fn(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/tango/device_proxy.py", line 352, in __DeviceProxy__read_attribute
    return __check_read_attribute(self._read_attribute(value, extract_as))
  File "/usr/lib/python2.7/dist-packages/tango/device_proxy.py", line 128, in __check_read_attribute
    raise DevFailed(*dev_attr.get_err_stack())
PyTango.DevFailed: DevFailed[
DevError[
    desc = TypeError: object of type 'NoneType' has no len()
           
  origin =   File "/home/tnunez/sardana_read-macrostatus/src/sardana/tango/macroserver/Door.py", line 355, in read_MacroStatus
    status = codec.encode(('', macro_status))
  File "/usr/lib/python2.7/dist-packages/taurus/core/util/codecs.py", line 320, in encode
    if len(data[1]) == 3:

  reason = PyDs_PythonError
severity = ERR]

DevError[
    desc = Failed to read_attribute on device door/read_macrostatus/1, attribute macrostatus
  origin = DeviceProxy::read_attribute()
  reason = API_AttributeFailed
severity = ERR]
]


Could be related to the taurus version?

reszelaz:
I don't think so I've just checked it with support-3.x. But you gave me a hint. It must be related to the Tango version. I suspect that you use Tango8, isn't it? Also on travis-ci we still use Tango8...

teresanunez:
Yes, we have tango8

reszelaz:
Hmm, more strange. I have just used the docker image that we used in travis-ci and there are no problems with reading the macrostatus attribute before executing any macro, it returns null as in my development environment (Tango 9). On this docker image we use PyTango 8.1.5. So there must be two different problems: the one present in travis-ci and the one that you found. Which exact version of Tango do you use?

Comming back to the taurus version, do you use any special version? I cannot find in the support-3.x niehter in 3.7.4 this line that is present in the traceback that you have posted:
```
File "/usr/lib/python2.7/dist-packages/taurus/core/util/codecs.py", line 320, in encode
if len(data[1]) == 3:
```

Or maybe you have any other clue on what may be different between your environment and the docker [image](https://hub.docker.com/r/reszelaz/sardana-test/~/dockerfile/).

teresanunez:
Hi @reszelaz, I have just seen that this line was introduced by us (me or Jan, I don't remember now who or why is this code there) in the DESY version. I will figure it out and then I will contact you again. Many thanks.

teresanunez:
HI @reszelaz, I introduced that change, fast two years ago, for reading the 1d in Sardana, without
this it was not working (I will send you my comments about that), but I think this could be obsolete
now. I will check it.

teresanunez:
Removing my changes the PR works as expected.

reszelaz:
Great! Thanks for the investigation. Then there are only the problems revealed by the unittests on travis. These may be reladed to the testing infrastructure - macroexecutor that uses the macrostatus attribute events. I will check it on Wednesday.

reszelaz:
I think that integrating this PR just before the Jan18 release is a little bit dangerous. The introduced change affected the testing infrastructure, which at the end is as any other MacroServer client. I suspect that there are users which developed their solution relying on the MacroStatus events.

I propose to delay its integration until after the release, and live with this change for half a year in the develop branch. After testing it more, we could take a decission on releasing it in Jul18, or if revealed that it breaks the backwards compatibility, either delay it to the next major version release or just drop it - at the end it does not add any very important feature.

PR #676, opened 2018-01-05
Improvements when server startup process is aborted

reszelaz:
Add two tiny improvements:
- Exit smoothly when user aborted MacroServer creation, using the Ctrl+c on the Pool selection
- Abort server startup process when it is known that its instance won't exist in the DB so it won't be able to
startup


teresanunez:
I have tested this and it works as described.

reszelaz:
Merging after @teresanunez's review.

PR #677, opened 2018-01-08
Disabled channels and offline controllers

rhomspuron:
The measurement group (MG) does not work, if the controller of the disabled channel  is offline. IMHO this is a bug, because if the channels are disabled the state of the controller is not matter, and it should not affect the working of the other channels. 

We found this problem because we use one MG with all channels of the beamline (4 detectors). The user changes the experiment type by enabling/disabling the channels. If the experiment type is not require one detector, it could be off, in this case, the controller of this detector will be offline, because it can not connect to the hardware.

### DIFF
--- a/src/sardana/pool/poolacquisition.py
+++ b/src/sardana/pool/poolacquisition.py
@@ -560,6 +560,11 @@ class PoolAcquisitionBase(PoolAction):
                 ctrl_enabled = True
             # check if the ctrl has enabled channels
             if ctrl_enabled:
+                # enabled controller can no be offline
+                if not ctrl.is_online():
+                    self.main_element.set_state(State.Fault, propagate=2)
+                    msg = "controller {0} is offline".format(ctrl.name)
+                    raise RuntimeError(msg)
                 pool_ctrls.append(ctrl)
                 # only CT will be read in the loop, 1D and 2D not
                 if ElementType.CTExpChannel in ctrl.get_ctrl_types():
--- a/src/sardana/pool/poolacquisition.py
+++ b/src/sardana/pool/poolacquisition.py
@@ -39,7 +39,6 @@ from taurus.core.util.enumeration import Enumeration
 
 from sardana import SardanaValue, State, ElementType, TYPE_TIMERABLE_ELEMENTS
 from sardana.sardanathreadpool import get_thread_pool
-from sardana.sardanautils import is_non_str_seq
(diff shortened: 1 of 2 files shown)


teresanunez:
Hi, I have checked this and the configuration is not loaded if the controller is offline. So there are
no problems if channels are disable and its controller is off.

rhomspuron:
Hi @teresanunez, thanks to check it. 

I have a doubt, did you check it with or without this PR?

teresanunez:
I checked it with this PR.
What I was doubting if it should be told or not ... You skip the loading of not is_online controllers,
 but it doesn't mean that you disable them for the acquisition in case they are not disable in the MG,
 but you already wrote in the PR 'Disabled channels'

rhomspuron:
You are right, maybe the Title of the PR is not correct. The idea is to do not abort the acquisition if the channel is disabled. If the channel is not disabled and the controller is offline, I think that the acquisition should abort because the user wants to use this channel.

teresanunez:
Yes, I also think that the PR should be more general and take care in general of offline controllers and
 check if the channel is disable or not for proceeding.

rhomspuron:
In this case, we should verify if all channel of the controller are disabled before to load the configuration, but this verification will do it after to load it. Do you propose to move the verification before to load the configuration?

teresanunez:
Well, in fact I don't think that is needed. It could stay also like it is. If the channel is enable but the controller is off one gets in any case the error at some point, like it happens know. So the decision would be where one want to make the change. But for performance it would work in any case.

rhomspuron:
I checked the verification code (enable/disable channel), to move it before to load the configuration, it means to change the acquisition action, it could be a big change. The loading of the configuration is doing on the MG start method, and the verification is doing on the acquisition action.

teresanunez:
Hi @rhomspuron, thanks for checking that. For me it is ok like it is, let's see what the others say.

reszelaz:
Thanks @rhomspuron and @teresanunez! This PR will certainly make the system more robust. I agree that if the controller is offline (there were some problems in the initialization) and it is disabled in the measuremet group it should not affect the acquistion.

I have just modified the following things:
* added an extra check whether the controller is offline before the load sequence is actually executed. 
* remove the extra information in case of the loead sequence error - the newly added check takes care about this case now.

Nevertheless, there are some other places where the offline controller may cause uncontrolled exceptions. We should treat them at some point as well.

PR #681, opened 2018-01-17
Avoid errors when acquiring with offline controller

reszelaz:
An experimental channels from an offline controller does not export its
attributes. Then, it is not possible to obtain the Value attribute
configuration and an empty AttributeInfoEx is created for it. Its default
data_type is set to DevVoid. Handle this case and consider the
TangoChannelInfo with an unknown data_type. This avoids errors.

A change was proposed to Taurus that would avoid this PR: https://github.com/taurus-org/taurus/pull/666.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/pool.py
+++ b/src/sardana/taurus/core/tango/sardana/pool.py
@@ -1140,9 +1140,9 @@ class TangoChannelInfo(BaseChannelInfo):
                 # For backwards compatibility:
                 # starting from Taurus 4.3.0 DevVoid was added to the dict
                 if data_type == PyTango.DevVoid:
-                     self.data_type = None
+                    self.data_type = None
                 else:
-                     raise e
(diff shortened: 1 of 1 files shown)


reszelaz:
Thanks! I think it would be beneficial for the Taurus3 uesrs to keep it. I will just modify it to reflect that it only applies to Taurus3.

reszelaz:
I've changed the code so it assumes to use Taurus 4.3 implementation and fallback for previous versions of Taurus. @mrosanes is going to test it.

mrosanes:
If a controller cannot initialize correctly, the error message appearing when performing an acquisition with a given measurement group is more explicit right now using Taurus3 and also Taurus4:
<pre>
Cannot execute 'get_ctrl_par' because 'ctctrl...' is offline
</pre>

reszelaz:
Great! Than I'm merging this PR.

PR #682, opened 2018-01-17
Provide Record data to ct macro

droldan:
Provide Record object as a data.


droldan:
How to Test it:
class getData(Macro):
    def run(self):
        ct, _ = self.createMacro('ct 2')
        self.runMacro(ct)
        data = ct.getData()
        for idx, rc in data.items():
            self.info("Record Data for record No: %r" %(idx))
            for counter, value in rc.data.items():
                self.info("Counter %r: %r" % (counter, value))

teresanunez:
Hi, I have tested and it works fine. IMO it can be merged.

reszelaz:
Thanks @droldan and @teresanunez!
I think we can simplify the data and eliminate the dictonary wrapping dictionary and just leave the Record object as data. I will proceed with the change.

PR #683, opened 2018-01-17
Fix getData with aNscan macros

droldan:
The 'getData' method of a aNscan Macro does not return the data, raising the exception "Macro XXX does not produce any data" because the Macro class is returning the value of self._data.

In a aNscan macro the data is available in data @property -> self.data, this PR fix the problem on execute the 'getData' method returning the data.


reszelaz:
Merging. Thanks!

PR #684, opened 2018-01-17
Provide Record data to uct macro

droldan:
Provide Data for 'uct' macro


droldan:
How to test it:

class getData(Macro):
    def run(self):
        ch_name = 'ct01'
        ct, _ = self.createMacro('uct 5')
        self.runMacro(ct)
        data = ct.getData()
        for idx, rc in data.items():
            self.info("Record Data for record No: %r" %(idx))
            for counter, value in rc.data.items():
                self.info("Counter %r: %r" % (counter, value))

reszelaz:
Merging it.

PR #685, opened 2018-01-18
Fix macro hooks setter (broken in #655)

reszelaz:
Pull request #655 breaks the hooks setter - if the hooks were never
set the _hookHintsDict was never initialized so it can not be deleted.
Protect the deletion by checking if the member exists.

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -238,7 +238,8 @@ class Hookable(Logger):
                         'Deprecation warning: hooks should be set with a list of hints. See Hookable API docs')
 
             # delete _hookHintsDict to force its recreation on the next access
-            del self._hookHintsDict
+            if hasattr(self, '_hookHintsDict'):
+                del self._hookHintsDict
             # create _hookHintsDict
             self._getHookHintsDict()['_ALL_'] = zip(*self._hooks)[0]
             nohints = self._hookHintsDict['_NOHINTS_']


reszelaz:
This problem can not be easily reproduced. For example the macro `sardana.macroserver.macros.examples.hooks.hooked_scan` is not affected. But if one attach a hooks to the `ascanct`, like below, it is affected.

~~~
class hooked_ascanct(Macro):

        param_def = [['motor', Type.Moveable, None, 'Moveable name'],
                 ['start_pos', Type.Float, None, 'Scan start position'],
                 ['final_pos', Type.Float, None, 'Scan final position'],
                 ['nr_interv', Type.Integer, None, 'Number of scan intervals'],
                 ['integ_time', Type.Float, None, 'Integration time'],
                 ['latency_time', Type.Float, 0, 'Latency time']]


        def pre_configure_hook(self):
            self.output("In pre_configuration_hook")

        def post_configure_hook(self):
            self.output("In post_configuration_hook")

        def pre_start_hook(self):
            self.output("In pre_start_hook")

        def post_move_hook(self):
            self.output("In post_move_hook")

        def cleanup(self):
            self.output("In cleanup hook")


        def run(self, motor, start_pos, end_pos, nr_of_trigger, int_time,
                latency_time):
            scan_macro, pars = self.createMacro("ascanct",
                                                      motor,
                                                      start_pos, end_pos,
                                                      nr_of_trigger,
                                                      int_time)

            scan_macro.hooks = [(self.pre_configure_hook,
                                 ["pre-configuration"]),
                                (self.post_configure_hook,
                                 ["post-configuration"]),
                                (self.pre_start_hook,
                                 ["pre-start"]),
                                (self.post_move_hook,
                                 ["post-move"]),
                                (self.cleanup,
                                 ["post-cleanup"]),]
~~~

This is because in the `ascanct` the time estimation is disabled for the moment. And in case of the step scans the time estimation is the one which fills this the `_hookHintsDict` during the iteration over the points.

reszelaz:
@sardana-org/integrators could you take a look on this.

teresanunez:
i can have a look

teresanunez:
I have checked and it solves the described problem. IMO it can be merged.

PR #686, opened 2018-01-22
Move index offset setting

rhomspuron:
Move the setting of _index_offset attribute before the post-configuration hook, to allow its changing if the nr_point changes between waypoints.


reszelaz:
Thanks! Merging it.

PR #687, opened 2018-01-22
Fix getData error.

rhomspuron:
Applied the solution of the PR #683 to the other scan macros

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -723,6 +723,14 @@ class mesh(Macro, Hookable):
 
         self._gScan = SScan(self, generator, moveables, env, constrains)
 
+        # _data is the default member where the Macro class stores the data.
+        # Assign the date produced by GScan (or its subclasses) to it so all
+        # the Macro infrastrucutre related to the data works e.g. getter,
+        # property, etc. Ideally this should be done by the data setter
+        # but this is available in the Macro class and we inherit from it
+        # latter. More details in sardana-org/sardana#683.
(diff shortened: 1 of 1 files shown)


reszelaz:
In the comment we already say that:
"Ideally this should be done by the data setter but this is available in the Macro class and we inherit from it latter."
Some of the scans from this PR already inherits from the Macro class. So I think it is better here to use the setter instead of assigning directly to `_data`. Would you mind changing it?

src/sardana/macroserver/macros/scan.py
```diff
@@ -723,6 +723,14 @@ def prepare(self, m1, m1_start_pos, m1_final_pos, m1_nr_interv,
 
         self._gScan = SScan(self, generator, moveables, env, constrains)
 
+        # _data is the default member where the Macro class stores the data.
+        # Assign the date produced by GScan (or its subclasses) to it so all
+        # the Macro infrastrucutre related to the data works e.g. getter,
+        # property, etc. Ideally this should be done by the data setter
+        # but this is available in the Macro class and we inherit from it
+        # latter. More details in sardana-org/sardana#683.
+        self._data = self._gScan.data
```
reszelaz: For example, this is a candidate to use the setter.

reszelaz:
Thanks for changing the code! Now please change the comments accordingly, basically remove:
"Ideally this should be done by the data setter but this is available in the Macro class and we inherit from it latter. More details in sardana-org/sardana#683." in order to be coherent with the code.

reszelaz:
Great! Thanks. Merging it...

PR #689, opened 2018-01-24
Door names follow conventions of names of other elements

reszelaz:
Follow convention of other elements in Sardana with respect to `name` and `full_name`

### DIFF
--- a/src/sardana/tango/macroserver/Door.py
+++ b/src/sardana/tango/macroserver/Door.py
@@ -214,7 +214,7 @@ class Door(SardanaDevice):
         door = self.door
         if door is None:
             full_name = self.get_name()
-            name = full_name
+            name = self.alias or full_name
             macro_server = self.macro_server_device.macro_server
             self.door = door = \
                 macro_server.create_element(type="Door", name=name,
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi, I have checked this and it works fine. IMO it can be merged, after solving the travis errors.

reszelaz:
Ups! The automatic tests indicates that this change breaks something. For example the lists macros (`lsa`, `lsm`) does not show output in spock. I will check it.

reszelaz:
**What is broken:** macro messages using the `output`, `error`, `critical`, etc. API are not delivered to the clients e.g. spock.

**Why it is broken:** 
* The Door's attributes: Output, Error, Warning, etc. are implemented using the `AttributeLogHandler`. This is a special handler which has the ability to push Tango change events. These handlers are attached to the Door (Tango device) which is a python logger. The name of this logger is the Tango normal name, as it arrives in the `Device_4Impl.__init__` e.g. a/b/c (without the database).
* A Macro instance itself is a logger, which parent is the `MSDoor` which as well is a logger. The Macro instance does not have any handlers attached to it, so its log events are passed to the parent logger.
* In python, loggers are singletons, based on the logger name.
* So what is now broken with this PR, used to work before cause the MSDoor was using its name as the Tango normal name e.g. a/b/c exactly the same as the Tango device Door. So it was the same logger and the chain of loggers was closed. Since we now changed the MSDoor’s name, the chain of loggers is broken.

**What's next:** I will propose soon some solutions for that, cause I think it is a technical debt and not a feature.

PR #690, opened 2018-01-25
Add expconf macros module

droldan:
The idea is to provide a standard macros to manage the measurement group, like enable/disable channels, show the current status or select a measurement group.
NOTE: Of course that the macros name can be redefined or changed.

Maybe we could thing to define more macros to change the scan directory, scan file or change the snapshot group.


teresanunez:
Hi @droldan, I have reviewed this. Some comments:

- the description of the macros meas_enable_ch y meas_enable_ch_only is the same. I think they
  should be different and remark which is the difference between both macros. The same applies for
 meas_disable_ch and meas_disable_ch_only

- I am not sure, but perhaps one could avoid to give the MG as argument and take the selected one.
  The best would be to have it as optional ... but there we have to take care about optional arguments  
  that are not the last one ... it could complicate the setting of the arguments.

- Testing this I have realized that the 'refresh' button of the expconf window does not work. The changes done in the MG with the macros are only seen if I reopen expconf, clicking at the refresh
button makes nothing ... perhaps it depends on my sardana/taurus versions. Have you checked this?

reszelaz:
Regarding the previous question: what is the difference between `meas_enable_ch` and `meas_enable_ch_only`? The first one leaves the rest of the channels without changes. The second one disables all the rest of the channels. But talking to @droldan he said that the `meas_enable_ch_only` is no more necessary.

Also the question of the optional measurement group parameters is very interesting. The optional parameters, as proposed in #285, I think would only solve the problem if we place it at the very end of the list of parameters? This influenced my naming proposals.

Now regarding the naming. I was thinking a little bit about different possibilities for for these macros. Below I propose some options. Please express your preferences or propose new.

**Option 1**

* Use different macros for changing configuration of the active measurement group and any of the measurement groups.
* Use configuration parameter as a selector of action, something similar to git <action> e.g. git fetch, git commit.

```
active_meas_conf_set <conf param> <value> [<channel>]
active_meas_conf_get <conf param> [<channel>]
meas_conf_set <meas grp> <conf param> <param value> [<channel>]
meas_conf_get <meas grp> <conf param> [<channel>]
```

Where:
* `<meas grp>` is of type MeasurementGroup
* `<conf param>` is a of type String e.g. enabled, output, plot_type, plot_axes, timer, monitor, synchronizer, synchronization (I would leave shape, data_type, conditioning, normalization, nexus_path for future)
* `<param value>` is of type String
* `<channel>` is a repeat parameter of type String (we need to admit experimental channels and external attributes). For `*_get` macros, it would have min=0, meaning that if no channels are passed it will list configuration for all of them.

**Option 2**

* Use different macros for changing configuration of the active measurement group and any of the measurement groups.
* Use different macros for changing of different configuration parameters

```
active_meas_conf_set_enabled <param value> [<channel>]
active_meas_conf_get_enabled [<channel>]
meas_conf_set_enabled <meas grp> <param value> [<channel>]
meas_conf_get_enabled <meas grp> [<channel>]
```

Where:
* `<meas grp>` is of type MeasurementGroup
* `<param value>` is of type Boolean
* `<channel>` is a repeat parameter of type String (we need to admit experimental channels and external attributes). For `*_get_*` macros, it would have min=0, meaning that if no channels are passed it will list configuration for all of them.


```
active_meas_conf_set_plot_type <param value> [<channel>]
active_meas_conf_get_plot_type [<channel>]
meas_conf_set_plot_type <meas grp> <param value> [<channel>]
meas_conf_get_plot_type <meas grp> <param value> [<channel>]
```

Where:
* `<meas grp>` is of type MeasurementGroup
* `<param value>` is of type String e.g. Spectrum, Image, No
* `<channel>` is a repeat parameter of type String (we need to admit experimental channels and external attributes). For `*_get_*` macros, it would have min=0, meaning that if no channels are passed it will list configuration for all of them.

And similar for other configuration parameters e.g. enabled, output, plot_type, plot_axes, timer, monitor, synchronizer, synchronization (I would leave shape, data_type, conditioning, normalization, nexus_path for future)

dschick:
these are very nice features I was also looking forward to - thanks @droldan for starting that!

One obvious action I am thinking of with these macros is to change the plotted counters.
Here I have the following use case:
I am running some kind of calibration scan from time to time, which requires to plot a different counter than most of the other time. The calibration scan is already included in a dedicated macro.
When I start the calibration, I would like to aks for the currently selected counters for plotting - change the plotted counter to the important one - do the calbration scan - change the plotted counters back to the original one from before the calibration.

I guess this would require some macros like
active_meas_conf_get allPlottedChannels

I guess in general one could also create a number of different measurement groups with the appropriate configurations and then just switch between them with out programatically modifiying one of them.
Is there a design rule for when to switch or modify measurement groups?

Best

Daniel

dschick:
@droldan btw. I already started a newfile macro to modify scanID and ScanFile, see #777

droldan:
Hi @reszelaz, IMO the option 1 could be more powerful and it will not necessary to create a new macro for each new 'config param'.
On the other hand, I wouldn't differentiate between 'active_meas_conf_X' and 'meas_conf_X', if the optional parameters worked the default measurement group could be the active one, else use a specific measurement group.

droldan:
Hi @dschick, to change the plotted counter the measurement group PoolElement should provide a method to change it managing the measurement group configuration, IMO should be easy like 'enableChannels' or 'disableChannels' methods.

Best 
Daniel

reszelaz:
> When I start the calibration, I would like to aks for the currently selected counters for plotting - change the plotted counter to the important one - do the calbration scan - change the plotted counters back to the original one from before the calibration.

> I guess this would require some macros like
active_meas_conf_get allPlottedChannels

Following the *Option 1* proposal this execution should do the job for you:
`active_meas_conf_get plot_type`.
This is because no repetitions of `[<channel>]` means all channels. 

But since you are talking about programatic configuration (backup/restore) I think that macro is not what you are looking for. As I see it, the macro should just send the current configuration via the output stream. As said already by @droldan, for programatic needs this should be implemented on the [MeasurementGroup](https://github.com/sardana-org/sardana/blob/da9e9ccaee69e31ed16338ca49ed53e9acfd65b6/src/sardana/taurus/core/tango/sardana/pool.py#L1511) or [MGConfiguration](https://github.com/sardana-org/sardana/blob/da9e9ccaee69e31ed16338ca49ed53e9acfd65b6/src/sardana/taurus/core/tango/sardana/pool.py#L1213) Taurus extensions classes level. Actually all these macros should use the Taurus extension classes internally and implement just the minimal logic and formatting of the output.

> I guess in general one could also create a number of different measurement groups with the appropriate configurations and then just switch between them with out programatically modifiying one of them.
Is there a design rule for when to switch or modify measurement groups?

In principle both solutions should work well.

droldan:
I propose the methods needed at MeasurementGroup PoolElement to allow manage the MeasurementGroup configuration.
NOTE: The Enable and Timer parameters already has methods at MeasurementGroup PoolElement

```python
def getEnableChannels(self):
    """get acquisition Enabled channels."""
    pass

def setEnableChannels(self, channels):
    """Enable acquisition of the indicated channels.

    :param channels: (seq<str>) a sequence of strings indicating
                     channel names
    """
    pass

def getOutput(self, channel=None):
    """get the output State of the channels.

    :param channel: (<str>) a string indicating the channel name, in case of
    None, it will return all the Outputs Info
    """
    pass

def setOutput(self, channels, state):
    """Set the Output state of the indicated channels.

    :param channels: (seq<str>) a sequence of strings indicating
                     channel names
    :param state: <bool> Indicate the state of the output.
    """
    pass

def getTimer(self):
    """get the acquisition Timer.
    """
    cfg = self.getConfiguration()
    return cfg.channels[cfg.timer]

def setTimer(self, timer_name):
    """Set the acquisition Timer.

    :param timer_name: <str> strings indicating the timer name
    """

    try:
        self.getChannel(timer_name)
    except KeyError:
        raise Exception("%s does not contain a channel named '%s'"
                        % (str(self), timer_name))
    cfg = self.getConfiguration().raw_data
    cfg['timer'] = timer_name
    import json
    self.write_attribute("configuration", json.dumps(cfg))

def getMonitor(self, channel=None):
    """get the Monitor for the channel indicated. In case of empty channel
    value it will return  all the Monitor Info

    :param channel: <str> Indicate the chanel to return the Monitor Info
    """
    cfg = self.getConfiguration()
    return cfg.channels[cfg.monitor]

def setMonitor(self, channels, monitor):
    """Set the Monitor for the indicated channels.

    :param channels: (seq<str>) a list of strings indicating the channels
    to apply the monitor
    :param monitor: <str> string indicating the monitor name
    """
    pass

def getSynchronizer(self, channel=None):
    """get the synchronizer for the channel indicated. In case of empty channel
    value it will return  all the Synchronizers Info

    >:param channel: <str> Indicate the channel to return the Synchronizer Info
    """
    pass

def setSynchronizer(self, channels, synchronizer):
    """Set the synchronizer for the indicated channels.

    :param channels: (seq<str>) a list of strings indicating the channels
    to apply the synchronizer
    :param syncronizer: <str> string indicating the synchronizer name
    """
    pass

def getPlotAxes(self, channel=None):
    """get the PlotAxes for the channel indicated. In case of empty channel
    value it will return  all the PlotAxes Info

    >:param channel: <str> Indicate the channel to return the PlotAxes Info
    """
    pass

def setPlotAxes(self, channels, axis):
    """Set the PlotAxes for the indicated channels.

    :param channels: (seq<str>) a list of strings indicating the channels
    to apply the PlotAxes
    :param axis: <str> string indicating the axis name
    """
    pass

def getSynchronization(self, channel=None):
    """get the Synchronization for the channel indicated. In case of empty
    channel value it will return  all the Synchronization Info

    >:param channel: <str> Indicate the channel to return the
    Synchronization Info
    """
    pass

def setSynchronization(self, channels, synchronization):
    """Set the Synchronization for the indicated channels.

    :param channels: (seq<str>) a list of strings indicating the channels
    to apply the Synchronization
    :param synchronization: <str> string indicating the synchronization name
    """
    pass
```

reszelaz:
> On the other hand, I wouldn't differentiate between 'active_meas_conf_X' and 'meas_conf_X', if the optional parameters worked the default measurement group could be the active one, else use a specific measurement group.

This would be only possible if we place the measurement group parameter as the last one. The optional parameters will have this limitation, similarly to the parameters with default values.
If you don't like the macro names cauese they are too long, we could consider the following two proposals:

**Option 3**
The same as `Option 1`, but the macros for changing the active measurement group configuration would start from `ameas_` instead of `active_meas`:

```
ameas_conf_set <conf param> <value> [<channel>]
ameas_conf_get <conf param> [<channel>]
```

**Option 4**
The same as `Option 3`, but the macros for changing the active measurement group configuration would start from `ameas_` instead of `active_meas`:

```
ameas_conf_set_enabled <param value> [<channel>]
ameas_conf_get_enabled [<channel>]
```
etc.

Another option would be to use private scope for macros which want to change the measurement group which is different than the active one:

**Option 5**

```
meas_conf_set <conf param> <value> [<channel>]
meas_conf_get <conf param> [<channel>]
_meas_conf_set <meas grp> <conf param> <param value> [<channel>]
_meas_conf_get <meas grp> <conf param> [<channel>]
```

**Option 6**

```
meas_conf_set_enabled <param value> [<channel>]
meas_conf_get_enabled [<channel>]
_meas_conf_set_enabled <meas grp> <param value> [<channel>]
_meas_conf_get_enabled <meas grp> [<channel>]
```

dschick:
how does one actually define optional arguments for a macro?

reszelaz:
> how does one actually define optional arguments for a macro?

This is actually not yet possible. It was proposed in #285 and should not be difficult to implement.

teresanunez:
Hi @dschick, giving them a default value, but they have to be put at the end.

teresanunez:
Well, sorry, I gave the solution with restrictions ... as it is discussed in #285

dschick:
@reszelaz  &  @teresanunez  I was actually exactly thinking about the proposal in #285 . So  @teresanunez if I put the parameters at the end with some default values like None or -1 I can do the same parsing as with the Optional parameters, right?

@droldan could you provide me a code snippet how I can enable/disable a counter for plotting from within a macro? I am thinking of a macro like the `plotselect` in SPEC.

teresanunez:
Hi @dschick, I am not sure if I understand your question.
First of all, None is not a default value, if you put None the parameter has to be given as argument of
the macro. You can put -1 as default value, and if you not give it it will take this value. Our users use things like this for taking any action (like requesting the value from the command line) if the value is the default one.

dschick:
@teresanunez you just got i right. I will try updating my macros acordingly.

droldan:
Reimplement the macros by using the MGConfiguration new API #867

reszelaz:
When manually testing #867 I needed some easy way to use the programatic API so I have prepared myself macros for changing the measurement configuration.

There were no votes on the above mentioned proposals (options with macro names and parameters). Only @droldan  said that he would prefer to not create different macros for active and non-active measurement groups. Taking this into account and having the _optional parameters_ and more user friendly way of using repeat parameters in Spock I propose only two macros: `set_meas_conf` and `get_meas_conf`. (Just a note for @rhomspuron and @droldan - this proposal is very similar to what we discussed a year ago or so in the offices at ALBA). 

Here is a sample of usage:

```
Door_zreszela_1 [9]: %get_meas_conf
ActiveMntGrp = mntgrp01
   Channel   Enabled   PlotType   PlotAxes
 --------- --------- ---------- ----------
      ct02      True   Spectrum      <idx>
      ct03      True         No        n/a
      ct04      True         No        n/a

Door_zreszela_1 [10]: %set_meas_conf enabled False ct03 
ActiveMntGrp = mntgrp01

Door_zreszela_1 [11]: %get_meas_conf
ActiveMntGrp = mntgrp01
   Channel   Enabled   PlotType   PlotAxes
 --------- --------- ---------- ----------
      ct02      True   Spectrum      <idx>
      ct03     False         No        n/a
      ct04      True         No        n/a

Door_zreszela_1 [12]: %get_meas_conf all
ActiveMntGrp = mntgrp01
   Channel   Enabled   Output   PlotType   PlotAxes   Timer   Monitor   Synchronizer   ValueRefEnabled   ValueRefPattern
 --------- --------- -------- ---------- ---------- ------- --------- -------------- ----------------- -----------------
      ct02      True     True   Spectrum      <idx>    ct02      ct02       software               n/a               n/a
      ct03     False     True         No        n/a    ct02      ct02       software               n/a               n/a
      ct04      True     True         No        n/a    ct02      ct02       software               n/a               n/a

Door_zreszela_1 [13]: %get_meas_conf all mntgrp04
   Channel   Enabled   Output   PlotType   PlotAxes   Timer   Monitor   Synchronizer   ValueRefEnabled   ValueRefPattern
 --------- --------- -------- ---------- ---------- ------- --------- -------------- ----------------- -----------------
      ct01      True     True   Spectrum      <mov>    ct01      ct01       software               n/a               n/a
      ct05      True     True   Spectrum      <idx>    ct01      ct01       software               n/a               n/a
      ct06      True     True   Spectrum       ct01    ct01      ct01       software               n/a               n/a
      ct07      True     True   Spectrum       ct05    ct01      ct01       software               n/a               n/a
      ct08      True     True   Spectrum       ct06    ct01      ct01       software               n/a               n/a
```

Please take a look on the macros' docstrings for more examples.

Note that I went for reducing the number of macro parameters for `get_meas_conf`. I don't think that users will at the end type the configration parameter name e.g. `PlotAxes` or the channel names e.g. `[ct01 ct02]`, if simply executing `get_meas_conf` will give them all of them in a nice table.

As an extra I added `set_meas` and `get_meas` for setting the `ActiveMntGrp`.

It would be great if you could comment and suggest changes. Thanks in advance!

reszelaz:
The flake8 check fails due to too long docstring lines however I would let them fail cause breaking the lines causes not very nice documentation page in html. I couldn't find a way to use #noqa due to [this](https://gitlab.com/pycqa/flake8/-/issues/375).

src/sardana/macroserver/macros/expconf.py
```diff
@@ -0,0 +1,339 @@
...
+      (for image use "|" as separator of axes e.g. idx|ct01)
+    - **Timer**: <channel name> e.g. ct01
+    - **Monitor**: <channel name> e.g. ct01
+    - **Synchronizer**: software or <trigger/gate name> e.g. tg01
+    - **ValueRefEnabled** - True/1 or False/0
+    - **ValueRefPattern** - URI e.g. file:///tmp/img_{index}.tiff
+
+    Examples of usage:
+
+    >>> set_meas_conf enabled True  # enable all channels in <ActiveMntGrp>
```
cpascual, line 172: why not document like this?

```
    >>>  # enable all channels in <ActiveMntGrp>
    >>> set_meas_conf enabled True 
    >>> # enable spectrum plotting for ct01 on <ActiveMntGrp>
    >>> set_meas_conf plottype spectrum ct01  
    (...)
```

cpascual:
@reszelaz said:
> The flake8 check fails due to too long docstring lines

why not document like this?

```
    >>>  # enable all channels in <ActiveMntGrp>
    >>> set_meas_conf enabled True 
    >>> # enable spectrum plotting for ct01 on <ActiveMntGrp>
    >>> set_meas_conf plottype spectrum ct01  
    (...)
```

reszelaz:
> why not document like this?

Good idea! Fixed in c265192

teresanunez:
Nice work, thanks, IMO it can be merged.

reszelaz:
Thanks for the review Teresa! Merging then..

PR #695, opened 2018-02-06
Delegate push event to EventTH in attribute handler (logging)

### LINKED ISSUE #693: Aborting a macro may cause deadlocks
Using the `AbortMacro` Door's command may cause deadlocks. This started to manifest since we integrated #503. This issue can be triggered with the `MacroButton` which uses this command.

The issue has to do with python logging and the `sardana.tango.core.attributehandler.AttributeLogHandler` which compete for the resources.

reszelaz:
Delegate push event to the EventTH worker thread to avoid deadlocks
in multithreading scenarios. This is because we may be asked to
handle stream messages while the Tango monitor is acquired causing
a deadlock if the owner of the monitor also requires to handle
messages. See sardana-org/sardana#693. Fix #693.

### DIFF
--- a/src/sardana/tango/core/attributehandler.py
+++ b/src/sardana/tango/core/attributehandler.py
@@ -60,7 +60,12 @@ class AttributeLogHandler(logging.Handler):
         attr = self._attr
         if attr is None or dev is None:
             return
-        dev.set_attribute(attr, output)
+        # Delegate push event to the EventTH worker thread to avoid deadlocks
+        # in multithreading scenarios. This is because we may be asked to
+        # handle stream messages while the Tango monitor is acquired causing
+        # a deadlock if the owner of the monitor also requires to handle
+        # messages. See sardana-org/sardana#693
+        dev.set_attribute(attr, output, synch=False)
 
     def read(self, attr):
         """Read from the buffer and assign to the attribute value"""


reszelaz:
Apparently it has sporadic problems on the MacroServer start. Simply the server does not start. This was revealed with the travis-ci which (re)starts the servers multiple times:

These are the traces:
```
[...]
MainThread     DEBUG    2018-02-06 18:40:18,284 MS_demo1_1.MacroManager: Adding macro class wum
MainThread     DEBUG    2018-02-06 18:40:18,287 MS_demo1_1: Creating pool Pool_demo1_1
MainThread     DEBUG    2018-02-06 18:40:18,402 Door/demo1/1: -> init_device
MainThread     DEBUG    2018-02-06 18:40:18,405 Door/demo1/1: <- init_device
MainThread     INFO     2018-02-06 18:40:21,610 TaurusRootLogger: Exiting
MainThread     CRITICAL 2018-02-06 18:40:21,610 TaurusRootLogger: Server exited with DevFailed
  File "/home/zreszela/workspace/sardana/src/sardana/tango/core/util.py", line 992, in run_tango_server
    util.server_init()
DevFailed: DevFailed[
DevError[
    desc = Not able to acquire serialization (dev, class or process) monitor
  origin = TangoMonitor::get_monitor
  reason = API_CommandTimedOut
severity = ERR]
]
MainThread     INFO     2018-02-06 18:40:21,611 TaurusRootLogger: Exited
terminate called after throwing an instance of 'omni_thread_fatal'
Aborted
```
To be investigated...

reszelaz:
This looks like another deadlock but this time between one of the Tango threads making the server initialization and the EventTH thread trying to push the change event of the `DebugIt` decorator that decorates the Door's `init_device`. I have submitted a question to the Tango experts: https://github.com/tango-controls/TangoTickets/issues/12.
Let's wait a couple of days for the feedback, otherwise we should look for another solution or a workaround eliminating the problematic debug messages.

reszelaz:
Since there are no news yet from the Tango experts, I will propose a workaround in another PR and move this to Jul18.

PR #696, opened 2018-02-06
Remove unused methods

mrosanes:
Remove unused methods stop_axes and abort_axes


reszelaz:
This PR is part of the work proposed in https://github.com/sardana-org/sardana/pull/592#issuecomment-362026115. Since these methods are not used by sardana and I really doubt that they are used anywhere outside I decide to remove them. There is no plan to bump the major version due to that and we will just mention it in the changelog. The equivalent methods are `stop_elements` and `abort_elements`.

PR #697, opened 2018-02-12
Correct documentation to reflect the implementation

reszelaz:
The Read and State sequences executes the hardware queries algorithms
per controller concurrently. This is the desired behavior.
Correct the documentation so it reflects the implementation.


reszelaz:
Since there are no negative comments I'm merging it myself.

PR #699, opened 2018-02-13
Logstash log handler

reszelaz:
The main purpose of this PR is to allow the use of [elastic stack](https://www.elastic.co/) for:
* logging analysis i.e. merging logs of Pool and MacroServer in a time ordered events
* extracting metrics from the logs about the system performace e.g. server restarts, macro errors, etc.

This PR adds a logging handler for pushing log events to the logstash. An alternative solution, well actually the recommended by elastic one, would be to use Beats. However Beats require a modern OS to run and we at ALBA are stuck with openSUSE 11.1 which does not support it. So, it adds an optional dependency to the [logstash](https://pypi.python.org/pypi/python-logstash/0.4.6) module.

Configuration is done via Tango properties (other soltutions e.g. sardanacustomsettings were also evaluated). Properties `LogstashHost` and `LogstashPort` are available on both Pool and MacroServer devices. In case a single Sardana server is in use, the Pool properties take precedence over the MacroServer properties.

Since this was not yet used in a real environments I decided to mark the new API as provisional just to make easier the eventual improvements.

Exemplary configuration of the elastic components can be found in https://github.com/reszelaz/sardana-elastic.

@sardana-org/integrators could anyone take a look on this?

### DIFF
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -890,6 +890,61 @@ def prepare_taurus(options, args, tango_args):
     factory.disablePolling()
 
 
+def prepare_logstash(args):
+    """Prepare logstash handler based on the configuration stored in the Tango
+    database.
+
+    :param args: process execution arguments
+    :type args: list<str>
+    """
+    log_messages = []
+    try:
+        import logstash
+    except ImportError:
+        msg = "Unable to import logstash. Skipping logstash configuration..."
+        log_messages.append(msg)
+        return
+
+    def get_logstash_conf(dev_name):
+        try:
+            props = db.get_device_property(dev_name, "LogstashHost")
+            host = props["LogstashHost"][0]
+        except IndexError:
+            host = None
+        try:
+            props = db.get_device_property(dev_name, "LogstashPort")
+            port = int(props["LogstashPort"][0])
+        except IndexError:
+            port = 12345
+        return host, port
+
+    db = Database()
+
+    bin_name = args[0]
+    instance_name = args[1]
+    server_name = bin_name + "/" + instance_name
+    if bin_name in ["Pool", "MacroServer"]:
+        class_name = bin_name
+        dev_name = get_dev_from_class_server(db, class_name, server_name)[0]
+        host, port = get_logstash_conf(dev_name)
+    else:
+        dev_name = get_dev_from_class_server(db, "Pool", server_name)[0]
+        host, port = get_logstash_conf(dev_name)
+        if host is None:
+            dev_name = get_dev_from_class_server(db, "MacroServer",
+                                                 server_name)[0]
+            host, port = get_logstash_conf(dev_name)
+
+    if host is not None:
+        root = Logger.getRootLog()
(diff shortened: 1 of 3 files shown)


reszelaz:
In order to test it one does not need to setup the whole elastic stack. It is just enough to run logstash and see if sardana logs are being processed by its pipeline. Some ideas on how to test it (assuming installation on Debian 9):
1. Install logstash.
2. Place the following pipeline configuration file: [sardana.conf.txt](https://github.com/sardana-org/sardana/files/1724106/sardana.conf.txt) in a directory e.g. `/etc/logstash/conf.d` (remember to remove the .txt extension).
3. Start logstash from command line using the following command: `sudo /usr/share/logstash/bin/logstash --path.settings=/etc/logstash --path.config=/etc/logstash/conf.d`
4. Configure the device properties `LogstashHost` to `localhost` and `LogstashPort` to, for Pool to `5959` or for MacroServer to `5960` (the pipeline configuration file assumes these ports, but these are configuratble as far as they are consistent on both sides).
5. Start a server, either Pool or MacroServer.
6. You should see the sardana log events printed on stdout on of the logstash process.

teresanunez:
Hi @reszelaz, I have tested it following your instructions. I see in the window running the logstash process printouts like:

{
          "level" => "DEBUG",
           "tags" => [
        [0] "macroserver"
    ],
     "@timestamp" => 2018-02-15T11:40:54.376Z,
           "path" => "/home/tnunez/sardana_logstash-log-handler/src/sardana/macroserver/msmacromanager.py",
        "message" => "[ END ] runMacro Macro 'mv([['mot53', '1.0']]) -> 146adb6e-1245-11e8-a327-901b0e40a03c'",
        "program" => "MacroServer",
           "port" => 54301,
    "logger_name" => "Door/logstash/1.MacroExecutor",
           "host" => "haso111n",
           "type" => "logstash",
       "@version" => "1"
}
{
          "level" => "DEBUG",
           "tags" => [
        [0] "macroserver"
    ],
     "@timestamp" => 2018-02-15T11:40:54.377Z,
           "path" => "/home/tnunez/sardana_logstash-log-handler/src/sardana/macroserver/msmacromanager.py",
        "message" => "Job ended (stopped=False, aborted=False)",
        "program" => "MacroServer",
           "port" => 54301,
    "logger_name" => "Door/logstash/1.MacroExecutor",
           "host" => "haso111n",
           "type" => "logstash",
       "@version" => "1"
}

Is this what one should expect?. Many thanks.

reszelaz:
Great! Exactly. These are the log events that are printed to the stdout after processing them by the logstash pipeline. If you have an elasticsearch instance up and running you could uncomment the following lines from the pipeline configuration:
```
output {
    elasticsearch {
        hosts => ["localhost:9200"]
    }
}
```
and these events would be then indexed by the elasticsearch. Afterwards with the kibana web you could analyze the logs, make queries, etc. But I think that this is out of the scope of the PR.

Whenever we give it a try in production and everything works as expected I would add an example to the wiki or the documentation.

Issue #694, opened 2018-02-05
Software synchonizer fails if motor controller ReadOne failed


reszelaz:
Software synchronizer needs protection for cases when the motor controller `ReadOne` method fails. We should consier this errors and log them as warnings but the software synchronization should continue - this may be just a temporary unavailability of the motor's position.

reszelaz:
Fixed by #700

PR #700, opened 2018-02-15
Add protection for SW synchronizer

mrosanes:
Add protection for software synchronizer for cases when the
motor controller ReadOne method fails. Log the errors in
the Pool logs, and continue the scan, given that may be
just a temporary unavailability of the motor's position.


mrosanes:
This PR is dealing with issue #694. It is ready to be reviewed.

PR #710, opened 2018-02-26
Add internal reference to keep subscribed attrs alive

cmft:
Sardana extensions wrongly expect attributes to be
alive even if no one keeps a ref to them.
After https://github.com/taurus-org/taurus/pull/692, this is no longer the case, and
sardana extensions must keep their own references.


reszelaz:
All changes looks ok to me, thanks! I've not done tests and rely on travis-ci and https://github.com/sardana-org/sardana/issues/28#issuecomment-368482827. Merging..

PR #712, opened 2018-02-27
spock commands not visible with bright background

### LINKED ISSUE #706: spock commands not visable with bright background
Hi,

I've just tested spock with taurus 4.3 and sardana 2.3.3 an I noticed that for themes with a  bright background terminal I do not see command I type. Probably it can be change somewhere but by default I assume that command in spock should be visible in all themes. i.e. take from the system.

Best regards,
Jan

cmft:
Using Spock with a bright background terminal themes commands are not visible
It seems related with an IPython V5  bug https://github.com/ipython/ipython/pull/9655

Change spock colors to  'Neutral' for IPython >= V5

Fix #706

### DIFF
--- a/src/sardana/spock/ipython_01_00/genutils.py
+++ b/src/sardana/spock/ipython_01_00/genutils.py
@@ -1063,6 +1063,9 @@ object?   -> Details about 'object'. ?object also works, ?? prints more.
     i_shell.confirm_exit = False
 
     if ipy_ver >= 50000:
+        # Change color for ipy_ver >= 50000 due to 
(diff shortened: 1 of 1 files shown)


reszelaz:
I tested it on gnome and the bug is fixed now. Also on kde everything works as before, well almost, the color of the numbers changed from violete to green - but for me this is ok:). Merging. And I think that we should bring it to the release branch as well. Thanks to all involved!

teresanunez:
Hi @reszelaz, many thanks, yes, please, make the change also in the release.

reszelaz:
I just pushed merged the develop into release-Jan18.

PR #715, opened 2018-03-02
Detail description for counter plot output

mrosanes:
Detail description for counter plot output


reszelaz:
Thanks! Now it is more clear. Merging it.

PR #720, opened 2018-03-09
Do not use os.path.join

mrosanes:
os.path.join is failing in windows for this case.
This was making the scans to abort when a h5 files was indicated as ScanFile.

Use the concatenation with plus signs instead.


mrosanes:
In the documentation of h5py we can see that the separators for hdf5 files are slashes '/'

http://docs.h5py.org/en/latest/quick.html

<pre>
“HDF” stands for “Hierarchical Data Format”. Every object in an HDF5 file has a name, and they’re arranged in a POSIX-style hierarchy with /-separators
</pre>

PR #726, opened 2018-04-03
SEP16 - Plugins (controllers, macros, etc.) catalogue

reszelaz:
SEP15 migrated the Sardana project from SourceForge to GitHub but left the third-party repositories, controllers and macros at their original location. [This SEP16](https://github.com/reszelaz/sardana/blob/sep16/doc/source/sep/SEP16.md) takes care about the third-party repositories reorganization so the plugins maintenance will become more developer friendly and plugins more accessible to the user. This is achieved by simply transferring the project ownerships from the Sardana administrators to the plugins developers at the same time giving some advices on how to organize the projects based on the current experience.

It is totally opened for changes (this are just my ideas written down very quickly...). I encourage you drop comments ideas.

Thanks in advance!


cpascual:
Some comments (just half-thought ideas):

- You propose creating a new repository, called sardana-plugins to act as the register. Have you considered using (a subdir of) the sardana repo itself as the register?. It would have the advantage of not having to maintain an extra repo and of giving visibility to new plugins... but it would also add noise into the main repo issues. Another possibility is to use the sardana wiki for that (which is actually a separated repo too). 
- You propose that people do PRs for registering new plugins. And you propose that some basic info should be provided when registering. Consider using [PR templates](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/) for this.
- Consider recommending to use setuptools in the plugin repos. This way you can eventually declare entry points in sardana and let the plugins insert themselves using them (something like what we did in [this Taurus feature](https://github.com/taurus-org/taurus/pull/684), which is a proof-of -concept for [TEP13](http://www.taurus-scada.org/tep?TEP13.md)

rhomspuron:
We can include the common plugins like IcePAP, Lima, AlbaEm, as repos in sardana-org. The permission can be shared by the developers and the sardana-core team,  the developers are the responsible to accept the pull request. For me, it is easer to find a plugin on the same organization. 

In addition, for the specifics plugin like Claess macros, the developers take care about where put the repo.

In both cases the developers must register the plugin on sardana-core.

reszelaz:
I'm glad to announce you that we agreed the design and implementation plan of this SEP. So it was moved to the CANDIDATE state.

Automatically this opens the [Plugins Catalogue](https://github.com/sardana-org/sardana-plugins/blob/master/README.md) to the developers. So, from now on, you can register your projects there. You all are very welcome to do it - just follow the [CONTRIBUTING.md](https://github.com/sardana-org/sardana-plugins/blob/master/CONTRIBUTING.md) guide.

If you find anything that can be improved or have suggestions of any kind, don't hesitate to post it here.

reszelaz:
I just added a link to the side navigation bar of the Sardana Documentation Home Page. This will increase the visibility of the catalogue. Hope you are fine with that.

dschick:
Hej,

I am very happy to see that the plugin register starts to fill.
I would also like to contribute some of our work here at MBI to the register.

I am just wondering what is the distinction between putting a hardware controller in the hardware register or as part of a beamline/experiment?

Ideally I would like to have an repository for an beamline/experiment including macros and hardware controllers and the individual controllers should be linked to hardware controller repos that also appear in the sardana plugin hardware register.

Otherwise, it is hard to find hardware controllers that are hidden in an beamline/experiment repository. So essentially every hardware controller should be listed in the hardware register and can be listed as part of a beamline/experiment plugin...

reszelaz:
> I would also like to contribute some of our work here at MBI to the register.

Great!
 
> I am just wondering what is the distinction between putting a hardware controller in the hardware register or as part of a beamline/experiment?
> 
> Ideally I would like to have an repository for an beamline/experiment including macros and hardware controllers and the individual controllers should be linked to hardware controller repos that also appear in the sardana plugin hardware register.
> 
> Otherwise, it is hard to find hardware controllers that are hidden in an beamline/experiment repository. So essentially every hardware controller should be listed in the hardware register and can be listed as part of a beamline/experiment plugin...

If I understand well, you prefer to have only one project (repo) for all you plugins, even if these includes controllers which are generic enough to be used outside of your lab. Then you plan to list this generic plugins in the hardware category, as well as your specific plugins in the system category, while both will point to the same project.

Even if we [advice](https://github.com/sardana-org/sardana/wiki/How-to-organize-your-plugin-project) to split these projects, I think it is also fine to do it like you describe. At the end, this is what we wanted to achieve - that everyone chooses how to organize the projects. The important thing is that the relevant information/links are accessible from the catalogue.

amilan:
I know that is always better to stay away from git submodule, but in this case it may be a valid option to consider. One can have a beamline/experiment repository with the specific controllers/macros/etc and link the generic controllers via git submodule. Probably it's not the easiest option, but at least the option is there.

If it helps, at MAXIV we use Ansible for our SW deployment and configuration and we have adapted our beamline inventories to be able to deploy and configure all the necessary packages.

Just to illustrate it a bit better the process: we have the generic repositories and the beamline specific repositories( as an example, look at the [biomax beamline repository](https://github.com/MaxIV-KitsControls/sardana-biomax)).
Then, we create our RPM packages for those projects (e.g. sardana-albaem, sardana-biomax, sardana-icepap, etc) using gitlab-ci.
After that, we list those packages as a dependencies for the beamlines (in Ansible).
And finally, we have develop a tango module for Ansible that allows to define and configure tango servers. In that way, we configure the PoolPath and MacroServerPath via ansible.

As a result, Ansible is our link between generic and specific controllers/macros.

I hope it helps.
Antonio.

dschick:
Thanks @amilan for this detailed description.

Is there maybe alternative strategies with out external tools like Ansible in order to really use the system plugins for automatic installation/updates of all required controllers/macros in case you also want to use sub-repositories?
An obvious example would be the TangoAttribureControllers, which e.g. I would like to automatically include into my system repo.

I am simply interested in how you organize your systems properly.

Thanks Daniel

reszelaz:
Hi Antonio, Daniel,

Very interesting to hear how you organize it at MaxIV using Ansible. At ALBA we are still not at this level. We are still deciding if we will package with Debian our plugins or will simply make _git clone_. It looks like either way we will install in the Python system path. So, the second option (git) will be complemented by the pip installation. Meanwhile we have some heterogeneous [Salt](https://www.saltstack.com/) (another SCM system similar to Ansible) _recipies_ (basically git clones + some copies) at 3 beamlines which start to use Sardana with Python 3. There we also manually set the PoolPath properties.

I can not comment the git submodule option, simply becasue I don't know it. If someone tries it it would be great to comment here how it works in the context of the Sardana plugins repositories.

Recently I have heared about another tool for managing multiple repositories called [repo](https://gerrit.googlesource.com/git-repo/) (available for Debian). Maybe this could be helpful here. I have a feeling it will be easier then the use of git submodule.

For example with this simple configuration [file](https://github.com/reszelaz/repo_test/blob/master/default.xml) and executing:
```console
mkdir my_plugins
cd my_plugins
repo init -u https://github.com/reszelaz/repo_test.git
repo sync
```
One could obtain two repositories: sardana-alba and sardana-ni660x.

Going further, one could add projects with plugins for the whole Sardana system, specify revisions, etc. Here are the docs for the [repo](https://source.android.com/setup/develop/repo) command and for the [minifest.xml](https://gerrit.googlesource.com/git-repo/+/master/docs/manifest-format.md) file.

reszelaz:
During the last [follow-up meeting](https://github.com/sardana-org/sardana-followup/blob/master/20191128/MINUTES.md) (see point 3) we agreed to ACCEPT this SEP. Thanks to all for your participation in the dicsussions!

I just made two commits to the old repos about their discontinuation as the official repos of the Sardana controllers/macros.

You can still use this thread for any related discussions if you don't find a better place to do it.

PR #728, opened 2018-04-06
Adding predefined values for string arguments

teresanunez:
This PR is created for opening a discussion about the implementation of predefined values for string macro arguments, requested from some DESY beamlines. 
The request: they would like to have some predefined values appearing in the screen when they pulse 
'Tab' after a macro name, in case the argument is of type String (in the same way as now, for example, one gets the name of all available Moveable sardana objects pressing 'Tab' after typing 'mv' in spock).
The predefined values are different for each macro.
The implementation I did works like this: one can define an environment variable with the name of the macro + 'Predefined' (as a list). The values in this variable will appear pressing Tab when the name of the macro is typed and the next argument is of type String. 
The implementation is done in the file genutils.py.

### DIFF
--- a/src/sardana/spock/ipython_01_00/genutils.py
+++ b/src/sardana/spock/ipython_01_00/genutils.py
@@ -638,6 +638,13 @@ def _macro_completer(self, event):
     ms = get_macro_server()
 
     macro_name = event.command.lstrip('%')
+
+    env = ms.getEnvironment()
+    pre_values = None
+    for el in env:
+        if el == macro_name + "Predefined":
+            pre_values = ms.getEnvironment()[el]
+
     # calculate parameter index
     param_idx = len(event.line.split()) - 1
     if not event.line.endswith(' '):
@@ -654,6 +661,9 @@ def _macro_completer(self, event):
         res = []
         for param in possible_params:
             res.extend(ms.getElementNamesWithInterface(param['type']))
+        if param['type'] == 'String':
+            for value in pre_values:
+                res.append(str(value))
         return res
 
 


teresanunez:
One disadvantage of this implementation (because of being done in genutils.py) is that the predefined values will be only shown in spock, not in any other clients.
It will be investigated the possibility of creating a new object type in sardana, so that the objects of these type will be shown for this type arguments. The disadvantages would be that one have to define the macro already with this new argument type, so now in advance that one could want to use some
predefined values (the current solution works for all available macros), the predefined values will appear for all macros with arguments of this type (in the current implementation each macro have its own predefined values) and the user has to know how to add objects of that type (now they have only to define a new environment variable in spock).

reszelaz:
Many thanks for this proposal! I like very much this idea and think that many users would appreciate it.

I see also some more, very similar needs. For example, there exist a macro for homing the IcePAP motors. And we would like to give hints to the user with only the IcePAP motors and validate if the motor passed as parameter belongs to the IcePAP controller.

So I think that when implementing your original request we should already think about extending it for other cases as well.

Now regarding the implementation:

Yes, we would need to modify the `_macro_completer` funcion. But IMHO the environment variable is not the best place where to specify the list of options. I see it more like a part of the macro parameter definition. A more conservative approach could be for example: 

```python
class pt1(Macro):
    """Macro with one float parameter: Each parameter is described in the
    param_def sequence as being a sequence of four elements: name, type,
    default value and description. Optional dictionary is used as fifth element
    to pass hints to the clients and decoders.
    Usage from Spock, ex.:
    pt1 1
    """

    param_def = [
        ['value', Type.Float, None, 'some bloody float', {"options": [2.7182, 3.1415]}]
    ]

    def run(self, f):
        pass
```
Or more revolutionary approach would be to refactor the whole macro parameters definition to use dictionaries (of course maintaining the backwards compatibility) - IMO more elegant:

```python
class pt1(Macro):
    """Macro with one float parameter: Each parameter is described in the
    param_def sequence as being a dictionary with at least keys name, type,
    and description. Optional keys are: default_value, options.
    Usage from Spock, ex.:
    pt1 1
    """

    param_def = [
        {'name': 'value',
         'type': Type.Float,
         'description': 'some bloody float',
         'options': [2.7182, 3.1415]}
    ]

    def run(self, f):
        pass
```

This *options* information would be then exported from the server to the clients via JSON (it is alrady implemented). Then the clients would need to use it. Spock in the completer and the macroexecuted populating the comboboxes accordingly.

As an alternative for for the requirement to parametrize the list of options I would suggest to program macros with two level hierachy. First an internal macro with parameters without options e.g. `_pt1` and then a specific `pt1` macro specifying the parameters with options and just calling the other one internally.

teresanunez:
Hi @reszelaz, many thanks for your comments. The only doubt that I have is that implementing it like this, the predefined values are defined by the person programming the macro ... and are not flexible or specific for each user. I mean, they can not define predefined values for already defined macros (well,
 they could make it calling the existing macro from a new one ... but in any case it would need a new macro with a new name).

reszelaz:
Yes, if we implement it within the macro parameters definition it would require this two-level macro hierarchy.

Another way could be to mix both approaches. The macro definition could refer to some configuration medium e.g. [sardanacustomsettings](https://sardana-controls.org/users/configuration/sardanacustomsettings.html) or if this is a custom macro a similar configuration file. Or an environment variable.

teresanunez:
I think it would be good it both approaches are mixed.

teresanunez:
#877 is created to deal with this new proposed implementation.

PR #730, opened 2018-04-09
Adding filter to spock logging filehandler

teresanunez:
The output of the implemented  spock logging feature was not accepted by some users.
A filter is added for adapting the output to the user requests.

### DIFF
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -42,6 +42,8 @@ import threading
 
 from lxml import etree
 
+import time
+
 from PyTango import DevFailed
 
 try:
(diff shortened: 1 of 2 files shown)


reszelaz:
I finally found some time to look at this PR. Sorry for the delay!
I've tried it and it works nice - the logged messages are stripped from _unnecessary_ information.

When analyzing the code I found it hard to understand which messages are not possing through the filter. Could you please provide some information, either in the forom of comments or docstring, with the examples of message that will be converted and filtered out.

reszelaz:
We were thinking about this proposal and find it more interesting to allow the user to add a custom filter instead of hardcoding one in the core. We propose two options of configuring this:
* [sardanacustomsettings](https://sardana-controls.org/users/configuration/sardanacustomsettings.html)
* MacroServer environment variable

The second option would allow to set it per MacroServer instance or per Door.

teresanunez:
Hi @reszelaz, I think it could be good to have some filters already defined in Sardana (at least one, the
 one that is there now) and allow to choose one prefefined one or a new one defined by the users.
Our users want was is there now and it is not so easy to implement it. Perhaps we could put both possibilities.

reszelaz:
Yes, I agree about providing the default filter class. It can be the one that you implemented. I would just leave the possibility to disable it, and eventually enable another one.

In this case I think that the easiest would be to add a variable to the sardana custom settings, for example, LOG_MACRO_FILTER with the default value "sardana.macroserver.msmacromanager.LogMacroFilter" (see that I suggest to rename the class `FileHandlerFilter` to `LogMacroFilter`).

Then the instantion of the filter could look very similar to the one implemented in Taurus for the formatter (see [how the setting is obtained](https://github.com/taurus-org/taurus/blob/develop/lib/taurus/qt/qtgui/base/taurusbase.py#L150) and [how the class could be obtained](https://github.com/taurus-org/taurus/blob/develop/lib/taurus/qt/qtgui/base/taurusbase.py#L771)).

If you prefer we can take care about this implementation.

teresanunez:
Hi @reszelaz. many thanks for your comments, I can try to implement it (unless you prefer to do it by yourself instead of having to review my code afterwards). Just let me know, for me it is not a problem to try it.

reszelaz:
Go ahead, I will focus on reviewing other PR for the Jul18. If you need any help just let us know! Many thanks!!

teresanunez:
Hi @reszelaz, I have added the modifications for allowing the users to define the own filters.

reszelaz:
Great! Thanks for implementing the filter configuration. I just made some cosmetic changes and whenever travis gets green I will merge it.

Just a summary of its use that later on could be used in the documentaion:
- if `LOG_MACRO_FILTER` is not defined no filter will be used.
- if `LOG_MACRO_FILTER` is wrongly defined warn user and no filter will be used
- if macro filter can not be initialized warn user and no filter will be used

PR #738, opened 2018-04-23
Emit Tango events for measurement group attributes: Moveable, Synchronization and LatencyTime

reszelaz:


reszelaz:
This also fixes the testsuite that on some environments was hanging on Debian 9 (Tango 9).
Now, we can finally add this environment (together with Taurus 4) to the travis-ci.

PR #743, opened 2018-05-02
Add CI for Debian 9 and Taurus 4

reszelaz:
Use two different docker images for CI:
- Debian 8 together with Taurus 3
- Debian 9 together with Taurus 4


reszelaz:
After fixing the sardana-test docker image this finally works.

Can someone from @sardana-org/integrators take a look on it, please?

PR #745, opened 2018-05-09
Add repeat macro to standard macros

reszelaz:
Add repeat macro (from third-party repository:
https://sourceforge.net/p/sardana/macros.git/ci/master/tree/ALBA_GENERAL/tools.py).
Fix issue-308.

Just after integrating it we should add a comment in the third-party repo that this was moved to the standar macros.

@dschick could you test it?

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -23,9 +23,9 @@
 
 """This is the standard macro module"""
 
-__all__ = ["ct", "mstate", "mv", "mvr", "pwa", "pwm", "set_lim", "set_lm",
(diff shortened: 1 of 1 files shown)


dschick:
I can try testing it within the next days. Do you rely on any test framework or are you asking me more for a real-world testing?

reszelaz:
We have test framework for some part of the code but I have not added any tests for this macro in order to integrate it fast for your needs. Actually, this macro is used quite often at Alba - I do not expect problems..

teresanunez:
I have tested it with the sequencer and in spock (adding a general hook in the position 'body') and
 it works fine.

reszelaz:
Thanks for the review @teresanunez. I'm merging it..

PR #746, opened 2018-05-10
Revert workaround from travis-ci for problems with mysql

reszelaz:
After applying a workaround in the sardana-test docker image it may be enough to remove this workaround from the travis-ci. Trying it...


reszelaz:
Seems like there are still some problems with the docker image. I'm closing this for the moment...

Issue #744, opened 2018-05-09
nested scans with sequencer


dschick:
Hej all,

I would like to use the sequencer for doing a set of nested scans similar to a mesh scan but having a single scan for the inner loop in the SPEC file.
Following the sequencer example in the documentation I simply nested two ascans for example having an outer loop for motorA and an inner loop for motorB, which is the one I would like to have my individual scans in the SPEC file.
Now the problem is that sardana also acquires data for the outer loop and also saves them into the SPEC file. This to some extent corrupts the SPEC file because there is an extra line of data atfer each scan of the inner loop. So i.e. PyMCA puts this last line of data into one of the inner loop scans.

My question is now, if there is a macro that is just scanning a motor without acquiring data and which is also compatible with the sequencer?

In good old SPEC I would have writting a simple macro with some for-loops instead, but I pretty much appriciate the Sequencer of Sardana. If there is no such scanning macro available I might try to write my own.

Best

Daniel

teresanunez:
Hi Daniel,
                   if you set in the sequencer something like it is in 
the attached image you
  get in the file only the scans of the inner motor, the outer motor 
does not appear.
  Perhaps this helps.

                             Regards,
                                              Teresa

teresanunez:
Hi again, 
                I am not sure if the image was attached, what I mean is to add the macro:

senv ScanFile mytestname.spec

before the inner scan and the macro:

usenv ScanFile

after the inner scan.

                          Regards,
                                             Teresa
![send](https://user-images.githubusercontent.com/5548680/39815724-4ad8aa6a-5399-11e8-8173-22fcf3b39013.jpeg)

reszelaz:
Very good idea @teresanunez!  I would also do it this way. I think it should enough to usenv the ScanFile just once - before the outer scan. This is because the scan creates the recorders at the beginning of the execution. Then the ScanFile environment variable is not evaluated in between the scan steps.

The only disadvantage is that the acquisition will take place anyway. Maybe you could create a dummy measurement group and play with the ActiveMntGrp environment variable in a similar way...
Actually, I think it should not be too difficult to make the scan to not acquire if the ActiveMntGrp is not defined and just scan the motor as you said.

Hope that this solution is good enough for your needs.

dschick:
thanks a lot.
seems like a workaround that should do it.
Is there also a macro to just repeat a scan several times like in a loop?

Best

Daniel

reszelaz:
[...]

PR #747, opened 2018-05-10
Fix appending of hooks from sequence XML

reszelaz:
#655 fixed hooks property however the MacroExecutor
was still using it in the wrong way. Properly use the appendHook method
to append hooks instead of setting the whole list of hooks.

Reported in #744.


reszelaz:
@dschick could you please confirm that it fix the issue reported in #744. Afterwards we will merge it into develop. Many thanks in advance!

reszelaz:
Thanks for the test. I'm mergin it..

PR #750, opened 2018-05-14
change EOL to always \n in SPEC_FileRecorder

dschick:
on Windows the default EOL is CR LF also for the SPEC files but this is not readable for programs like PyMCA. Force the EOL to only LF using io package instead of python's default io lib

### DIFF
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -33,6 +33,7 @@ import os
 import time
 import itertools
 import re
+import io
 
 import numpy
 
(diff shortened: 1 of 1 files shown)


cpascual:
Note: this has been discussed in https://github.com/sardana-org/sardana/issues/206#issuecomment-387649749

cpascual:
Note that using the `io` package implies using unicode in all writings. This could imply a performance penalty, but on the other hand it prepares us better for supporting python3, so I am merging the PR.

cpascual:
Thanks @dschick !

PR #711, opened 2018-02-26
Release Jan18

reszelaz:
This is a PR **to master** for the official [Jan18 release](https://github.com/sardana-org/sardana/milestone/3). Final tests (automatic and manual) should be done based on it before accepting the merge.

So, **this is a call for help from @sardana-org/integrators and any other person willing to help** to test this branch ASAP in their respective systems. 

We, at ALBA offer to do the manual tests for OpenSuse11.1 and Windows7 (unless someone else volunteers for doing them).

Notes:
 - The appveyor artifacts from this PR can be used for manual tests.
 - A suggested check-list for the manual tests can be found [here](https://github.com/sardana-org/sardana/blob/develop/doc/how_to_release.md).

PR #753, opened 2018-05-14
Fix Pre-Scan Snapshot

amilan:
This commit fixes the pre-scan snapshot issue where the values where not
saved before running an scan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This file follows the formats and conventions from [keepachangelog.com]
 ## [Unreleased]
 
 ### Fixed
+- Push change events from code for measurement group attributes: moveable,
+  latency time and synchronization (#736, #738)
 - Validation of starts and finals for a2scanct, a3scanct, meshct, ... (#734)
 
 ## [2.4.0] 2018-03-14
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -747,10 +747,6 @@ class GScan(Logger):
         ret = []
         for src, label in elements:
             try:
-                is_full_tango_name = src.startswith('tango://')
-                if not is_full_tango_name:
-                    src = 'tango://{0}'.format(src)
-
                 if src in all_elements_info:
                     ei = all_elements_info[src]
                     column = ColumnDesc(name=ei.full_name,
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -749,7 +749,7 @@ class GScan(Logger):
             try:
                 is_full_tango_name = src.startswith('tango://')
                 if not is_full_tango_name:
-                    src = 'tango://{}'.format(src)
+                    src = 'tango://{0}'.format(src)
 
                 if src in all_elements_info:
                     ei = all_elements_info[src]
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -747,6 +747,10 @@ class GScan(Logger):
         ret = []
         for src, label in elements:
             try:
+                is_full_tango_name = src.startswith('tango://')
+                if not is_full_tango_name:
+                    src = 'tango://{}'.format(src)
+
                 if src in all_elements_info:
                     ei = all_elements_info[src]
                     column = ColumnDesc(name=ei.full_name,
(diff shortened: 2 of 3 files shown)


amilan:
Not sure if that's the best place for fixing this issue, but it's an start.

reszelaz:
@amilan for the PR and @dschick for testing it! There is one thing which I don't understand from your comment. What do you mean by saved before running a scan?

The pre-scan snapshot implementation was not reviewed during the FQDN adaptations done in rush during the Jan18 release - see https://github.com/sardana-org/sardana/pull/711#issuecomment-371932946. Furthermore, during the release tests it was observed that they were not working 100% right - see https://github.com/sardana-org/sardana/pull/711#issuecomment-372349865.

From my tests I see that the external attributes e.g. Tango attributes work correctly, but the Sardana objects e.g. motors do not. What is strange to me is that both references are stored in the environment variables identically (having into account that one is a device and the other is an attribute):

~~~
In [1]: import shelve
In [2]: env = shelve.open("/home/zreszela/workspace/MacroServer/zreszela/macroserver.properties")
In [3]: env["PreScanSnapshot"]
Out[3]: 
[('pc255.cells.es:10000/sys/tg_test/1/ampli', 'ampli'),
 ('pc255.cells.es:10000/motor/motctrl01/2', 'mot02')]
~~~

So, I suspect that there must be some workaround code left from the times that Sardana was not using URI object references. Before applying this PR I would go and look for this code. I think that removing this workarounds should also fix this problem.

What seems also wrong to me is that the pre-scan snapshot references are missing "tango://" part. Most probably the same workaround code is stripping it.

amilan:
Hi, 

> What do you mean by saved before running a scan?

I mean, the snapshots were not saved when executing scans. 

>From my tests I see that the external attributes e.g. Tango attributes work correctly, but the Sardana objects e.g. motors do not. What is strange to me is that both references are stored in the environment variables identically (having into account that one is a device and the other is an attribute):

That's weird because I have only tested it with Sardana objects, basically Motors. Maybe we have different versions ... I'm using a fork from the development branch. 

> What seems also wrong to me is that the pre-scan snapshot references are missing "tango://" part. Most probably the same workaround code is stripping it.

I'm also worried that there should be another hidden issue, that's why I wrote in my comment that I was not sure if that was the best place to fix this issue, I wanted some feedback because to me, this PR seems more a workaround than a fix to the main problem.

... maybe I should add WIP to the title ..

dschick:
I can confirm, that the snapshots were working with Tango objects and not with Sardana motors/counters with the latest fork of the develop-branch.

amilan:
:+1:  something to check on my side, then.

amilan:
Hi again, 
I had another quick look to the issue and found the following. In the experiment config, we are removing the ```tango://``` part of the name. It's still something needed? Because commenting that part didn't give me the error, but maybe I'm missing something in somewhere else that I still don't know (I just did quick tests, nothing super exhaustive).

Also, after removing the ```/tmp/tango/MacroServer/macroservername/macroserver.properties``` and generating again the configuration, doesn't show the duplicated entries.

Let me know your opinion.

reszelaz:
@amilan, do you you think you will be able to work on this for the Jul17 release?

amilan:
Hi, 

Sorry, I've been on vacations and in a conference, but I'm back, finally. I will continue with this issue on Monday.

reszelaz:
Great news @amilan and don't worry!

amilan:
Hi again, 

I've found that the expconf widget was loading the snapshot list and always appending 'tango://' to each element. That's why it was giving the error that Zibi commented. Checking if this appending is needed or not fixes the issue. I've tested it with sardana elements and extra tango attributes and seems to work properly.

Now the question is if this fixes are enough or if we want to implement something like what Zibi was commenting in his latest suggestion.

In any case, please have a look to the PR and let me know your opinion.

reszelaz:
Great! I've checked it and works nicely.

> Now the question is if this fixes are enough or if we want to implement something like what Zibi was commenting in his latest suggestion.

I implemented this and made you a PR. This is almost the same as for the MeasurementGroup's configurations. The only difference here is that we do the converstion also for attributes.
I've also checked it int he following scenarios:

Scenario 1
1. Created MacroServer with sardana 2.3.2 + recent taurus and in expconf stored a snapshot.
2. Upgraded sardana to the new version (still using the same taurus) and opened expconf. This correctly showed deprecation warnings and the re-apply upgraded to the FQDN URI references.

Scenario 2
1. Created MacroServer with this version +taurus/support-3.x and played with the snapshot. No errors were showned. And as expected there were no upgrades of the snapshot names.

Just one more thing, are the changes from `src/sardana/macroserver/scan/gscan.py` still necessary?

reszelaz:
Many thanks! I removed the WIP tag and I'm merging it now.

PR #760, opened 2018-05-18
Fix set macro path

dschick:
on windows MacroPath is not working because these paths are split by ":" in the setMacroPath method of the macroManager which is a bad idea on windows with its paths like "C:\myMacros\"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /MANIFEST
 /dist
+/src/sardana.egg-info
 /build
 /setup.cfg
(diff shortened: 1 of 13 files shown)


reszelaz:
Thanks for the PR! I see that accidentally you included not related commits. Could you please create a new PR with just the commit correcting the macro path. In order to do that create a branch "fixSetMacroPath" from the tip of the upsream develop branch and then apply the changes there. Thanks a lot!

PR #761, opened 2018-05-18
Fix set controller path

dschick:


reszelaz:
Thanks for the PR! I see that accidentally you included not related commits. Could you please create a new PR with just the commit correcting the controller path. In order to do that create a branch "fixSetControllerPath" from the tip of the upsream develop branch and then apply the changes there. Thanks a lot!

PR #762, opened 2018-05-22
fix controller/macro path on windows

dschick:
on windows setMacro/ControllerPath is not working because these paths are split by ":" in the setMacro/ControllerPath method of the macroManager/poolcontrollermanager which is a bad idea on windows with its paths like "C:\myMacros"


reszelaz:
Many thanks for this contribution @dschick! However I have another proposal.

Now we accept the following format of `PoolPath` and `MacroPath` separators, example from [`PoolClass`](https://github.com/sardana-org/sardana/blob/42cb1bd6f590379649ae7c3f9cd7843b4d7ecf22/src/sardana/tango/pool/Pool.py#L1326):

> list of directories to search for controllers (path separators can be '\n' or ':')

What I suggest is to change it for:

> list of directories to search for controllers (path separators can be '\n' or character conventionally used by the OS to separate search path components, such as ':' for POSIX or ";" for Windows)

And in the code use [`os.pathsep`](https://docs.python.org/3/library/os.html#os.pathsep) for that.

The same would also apply for `PythonPath` property.

Hope you like it!

dschick:
yes sure, go ahead with this approach

PR #764, opened 2018-05-22
Stop MG in case of starting failure

rhomspuron:
If the measurement group takes more than 3s to start, there is a Corba timeout error. The MacroServer (MS) will stop, but the Pool will not do it. If the macro is the ascanct the only way to recover the system is to restart the Pool and the MS. This problem does not happen with Software synchronization, but with the other trigger elements it happens. 

To reproduce the problem, I created a copy of the DummyCounterTimerController and included on the StartOne method a time sleep:

```
 def StartOne(self, ind, value=None):
        self._log.debug('BAD StartOne(%d): entering...' % ind)
        time.sleep(3)
        self.counting_channels[ind].is_counting = True

```


teresanunez:
Hi Roberto,
                   I have tried to reproduce this. I have created a counter of the class DummyCounterTimerController where I add the time.sleep(3) in the StartOne method.
As you say, if I run ascanct with Software trigger one can not see any error. I used then as
trigger one trigger generator of the class DummyTriggerGateController and I run again a
ascanct macro. In this case I get the error:

An error occurred while running Macro 'ascanct(mot01, 1.0, 10.0, 2, 1.0, 0.0) -> 84713da4-5f34-11e8-84d0-901b0e39a972':
TRANSIENT CORBA system exception: TRANSIENT_CallTimedout

but I can continue running any other macro without any problem. I don't need to restart neither MacroServer nor Pool devices. I can even run a macro during the time in which the Pool is still
 trying to acquire from the other channels of the MG. Do I setting something wrong in order to
reproduce it?. Many thanks.

                               Regards,
                                                   Teresa

rhomspuron:
Hi Teresa, 

I reproduce the problem by running one ascanct with 10 points and run a second one without wait and abort it, after that the other scans will send the next error. In my case, the second scan will never finish if you not abort it. 
```
Door_rhoms_1 [15]: ascanct mot13 0 10 10 1
Operation will be saved in /siciliarep/tmp/rhoms/scans/20180523.dat (Spec)
Operation will be saved in /siciliarep/tmp/rhoms/scans/20180523.h5 (HDF5::NXscan)
Scan #1438 started at Thu May 24 11:41:35 2018. It will take at least 0:00:00
 #Pt No    mot13      ct13      ct14    zerod13   bad_ch1   zerod15     ct15     rhct1       dt   
Operation saved in /siciliarep/tmp/rhoms/scans/20180523.dat (Spec)
Operation saved in /siciliarep/tmp/rhoms/scans/20180523.h5 (HDF5::NXscan)
Scan #1438 ended at Thu May 24 11:41:35 2018, taking 0:00:00.013100. Dead time 100.0% (motion dead time 0.0%)
An error occurred while running Macro 'ascanct(mot13, 0.0, 10.0, 10, 0.1, 0.0) -> a5e5ab4e-5f36-11e8-b094-0030642abcf5':
Callback <bound method MeasurementGroup.valueBufferChanged of MeasurementGroup(tango://ictlael03.cells.es:10000/mntgrp/pool_rhoms_1/mg2)>(ct13) already reg. on expchan/ctctrl04/1 EG
```
The error is reported in PR 701 and is generated by the interruption of the macro. But if you run a ct it doesn't work either and shows data from the scan:
```
oor_rhoms_1 [5]: ct
Thu May 24 12:21:01 2018

An error occurred while running Macro 'ct(1.0) -> 2827445a-5f3c-11e8-a1bd-0030642abcf5':
TRANSIENT CORBA system exception: TRANSIENT_CallTimedout

Door_rhoms_1 [6]: ct   0      <nodata>     1         2      100.195      1      299.463      3         1      <nodata>
   0      <nodata>     1         2      100.195      1      299.463      3         1      <nodata>
   0      <nodata>     1         2      99.3049      1      299.463      3         1      <nodata>
   0      <nodata>     1         2      99.3049      1      299.463      3         1      <nodata>
   0      <nodata>     1         2      99.3049      1      299.463      3         1      <nodata>
   0      <nodata>     1         2      99.3049      1      299.463      3         1      <nodata>
   0      <nodata>     1         2      99.3049      1      299.463      3         1      <nodata>
Door_rhoms_1 [6]: ct
Thu May 24 12:21:05 2018

```

The problem arises when the Pool is acquiring and we do another scan without waiting for the first one to be completed and abort it. For the beamline user, this behavior is uncomfortable, since the MS is ready to run another scan and it is not true, so it generates a loop of restarting Pool and MS and try again to run the macro.  

Best,
Roberto

teresanunez:
Hi Roberto,
                   I have now played with ctrl-C also and I get the errors you show above ... but I have tried
your patch and for me the errors are still there. I think is better is somebody try in your same 
conditions.

                                       Regards,
                                                        Teresa

rhomspuron:
Hi Teresa, 

The patch does not solve the error of the PR 701. It only abort the acquisition on the Pool when there is a timeout. IMHO the Pool must abort the acquisition if there is a problem, because the MS is ready for another acquisition with the same MG, but the Pool is not.

Best,
Roberto

teresanunez:
Hi Roberto, no problem to merge it, if the purpose is only that. Should I ?.

reszelaz:
It looks ok as a workaround and we will integrate it until we fix it properly on the Pool side. The measurement groups should be self protecting itself.

PR #766, opened 2018-05-23
change epoch from float to formatted date time

dschick:
when using xrayutilities one cannot read the spec file written by Sardana because the "#D" field is not in date time format like "Sun May 17 08:09:19 2015"
This is also the way SPEC is saving the "#D" field


cpascual:
You are right. Merging.
Thanks for contributing!

PR #767, opened 2018-05-25
[WIP] Propagate keyboard interrupt exception

rhomspuron:
If you run on spock a loop and call a macro inside it, it is not possible to abort the loop with one Ctrl-C.

```
for i in range(10):
    %ct
```
If we propagate the keyboard interruption to the ipython it is possible abort the loop. In addition, it can allow to execute ipython scripts with the magic command %run and abort it.


teresanunez:
I have tested it and it allows to interrupt loops in spock. The scripst running with %run are interrupted
with ctrl-c without the need of this change.

PR #768, opened 2018-05-28
Fix meshct intervals estimation

rhomspuron:
Fix error on the interval estimation and avoid an exception at the begging of the scan:
```
....
'tbl2201:10000/expchan/xspress3_rois/21', 'tbl2201:10000/expchan/xspress3_rois/22', 'tbl2201:10000/expchan/xspress3_rois/23', 'tbl2201:10000/expchan/xspress3_rois/24', 'tbl2201:10000/expchan/xspress3_rois/25', 'tbl2201:10000/expchan/tattr_ct_ctrl/6', 'tbl2201:10000/expchan/tango_temp_ctrl/1', 'tbl2201:10000/expchan/tango_temp_ctrl/2'], 'scanfile': ['20180528.dat', '20180528.h5', '20180528.lima'], 'scandir': '/beamlines/bl22/controls/scans/20180528', 'total_scan_intervals': array([3, 3], dtype=int32)}, 'macro_id': '94707bee-627b-11e8-b83a-001999eaf6e9', 'type': 'data_desc'})>
SardanaTP.W003 DEBUG    2018-05-28 15:32:35,663 BL22/DOOR/01: Details
  File "/homelocal/sicilia/lib/python/site-packages/sardana/sardanabase.py", line 130, in fire_event
    listeners=listeners)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/sardanaevent.py", line 110, in fire_event
  File "/homelocal/sicilia/lib/python/site-packages/sardana/sardanaevent.py", line 132, in _fire_event
    real_listener(self, event_type, event_value)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/tango/macroserver/Door.py", line 275, in on_door_changed
    event_value = codec.encode(('', value))
  File "/homelocal/sicilia/lib/python/site-packages/taurus/core/util/codecs.py", line 320, in encode
    return format, json.dumps(data[1], *args, **kwargs)
  File "/usr/lib64/python2.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib64/python2.7/json/encoder.py", line 201, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python2.7/json/encoder.py", line 264, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python2.7/json/encoder.py", line 178, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: array([3, 3], dtype=int32) is not JSON serializable
```


reszelaz:
Many thanks for this PR! However there are some fundamental things that IMO we need to clarify earlier and they also affect the solution that you propose.

The previous generation of continuous scans e.g. `ascanc`, `meshc`, `d2scanc` and the `CSScan` GSF class introduced the concept of waypoints. The idea is that the new generation of continuous scans e.g. `ascanct`, `meshct`, `d2scanct` fully adopt this concept - in case of these scans the waypoint generation still needs to be improved, for example, by allowing to pass the synchronization description to each waypoint yield.

The previous generation of continuous scans interpreted the intervals concept (as returned by `getInteralEstimation`) as strictly related to the motion intervals and not the acquisition intervals. This is still in accordance to the intervals concept used by the step scans. And I think it is correct.

So, if we interpret that a single motor continuous scan e.g. `ascanc` or `ascanct` has two waypoints (start and final position)? then the number of intervals should be equal to the number of waypoints - 1. Now in `ascnac` it is equal to number of waypoints which IMHO is a bug. In case of `ascanct` it is equal to the number of intervals passed by the user in order to determine the acquisition points which IMHO is also a bug. The `meshc` continuous scan has IMHO correct number of itnervals (`m2_nr_interv` + 1 which is exactly the number of waypoints - 1).

For now I propose to implement the `meshct.getIntervalEstimation`  the same way it is done for `meshc`. This way we will avoid the error that you report. If you agree, please could your modify the code accordingly. Then I would carefully verify the above assumptions and correct the supposed bugs in other continuous scan after the Jul18 release.

rhomspuron:
Thanks for the review! I agree with the concepts, I will do the change.

PR #773, opened 2018-05-30
SEP18 - Extend acquisition and synchronization concepts for SEP2 needs

reszelaz:
When brainstorming about integration of 1D and 2D detectors channels (SEP2) we found that some acquisition and synchronization concepts are missing. This [SEP18](https://github.com/reszelaz/sardana/blob/sep18/doc/source/sep/SEP18.md) introduces them.

This is just a draft. Your comments are more than welcome!


rhomspuron:
1. Document well that:

- [ ] Software(Trigger|Gate) are synonyms to Internal (Trigger|Gate).
- [ ] Hardware(Trigger|Gate) are synonyms to External(Trigger|Gate).

2. Extend AcqSynch with two new options:

- [ ] SoftwareStart (which means internal start)
- [ ] HardwareStart (which means external start)

3. Extend AcqSynchType with one new option (supported from expconf):

- [ ] Start

4. Allow different types of preparation of channels:

- [ ] Per measurement preparation with repetitions=n e.g. Prepare(One|All) or a controller parameter
- [ ] Per acquisition preparation with repetitions=1 e.g. Load(One|All)

5. Modify acquisition actions (and synchronization action if necessary) 

- [ ] They support the new concepts added in points 2 and 4.

6. Extend GSF 

- [ ] Step mode measurement preparation (repetitions=n) if possible i.e. scan macro knows beforehand the number of points.

reszelaz:
I just completed the SEP document (you can find the link in the initial comment). Now it contains quite detailed implementation roadmap... I would like to know your opinions about it, especially about the [controllers API](https://github.com/reszelaz/sardana/blob/sep18/doc/source/sep/SEP18.md#controllers). Answering on the following question will help in selecting between the proposed options. But let's start with some background...

In Sardana world, C/T, 1D and 2D experimental channels are integrated with the so called "Loadable" controllers, which implements `LoadOne`, etc. methods. Currently these methods have two arguments: integration time (or monitor count) and number of repetitions (triggers/gates).

We see a need to add more configuration parameters: latency time and number of starts.

In order to define the best interface to these controllers I have one specific question about all the acquisition hardware that you use/have/know.

"Is it enough to configure the hardware only once before executing multiple times the start?"

An example would be a step scan of 5 points with 0.1 s integration time: "Is it enough to configure only once the hardware with 0.1 s integration time (and eventually prepare it for 5 starts) and then start the acquisition 5 times?"

At ALBA we already make a research on our hardware and it would be nice to hear your experience as well.

Finally, at ALBA we already started working on the implementation, of course we wait for the decision on the controller API. As always you are more then welcome if you want to help. To coordinate the work I have prepared the [SEP18 project board](https://github.com/sardana-org/sardana/projects/4).

reszelaz:
I just wanted to let you know that the implementation of SEP18 is advancing really fast - you can consult the project board. There are still 3 major parts to be done:
* finish refactoring of the measurement configuration on the server side
* implement software start action
* documentation

The decision on the controller's API is still not taken. At Alba we cross-checked the following hardware Keithley, Keysight, ALBA Em# and for them option 2/3 would be enough.

I move this SEP to CANDIDATE state.

reszelaz:
Just an update about the SEP18 status.

## In summary...
It looks almost ready to be voted. There are just 6 small comments in the code review and whenever these will be solved we will call for voting for acceptance.

## In detail...
Finally, on the last two Sardana follow-up meetings we (ALBA, DESY, MAXIV and SOLARIS) agreed to choose the Option 1 controller's API (`SEP18->Implementation->Controllers`) which is more flexible and conservative but at the same time adds more complexity when developing a new timerable controller. Alternative options are described in the SEP18 as appendixes. **You are still on time to rise your hand and we can rediscuss it.**

There are some TODOs which I think it is safe and more convenient to leave for the future (SEP18 blocks SEP2).

These are the TODOs:
* **Stop channels using software trigger/gate on software synchronizer's `end` event.** This is necessary since we introduced the preparation of channels, and they may be aware in how many software trigger/gates they will participates. Sardana does not guarantee to acquire on all triggers/gates, for example if the previous acquisition is still in progress and a new trigger/gate should be issued, and some channels may be waiting for them - this is the case of Lima.
* **Documentation**:
    * measurement group additional API
    * how to write timerable controllers
* **Remove `StartMultiple` Tango command** and its downstream core implementation

reszelaz:
The very last code review comments were implemented. There is nothing else but the agreed TODOs pending.

Since on the last two follow-up meetings we all agreed on the contents and implementation of this SEP I won't call the official voting which would be just a formality. On Monday we will proceed to:
* integrate this PR
* update the CHANGELOG
* create issues for the agreed TODOs

Feel free to comment.

PR #774, opened 2018-06-01
Correcting lsgh if macro several times in same hook position

teresanunez:
The macro lsgh was listing each hook macro only once, even if it was added several times in the same
positon, but the macro was called so many times as it was added. This PR makes consistent the lsgh with what is happening.


reszelaz:
I 100 % agree with the change. If the hook will be executed multiple times `lsgh` should also list it multiple times. The change is trivial so I merge it directly without making functional tests. Thanks a lot!

Issue #776, opened 2018-06-04
newfile macro


dschick:
I am actually missing a `newfile ` macro which set a new data file, resets the scanID to 1 and also writes a proper scanFileHeader for SPEC files.

As some analysis software requires a proper SPEC file header containing the motor and counter names such as here:

```
#F 2017-08-Schick-02
#E 1502445176
#D Fri Aug 11 11:52:56 2017
#C fourc  User = specadm

#O0 TwoTheta  Theta  Chi  Phi  mono  threl  xs  ys
#O1 zs  Energy  DetHeight  CCDH  MagnetHeight  FYHeight  MagnetRot  Current
#O2 TwoThetaR  Volt  ExitSlit  Lake  
#o0 tth th chi phi mono threl xs ys
#o1 zs en deth ccdh mh fyh mr cur
#o2 tthr volt exs lake 

#J0 Seconds  Monitor  Detector  Drain  Energy  Ringcurr  Tsample  Tcryo
#J1 Theta  TwoTheta  Gap  Shift  GapD  ShiftD  Bandwidth  PSDPos
#j0 sec mon det drain e ri tsample tcryo
#j1 theta twoth gap shift gapD shiftD bw psdpos
```

I guess it should not be to hard to do the first to things, namely changing the ScanFile and the ScanID from within a macro accessing the environment variables.
Where would one write a scan file header? In the init method of the according recorder class?

Best

Daniel

reszelaz:
Thanks for this suggestion! Actually it sounds familiar to me.. If I remember correctly we all agreed on the Sardana Workshop in 2015 that it would be nice to have the`newfile` macro in Sardana. The only thing is that we did not have time to implement it yet..

At that time the proposal was to use the repeat parameter in the macro to be able to set multiple files (set `ScanFile` environment variable in a more user friendly way). I assume that at that time we were not thinking about resetting the `ScanID`, but I think it make sense. Or at least we could give another macro parameter (bool) if one wants to reset it or not, with a default value.

Regarding the scan file header, I think we should go for it (see this recommendation: https://github.com/sardana-org/sardana/issues/735#issuecomment-382312675).

We just need to agree on some things:
* Is `#F` an absolute path: \<ScanDir\> + "/" \<Spec file name extracted from ScanFile\> ?
[...]

Issue #771, opened 2018-05-29
write comments to scan header


dschick:
I would like to add some comments to the scan header of my data files.
This would be similar to the prescan-snapshot functionality.
I am thinking of putting some env parameters such as strings (sample name, etc) into the scan header.
I guess I could solve that problem using the available functionality with the prescan snapshot, by storing my comments in a TangoDeviceServer, right?

Are there any other implementations possible which are general enough to work with any of the DataRecorders?

Daniel

rhomspuron:
Hi Daniel, 

The snapshot does not save strings on SPEC, it only support numbers, because to save the snapshots, the recorder users the labels for the motor position "#O" and "#P". The supported types are:
```
 supported_dtypes = ('float32', 'float64', 'int8',
                        'int16', 'int32', 'int64', 'uint8',
                        'uint16', 'uint32', 'uint64') 
```

The same supported types are used for the Nexus recorder.

If you want to add commend on the file you should do it like the example macro   ascan_with_addcustomdata on:

[lhttps://github.com/sardana-org/sardana/blob/develop/src/sardana/macroserver/macros/examples/scans.py](url)

Regads,
Roberto

dschick:
Hi Roberto,

great, the addCustomData method of scan, is exectly what I was looking for.

How can I access the `_gScan._data_handler` object from within a hook, i.e.
at the pre-scan hook-place?
Do I need to use to call the parent of the hook?

Best

Daniel

Roberto Javier Homs Puron <notifications@github.com> schrieb am Di., 29.
Mai 2018 um 16:51 Uhr:

rhomspuron:
Hi Daniel, 

The `_gScan` is an object of the scans (aNscan, aNscanct, mesh, meshct, etc). You can access to it from the hooks.

```
class test_customdata(Macro):
    def pre_scan(self):
        self.info('add comment')
        dh = self.scan._gScan._data_handler
        # at this point the entry name is not yet set, so we give it explicitly
        # (otherwise it would default to "entry")
        dh.addCustomData('Hello world1', 'dummyChar1')

    def run(self):
        self.scan, _ = self.createMacro('ascan mot13 0 1 10 0.1')
        self.scan.hooks = [(self.pre_scan, ['pre-scan'])]
        self.info(self.scan.hooks)
        self.runMacro(self.scan)

```

Spec file:
```
#S 1874 ascan mot13 0.0 1.0 10 0.1
#U sicilia
#D 1527664730.0
[...]

PR #777, opened 2018-06-08
Newfile macro

dschick:


dschick:
one could think of doing a N-1 in the macro - so if one sets ScanID to 1 it will be internally set to 0 because for the next scan it will be incremented by one.

dschick:
what is the problem with the 

> Travis CI build failed

?

cmft:
Hi @dschick ,
You have flake8 errors:
```
--------------------------------------------------------------------------------
src/sardana/macroserver/macros/standard.py:852:72: W291 trailing whitespace
    """This macro sets a new ScanFile and ScanID values into the env.
    The ScanID should be set to the value N of the last executed ScanID 
    in order to continue with a ScanID N+1 for the next scan."""
                                                                       ^
src/sardana/macroserver/macros/standard.py:863:1: W293 blank line contains whitespace
    
^
src/sardana/macroserver/macros/standard.py:864:42: W291 trailing whitespace
    def run(self, ScanID, ScanFile_list):        
                                         ^
src/sardana/macroserver/macros/standard.py:867:1: W293 blank line contains whitespace
        
^
src/sardana/macroserver/macros/standard.py:872:1: W293 blank line contains whitespace
        ^
src/sardana/macroserver/macros/standard.py:872:1: W391 blank line at end of file
        ^
--------------------------------------------------------------------------------
```

reszelaz:
@dschick I see that you faced problems with the flake8 checks. I have just complemented the [CONTRIBUTING.md](https://github.com/sardana-org/sardana/blob/develop/CONTRIBUTING.md) guide with some more details about the flake8 check.

I see that you managed to fix all the errors - great! BTW were you able to navigate to these errors in the travis-ci PR build history?

If in the future you find problems, don't hesitate to ask for help!

dschick:
@reszelaz , thanks for updating the docs!
It was actually not that easy for me to find the errors, but @cmft was so kind to point them out in the first place.
I was only able to find the travis-ci errors when I searched the raw-log from the flake8 check:
https://api.travis-ci.org/v3/job/390683877/log.txt

Is there a easier way to find the errors?

I will try to run the flake8 check locally in the future, in order to avoid the unnecessary commits...

reszelaz:
Great! Running the check locally is a good idea. But I have to admit that I don't always remember to do it. The good thing is that there is always the travis guardian:)

> Is there a easier way to find the errors?

I navigate there in the following way:
1. In the Converstation tab of the PR I click on the red cross next to the commit hash.
2. In the pop-up window I click on "Details" next to the travis part of the continuous integration.
3. I click on the failed job with with TEST="flake8".
4. I scroll to the end of the "Job log" to see the errors.

dschick:
continuing the discussion of #776 I was wondering if the actual file header should be part of the `newfile` macro or if this should be part of the `init` method of each recorder when a new data file is initialized?

The comment issue is discussed in #771 and I guess there is some convergence regarding the format of the file header, see #776 

As I am currently having a beamtime using SPEC I just noticed, that the `newfile` macro here either accepts the `scanID` and `scanFile` name as arguments or switches to an interactive mode asking for both inputs, if no arguments are given.
Is there a general rule for Sardana if and how to realize such interactive dialogues?
As for me currently under windows the qt-inputs are not working, I would have to realize it with `raw_input`.

teresanunez:
@dschick, there is a way of asking parameters interactively, you can have a look to some examples in the file hkl.py of the macroserver/macros directory. Search for self.input.

reszelaz:
> I was wondering if the actual file header should be part of the newfile macro or if this should be part of the init method of each recorder when a new data file is initialized?

I think that the recorder constructor is fine to create the file header the first time it writes to the file.

> Is there a general rule for Sardana if and how to realize such interactive dialogues?
As for me currently under windows the qt-inputs are not working, I would have to realize it with raw_input.

Yes, there are some issues with the interactive macros. Hopefully we will solve them soon..

From what you describe how Spec implements the `newfile` macro, I think that the #285 implementation would be helpful here. Menwhile, I suggest to use some predefined default values (something that the user would never want to set) for these parameters. So the macro code could recognize them and show the `self.input` as @teresanunez suggest.

src/sardana/macroserver/macros/standard.py
```diff
@@ -845,3 +845,27 @@ def run(self, nr):
...
+
+
+class newfile(Macro):
+    """This macro sets a new ScanFile and ScanID values into the env.
+    The ScanID should be set to the value N of the last executed ScanID
+    in order to continue with a ScanID N+1 for the next scan."""
+
+    env = ('ScanFile', 'ScanID')
+
+    param_def = [
```
reszelaz: I would try to follow the spec command parameters order i.e. `ScanFile_list` first and then `ScanID`. In order to make the macros execution as user friendy as possible this would require extending the _spock syntax_. More details in #781.

src/sardana/macroserver/macros/standard.py
```diff
@@ -845,3 +845,27 @@ def run(self, nr):
...
+        ['ScanID', Type.Integer, 1, 'Scan ID'],
```
reszelaz: Do you know how spec deals the default value? If ones executes `newfile myfile.dat` does it reset the ScanID?

reszelaz:
@dschick I've reviewed this PR. There are some things to clarify - see my inline comments.
Finally, I have one more comment regarding the modifications fo the `ScanDir`. Let's see if you like my idea (it assumes that we agree on integrating #781).

One could run the macro with the file name and this would just change the `ScanFile` e.g.
```
Door> newfile myfile.dat
```
But we could also allow to set the full path and this would change the `ScanDir` and the `ScanFile` (here the macro would need to do the splitting) e.g. 

```
Door> newfile /tmp/myfile.dat
```

If ones runs the macro with multiple files, we need to take care about these questionable possibilities:

```
Door> newfile [/tmp/myfile.dat /tmp/myfile.h5]
```
I propose to set `ScanDir` to `/tmp`.

```
Door> newfile [/tmp/myfile.dat myfile.h5]
```
I propose to set `ScanDir` to `/tmp`.

```
Door> newfile [/tmp/myfile.dat /home/user/myfile.h5]
```
I propose to raise an exception saying that the directories are different.

src/sardana/macroserver/macros/standard.py
dschick: I agree on that.

src/sardana/macroserver/macros/standard.py
dschick: I just tried it in SPEC and it sets the ScanID by default to 0 so the next scan starts with 1.
So this has to be implemented.

src/sardana/macroserver/macros/standard.py
```diff
@@ -845,3 +845,27 @@ def run(self, nr):
...
                 progress = ((i + 1) / float(nr)) * 100
                 yield progress
+
+
+class newfile(Macro):
+    """This macro sets a new ScanFile and ScanID values into the env.
+    The ScanID should be set to the value N of the last executed ScanID
+    in order to continue with a ScanID N+1 for the next scan."""
+
+    env = ('ScanFile', 'ScanID')
```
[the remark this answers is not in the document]
dschick: okay, so I have to remove that.

dschick:
okay, I agree on that setting the ScanDir and ScanID is a good idea.
I will try to implement it together with your earlier suggestions.
But as you said, I would wait until the new spock syntax is approved.

dschick:
In SPEC the newfile macro handles also the three cases listed below which should be combined with the suggestions of @reszelaz :

1. `Door> newfile /tmp/myfile.dat`
data file is created in the given absolute path
2.  `Door> newfile ./myDataFolder/myfile.dat`
data file is created in the sub-folder myDataFolder relative to the working directory of spock
3. `Door> newfile myfile.dat`
data file is created in the working directory of spock

reszelaz:
Thanks for continuing with this issue!

Please remember that the spock and the MacroServer may run on different machines with different file systems. So the working directory of spock may not exist on the MacroServer's host where actually the file will be written.

dschick:
ah, okay that's a good point!
So would you suggest skipping the part with the pwd? or should one check if spock and macroserver run on the same machine for this feature?

I am just used to run SPEC in some kind of experiment folder for each beamtime. Thus all files, such as the SPEC files and local macro files, have been all saved in this folder or subfolders of it.

reszelaz:
> So would you suggest skipping the part with the pwd? or should one check if spock and macroserver run on the same machine for this feature?

I think it would be good enough to accept the full paths (on the server file system) or just the file names. We could skip the relative paths.

reszelaz:
Yes. I see it like this.

dschick:
what I do not like about this option is, that you can run the newfile macro, but when no ScanDir is applied, storing the Data will not work. So in case no path is present in the newfile input parameters, I will check, if the ScanDir env parameters is set. Otherwise, I will return a message, stating that no ScanDir is set.

reszelaz:
This sounds even better!

dschick:
thanks @reszelaz for the update.

I just had to think a bit more about the newfile macro, because:

as I am integrating a camera to my setup I would like to save the camera images in a folder named like the SPEC file in Sardana. Therefor I would like to send a command to the camera controller sending the new SPEC file name when using the newfile macro.

as discussed in #979 this might be generally a bad design idea.

However, would it be possible to hook a custom macro to the newfile macro which would do the job of setting the data path for my camera?

Best Daniel

reszelaz:
Saving configuration is going to be covered by SEP2 as well.

Meanwhile I think that adding a hook to the `newfile` macro is fine. Any other opinions? Hook place could be called `post-newfile`.

BTW, when taking a quick look at the PR changes I think that it is quite ready to be reviewed and integrated. Should we proceed with the review?

dschick:
okay, I will try to update the newfile macro with the hook place and polish it a bit.
After that it should be fine for review.

Thanks for your opinion

dschick:
I just did some minor changes and added the post-newfile hook.
So indeed one could test and merge the macro

dschick:
Thank you for reviewing it. I am pretty fine with your changes.

Am Mi., 6. Feb. 2019, 19:55 hat reszelaz <notifications@github.com>
geschrieben:

PR #779, opened 2018-06-11
Fix nexus path windows

dschick:
as pointed out in #720 in `h5storage.py` the `os.path.join` has to be replaced to always use "/" as path separators for NeXuS paths.
In order to do so, the `posixpath` module is used which has the exact same interaface as the `os.path` module but uses always UNIX-style path-separators, see https://docs.python.org/2/library/os.path.html

### DIFF
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -33,6 +33,7 @@ __all__ = ["NXscanH5_FileRecorder"]
 __docformat__ = 'restructuredtext'
 
 import os
+import posixpath
 from datetime import datetime
 import numpy
 import h5py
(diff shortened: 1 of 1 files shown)


cpascual:
Merging. Thanks a lot for contributing!

dschick:
hej @cpascual, is this PR already merged? You have written so, but github does not say so ...

cpascual:
> hej @cpascual, is this PR already merged? You have written so, but github does not say so ...

dammit! You are right.... I must have clicked on close instead of merge...

Now GH is not letting me reopen this PR because you deleted your fixNexusPathWindows  branch...
Could you check if you can undelete or re-push the branch from your local repo?

If not, we can redo the commits [from the diffs](https://github.com/sardana-org/sardana/pull/779/files)

dschick:
just reopened it. I think I closed it because I thought it was merged too :)

cpascual:
Thanks @dschick , and sorry for the inconveniences!

PR #780, opened 2018-06-13
Acquisition hooks

jordiandreu:
Add _pre-acq_ and _post-acq_ hooks to aNscanct macro.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -2300,10 +2300,6 @@ class CTScan(CScan, CAcquisition):
                 self.macro.debug(
                     "Moving to waypoint position: %s" % repr(final_pos))
                 motion.move(final_pos)
-            finally:
-                # To ensure it is executed.
-                for hook in waypoint.get('post-acq-hooks', []):
-                    hook()
                 # wait extra 15 s to for the acquisition to finish
                 # if it does not finish, abort the measurement group
                 # (this could be due to missed hardware triggers or
@@ -2322,6 +2318,9 @@ class CTScan(CScan, CAcquisition):
                     self.debug(msg)
                     measurement_group.Stop()
                     raise ScanException("acquisition timeout reached")
+
+            for hook in waypoint.get('post-acq-hooks', []):
+                hook()
 
             if macro.isStopped():
                 self.on_waypoints_end()
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -2301,6 +2301,7 @@ class CTScan(CScan, CAcquisition):
                     "Moving to waypoint position: %s" % repr(final_pos))
                 motion.move(final_pos)
             except Exception as e:
+                measurement_group.waitFinish(timeout=0, id=mg_id)
                 msg = "Motion did not start properly.\n{0}".format(e)
                 self.debug(msg)
                 raise ScanException("move to final position failed")
@@ -2312,7 +2313,6 @@ class CTScan(CScan, CAcquisition):
                 # TODO: allow parametrizing timeout
                 timeout = 15
                 measurement_group.waitFinish(timeout=timeout, id=mg_id)
-            finally:
                 # TODO: For Taurus 4 / Taurus 3 compatibility
                 if hasattr(measurement_group, "stateObj"):
                     state = measurement_group.stateObj.read().rvalue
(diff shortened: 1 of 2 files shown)


jordiandreu:
Thank @teresanunez ,I have fixed the typo.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -2295,6 +2301,9 @@ def _go_through_waypoints(self):
                     "Moving to waypoint position: %s" % repr(final_pos))
                 motion.move(final_pos)
             finally:
+                # To ensure it is executed.
+                for hook in waypoint.get('post-acq-hooks', []):
```
reszelaz: Since at this moment the measurement group may be still acquiring (or extracting the data) I would put the hook execution after the `waitFinish`.

There we have two options:
* **Option 1**: execute them always i.e. even if there was a timeout
* **Option 2**: execute them only if there was no timeout

If we decide for the Option 1 I suggest to do it after the measurement group abort.

What are your opinions about this?

src/sardana/macroserver/scan/gscan.py
```diff
@@ -2277,7 +2277,13 @@ def _go_through_waypoints(self):
             self.data.initial_data = initial_data
 
             if hasattr(macro, 'getHooks'):
-                for hook in macro.getHooks('pre-start'):
+                pre_acq_hooks = macro.getHooks('pre-start')
+                if len(pre_acq_hooks) > 0:
+                    self.macro.warning("pre-start hook place is deprecated,"
```
reszelaz: Here you are using the macro logging stream. This will appear in spock. I don't mind. For sure this will be more efficient to get rid of the deprecated calls. Just a comment ...

src/sardana/macroserver/scan/gscan.py
reszelaz: If there are no other opinions I will move it there and integrate the PR.

teresanunez: I agree

reszelaz: Now I see that there is one more point. In case the move gave an error, do we also want to execute the `post-acq` hook? Now it is like this.

rhomspuron: Good point, should not do it.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -2314,6 +2325,9 @@ def _go_through_waypoints(self):
                     measurement_group.Stop()
                     raise ScanException("acquisition timeout reached")
 
+            for hook in waypoint.get('post-acq-hooks', []):
+                hook()
+
```
jordiandreu: Execute hooks after the try-except block ended successfully.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -2294,14 +2300,19 @@ def _go_through_waypoints(self):
                 self.macro.debug(
                     "Moving to waypoint position: %s" % repr(final_pos))
                 motion.move(final_pos)
-            finally:
+            except Exception as e:
+                msg = "Motion did not start properly.\n{0}".format(e)
+                self.debug(msg)
+                raise ScanException("move to final position failed")
+            else:
```
jordiandreu: Only wait for the measurement group to end (including the timeout) only if no exception was raised.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -2294,14 +2300,19 @@ def _go_through_waypoints(self):
...
                 # wait extra 15 s to for the acquisition to finish
                 # if it does not finish, abort the measurement group
                 # (this could be due to missed hardware triggers or
                 # positioning problems)
                 # TODO: allow parametrizing timeout
                 timeout = 15
                 measurement_group.waitFinish(timeout=timeout, id=mg_id)
```
jordiandreu: Only wait for the measurement group to end (including the timeout) only if no exception was raised.

src/sardana/macroserver/scan/gscan.py
jordiandreu: We have moved the hooks execution after the try-except-finally block, so they will be executed only if no exception occurred.

rhomspuron:
Tested `ascanct mot29 0 10 20 1` the three cases:

1) Motor finish before the MG: Add on the post-acq general hook the next macro:
```python
class change_vel(Macro):
    def run(self, *args):
        m = self.getMotor('mot29')
        m.write_attribute('velocity', 10000)
        self.info('change vel')
```

2) Motor fails on start the movement: Add on the post-acq  general hook `set_lim mot29 5 10`
3) Run macro without errors.

reszelaz:
Great! Thanks to both! Merging..

PR #672, opened 2017-12-21
645 issue - Allow string ParamRepeats on load Sequencer plain text

droldan:
This pull request allows to use a paramrepeats macros with a new/old interface syntax when a plaintext file is loaded in a sequencer.
Also an Official sardana ParamParser is used to extract the params.
The ParamFactory has been improved to allow create all kind of parameters. (SingleParamNode, RepeatParamNode) passing macroInfo dictionary or a simple values (list or single value)

Issue #645, opened 2017-11-10
Allow string ParamRepeats on load Sequencer plain text. 


droldan:
There are problems on try to load a plain text sequence that contains strings ParamRepeat on Sequencer.

reszelaz:
@droldan please take a look on how was it solved in Spock. The parameter parser https://github.com/sardana-org/sardana/blob/01b154244c97e09b316be39fbf2e7a15d149e4f3/src/sardana/spock/parser.py#L61 together with https://github.com/sardana-org/sardana/blob/01b154244c97e09b316be39fbf2e7a15d149e4f3/src/sardana/taurus/core/tango/sardana/macro.py#L1188 should be very close to what you need.

reszelaz:
Solved in #781

PR #781, opened 2018-06-13
Extend spock syntax

### LINKED ISSUE #438: spock ignores extra parameters
This is probably known, but IMHO it should be fixed.

Consider, e.g. the example macro `pt0`, which defines no parameters.

spock does not complain if one executes `pt0 foo bar 1 2 3 4`

Similarly, pt1 (which accepts a single parameter) would ignore all but the first parameter passed to it.

IMHO, the parser should complain and refuse to execute the macro.

reszelaz:
This PR is a proof-of-concept only and I would like to hear your opinion about it before asking for its integration.
A little bit of background..
_Spock syntax_ is based on space separated list of parameter values. Not all macros are allowed to be used with the spock syntax. For example, the following macros would not be allowed:
1. macros defining more than one repeat parameter
2. macros defining repeat parameter which is not at the end of the parameters definition
3. macros defining nested repeat parameters

In #405 we introduced the _Advanced spock syntax_. This one introduces the use of square brackets to group the repeat parameters and its repetitions. There are no limitations anymore if the advance spock syntax is in use.

The newfile macro proposed in #777 defines two parameters in the following order: `ScanId` and `ScanFile_list`. The `ScanFile_list` is a repeat parameter. In spec, the opposite order of parameters is defined, first the `ScanFile` and then the optional `ScanId` (from [Spec Manual](https://certif.com/downloads/css_docs/spec_man.pdf) page 6). I guess that the order of parameters in the current proposal for the `newfile` macro comes from the previously mentioned limitation of the spock syntax. As it is now, in order to use the `newfile` macro with the spock syntax one needs to place the repeat parameter at the end. This forces the users to use for example:

```
# to switch to myfile.dat and reset the ScanId
Door> newfile 1 myfile.dat
# to switch to two files myfile.dat and myfile.h5 and reset the ScanId
Door> newfile 1 myfile.dat myfile.h5
# to switch to myfile.dat and continue with the ScanId
Door> newfile <the highest scan id in the file + 1> myfile.dat
# to switch to myfile.dat and myfile.h5 continue with the ScanId
Door> newfile <the highest scan id in both files + 1> myfile.dat myfile.h5
```

If we inverse the order of parameters definition we would force the users to use the advanced spock syntax:

```
# to switch to myfile.dat and reset the ScanId
Door> newfile [myfile.dat]
# to switch to two files myfile.dat and myfile.h5 and reset the ScanId
Door> newfile [myfile.dat myfile.h5]
# to switch to myfile.dat and continue with the ScanId
Door> newfile [myfile.dat] <the highest scan id in the file + 1>
# to switch to myfile.dat and myfile.h5 continue with the ScanId
Door> newfile [myfile.dat myfile.h5] <the highest scan id in both files + 1>
```

This PR relax the spock syntax limitations 1 and 2 and extends the spock syntax possibilities. The parsing limitation is still there - without grouping repeat parameters and its repetitions (square brackets) it is not possible to parse correctly all the cases. The implementation simply assumes that the parameter values passed by the user are a single repetition of the repeat parameter. This would allow the following using the spock syntax:

```
# to switch to myfile.dat and reset the ScanId
Door> newfile myfile.dat
# to switch to myfile.dat and continue with the ScanId
Door> newfile myfile.dat <the highest scan id in the file + 1>
```

For passing more than one repetitions one would still need to use the advanced spock syntax.

```
# to switch to two files myfile.dat and myfile.h5 and reset the ScanId
Door> newfile [myfile.dat myfile.h5]
# to switch to myfile.dat and myfile.h5 continue with the ScanId
Door> newfile [myfile.dat myfile.h5] <the highest scan id in both files + 1>
```

IMO this PR is beneficial for `newfile` macro. Thanks to it we could follow the spec command syntax and I think that the users would appreciate it. And I imagine that some other macros could also benefit from this...

If you agree on introducing this feauture I will propose some automatic tests to validate this PR in more advanced scenarios.

Fixes #438.


teresanunez:
Hi Zibi,I think that the extension you propose is important to have. You can propose the tests for 
validating it, or tell me if I should make any extra one. Thanks.

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1301,45 +1301,60 @@ def createMacroNode(macro_name, params_def, macro_params):
...
-                    # repetitions e.g. min=0
+                    # Add a new repeat node. This is needed when the raw
+                    # values fill more repeat nodes that the minimum number
+                    # of repetitions e.g. min=0.
                     if repeat_node is None:
                         repeat_node = param_node.addRepeat()
                     member_node = repeat_node.child(mem)
                     if isinstance(member_node, RepeatParamNode):
-                        msg = ("Nested repeat parameters are not allowed")
+                        msg = ("nested repeat parameters are not allowed "
```
cpascual: I would suggest the following message instead:
`"Nested parameters not allowed in (old) spock syntax. Use "bracket" syntax instead."`

src/sardana/taurus/core/tango/sardana/macro.py
```diff
@@ -1301,45 +1301,60 @@ def createMacroNode(macro_name, params_def, macro_params):
...
+                if repeat_node is None:
+                    repeat_node = param_node.addRepeat()
+                if len(repeat_node.children()) > 1:
+                    msg = ("repeat parameter with more than one member must "
+                           "be defined at the end when intended to "
+                           "be used with spock syntax")
+                    raise Exception(msg)
+                member_node = repeat_node.child(0)
+                if isinstance(member_node, RepeatParamNode):
+                    msg = ("nested repeat parameters are not allowed when "
```
cpascual: I would suggest the following message instead:
`"Nested repeat parameters not allowed in (old) spock syntax. Use "bracket" syntax instead."`

cpascual:
The proposed enhancement looks ok to me as well. If it does not break anything else, go ahead.

dschick:
I agree on it as well. Besides the newfile macro also the regscan macro has a similar issue with putting the integration time not at the end, as in SPEC and for all other scan macros, but before the start, end, nbOfIntervalls parameters.

Go ahead!

teresanunez:
Hi @reszelaz , I have updated the file macros/examples/parameters.py for the case this is accepted. I send a PR to your fork.

reszelaz:
The initial implementation here was not using the correct approach. It was necessaryt o add more inteligence to the [ParamParser](https://github.com/sardana-org/sardana/blob/a1104ece5abf1a5abd741861b07c2160398c73c7/src/sardana/spock/parser.py#L61) in order to correctly treat all cases. So I've added the macro parameters definition to the `ParamParser` and refactored the parsing logic - now it is much complicated but supports everything what was supported up to now + the changed proposed in the initial comment of this PR.

There is still some work to be done here:
* More unit tests are necessary. I have added some cases of [macro parameters examples](https://sardana-controls.org/devel/examples/macro_parameter_examples.html) but we need to add all of them. Also the cases of the previous `ParamParser`, commented out in 7e7f724, needs their unit tests.
* Finally we need to use this parser everywhere in the code where we parse macro parameters passed as strings:
  * Macro API calls e.g. `runMacro`, `execMacro`, etc
  * sequence loading from plain text file in the sequencer - this will implement #645 (#672 won't be necessary anymore) @droldan can you take a look on this PR and check if it satisfies your needs?
  * `sardana.taurus.core.tango.sardana.macroserver.BaseDoor._createMacroXmlFromStr`
  * any other places that I don't recall now..

Also take a note that this feature fixes #438.

If you want to help with this development you are more than welcome:)

reszelaz:
@droldan had implemented all required tests. We have fixed some bugs in the parser revealed by those tests.
I have also implemented parser's move from spock to util as proposed in #908.
Now I will proceed to use the parser in all the Sardana internals but the sequencer and hopefully this PR will be ready for integration soon.

src/sardana/tango/macroserver/test/base.py
```diff
@@ -44,7 +44,7 @@ class BaseMacroServerTestCase(object):
     door_name = getattr(sardanacustomsettings, 'UNITTEST_DOOR_NAME',
                         "door/demo1/1")
 
-    def setUp(self, pool_name):
+    def setUp(self, pool_name=None, properties=None):
```
reszelaz: Is it difficult to merget both:` pool_name` and `properties` into one properties only? If yes, I think it is worth to do it. Also, all the tests already using `pool_name` shoudl be adapted.

src/sardana/macroserver/test/res/macros/testmacros.py
```diff
@@ -0,0 +1,148 @@
...
+
+        expect_params = ([999.], True)
+        macro = self.execMacro("testParamsFirstRepeat", 999, True)
+        self.runMacro(macro)
+        result = macro.data
+        msg = MSG_TEMP % (result, expect_params)
+        assert expect_params == result, msg
+
+
+class testParamMacros1(Macro):
```
reszelaz: I like the idea of analogy with the parameters examples macros. But we need to keep the macro name and its docsting consitent. For this need, I propose to rename the macro to `pt5_base` and in the docstring use reference to the `pt5` example explaining that we use base python types instead of pool elements e.g. motors.

src/sardana/macroserver/test/res/macros/testmacros.py
```diff
@@ -0,0 +1,148 @@
...
+    ]
+
+    def run(self, *args, **kwargs):
+        self.data = args
+
+
+class pt10_base(Macro):
+    """Macro with a list of numbers followed by a number parameter.
+    Usages from Spock, ex.:
+    pt10_base 99 [1 3]
```
reszelaz: I think this is not correct, the first parameters is repeat parameter and not the second one.

src/sardana/macroserver/test/res/macros/testmacros.py
```diff
@@ -0,0 +1,148 @@
+from sardana.macroserver.macro import Type, Macro
+
+MSG_TEMP = "Parsing failed (result: %r; expected: %r)"
```
reszelaz: I would say that decoding failed. Or better "Parsing or decoding failed..."

src/sardana/macroserver/test/res/macros/testmacros.py
```diff
@@ -0,0 +1,148 @@
...
+        params = (91, True)
+        expect_params = ([91], True)
+        macro, _ = self.prepareMacro("pt10_base", *params)
+        self.runMacro(macro)
+        result = macro.data
+        msg = MSG_TEMP % (result, expect_params)
+        assert expect_params == result, msg
+
+
+class createMacro(Macro):
```
reszelaz: I miss the execution using spock syntax string, e.g. "pt6_base 99 [1 2]" or "pt6_base 99 1 2". The same applies for execMacro macro.

droldan: Solved at [commit 6923430](https://github.com/sardana-org/sardana/pull/781/commits/6923430d779d98e034b2f19930d42d0e6d6e5185)

src/sardana/macroserver/test/res/macros/testmacros.py
droldan: Solved at [commit b31a5b3](https://github.com/sardana-org/sardana/pull/781/commits/b31a5be9ebfbb9d9d5226dc0ab3e395cc52468c2)

src/sardana/macroserver/test/res/macros/testmacros.py
droldan: Solved at [commit 6923430](https://github.com/sardana-org/sardana/pull/781/commits/6923430d779d98e034b2f19930d42d0e6d6e5185)

reszelaz:
@sardana-org/integrators this PR is now ready to be reviewed and eventually integrated.
Unittests for param parser covering all marco parameters examples were provided.
Furthermore tests of macro API e.g. `execMacro`, `createMacro`, `runMacro` were also provided.
All tests are greeen now. So, I think that manual tests are not required, but of course one can try if everyting works as expected manually.

When implementing this feature we doubt if implement the same optional usage of square brackets for the programmatic macro API `execMacro`, `createMacro`, `runMacro` , etc. In this case it is to allow passing repeat parameter's values (only if the repeat paramter contains one member) as single value and not a list with only one element e.g.

`Macro.execMacro("pt10", 1, mot1)`

similarly to what we allow now in spock: 

`Door> pt10 1 mot1`

Of course, as in case of the spock syntax it is optional, so one can pass the list:

`Macro.execMacro("pt10", [1], mot1)`

similarly to what is possible in spock: 

`Door> pt10 [1] mot1`

We implemented the same feature for the programatic macro API already, but if you don't agree on that it is easy to remove it, just reverting a8814ee and 43c059d.

reszelaz:
Thanks @teresanunez for the review! I'm merging it then. 

I someone does not agree on this:

> We implemented the same feature for the programatic macro API already, but if you don't agree on that it is easy to remove it, just reverting a8814ee and 43c059d.

we will be on time to remove it until the official release Jan19.

Thanks!

PR #782, opened 2018-06-13
Add macro stage logging

AntoineDupre:
Hi,

The logging system and the hooks features presented at the last tango meeting look quiet promising,  this PR add some logs in the macro server during the scan execution. 
A log is generated before each hook.

Cheers,
/Antoine

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -975,6 +975,7 @@ class SScan(GScan):
 
         if hasattr(macro, 'getHooks'):
             for hook in macro.getHooks('pre-scan'):
+                self.info("Macro stage: pre-scan hook {} ...".format(hook))
                 hook()
 
         self._sum_motion_time = 0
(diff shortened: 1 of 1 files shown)


reszelaz:
@AntoineDupre thanks for your contribution!

I couldn´t try the code yet, just quickly reviewed it. I have some initial comments:
I see that there are new logs for:
1. Before motion and acquisition in step scan.
You used the info level. I think this will be printed in the spock output, and I´m affraid that not all the users would be interested in seeing this. Actually there are already debug logs like `[START] motion` printed just before. If you are interested in seeing the debug logs in spock you could just enable them by executing the magic command `debug on|off`. If you miss in this messages the actual positions that the motors will be moved to or the integration time, IMHO it is fine to add them (let´s see what others think as well..).
2. Before hooks execution.
IMHO it is very interesting to add these messages. I would also go for the debug log level here (again let's wait for others opinions). Regarding the implementation, maybe it would be possible to add it on the `Hookable` class level. I imagine it should be possible to somehow decorate/wrap the callable object at the moment of adding it to the list of hooks. Then, we could avoid the duplication of code and other than just the scan macros could benefit from this.

reszelaz:
> I think this will be printed in the spock output

Forget about it. You used the scan object logging `self.info` and not the macro logging `self.macro.info`.

PR #783, opened 2018-06-14
Add public access to data_handler in GScan

reszelaz:
As agreen in #711 we need a public access to data_handler object of `GScan`, for example for adding
custom data. Add a read only property.

Whenever integrated we should change the examples accessing to it via `_data_handler` protected member e.g. `ascan_with_addcustomdata` macro.


teresanunez:
I have tested it and it works fine. IMO it can be integrated.

reszelaz:
Great! Please go ahead then. Many thanks!

PR #785, opened 2018-06-18
Deprecate PoolMotorSlim in favor of PoolMotorTV

reszelaz:
PoolMotorSlim is an old widget for controlling motors that was
developed before TaurusForm and TaurusValues existed. PoolMotorTV
is the modern alternative to PoolMotorSlim. It was discussed several
times that PoolMotorSlim should be deprecated, for example in #163, so
deprecate it now.


reszelaz:
If all are ok with this please any of @sardana-org/integrators could click on merge?

PR #786, opened 2018-06-21
Allow function generation when group has 1 repeat only

reszelaz:
Allow function generation when group has 1 repeat only

When group has one repeat only the generation of active and passive events
is much simpler. It requires only Initial and Active parameters and Total
parameter is not needed. Implement different algorithm of generation of
events when repeats is 1.

Any @sardana-org/integrators could take a look on this, please?

### DIFF
--- a/src/sardana/util/funcgenerator.py
+++ b/src/sardana/util/funcgenerator.py
@@ -38,6 +38,7 @@ def strictly_increasing(l):
     """Check whether list l has strictly increasing values"""
     return all(x < y for x, y in zip(l, l[1:]))
 
+
 def strictly_decreasing(l):
     """Check whether list l has strictly deacreasing values"""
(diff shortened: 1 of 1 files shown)


reszelaz:
This PR allows to use funciton generator with simplified synchronization descriptions, for example:

```python
from sardana.pool import SynchDomain, SynchParam 

Initial = SynchParam.Initial
Active = SynchParam.Active
Time = SynchDomain.Time

valve1_conf = [
    {
        Initial: {Time: 1},
        Active:  {Time: 0.3}
    },
    {
        Initial: {Time: 5},
        Active:  {Time: 0.3}
    }
]
```

teresanunez:
I have tested the function generator with the configuration above and it works with the changes proposed by @reszelaz. I merge it.

reszelaz:
Great! Thanks for the review!

PR #794, opened 2018-06-26
Continuous scans may loose data

reszelaz:
In one of our setups using continuous scans we found that sporadically the hardware synchronized channels were loosing data but the scan was correctly finishing. With the involved hardware this is not possible and was pointing to a bug in software.

After some debugging we found a wrong implementation on handling the value buffer (data) events on the MacroServer side - see 92dc5c6 comment for more detail.

This issue can be reproduced with the following simplified snippet, when while executing one would see "i is different than idx" messages:

```python
import time
import threading
from sardana.sardanathreadpool import get_thread_pool

pool = get_thread_pool()


lock = threadingLock()
idx = 0


def job(i):
    global idx
    with lock:
        print "Entering critical section of job %d" % i
        if i != idx:
            print "i is different than idx"
        idx += 1
        time.sleep(0.5)
        print "Leaving critical section of job %d" % i


for k in xrange(1000):
    pool.add(job, None, k)

time.sleep(1000)

```

And can be fixed by using a dedicated pool with only one thread where while executing the above messages are not printed:

```python
import time
import threading
from taurus.core.util.threadpool import ThreadPool


__thread_pool_lock = threading.Lock()
__thread_pool = None


def get_thread_pool():
    global __thread_pool

    if __thread_pool:
        return __thread_pool

    global __thread_pool_lock
    with __thread_pool_lock:
        if __thread_pool is None:
            __thread_pool = ThreadPool(name="ValueBufferTH", Psize=1,
                                       Qsize=100000)
        return __thread_pool


pool = get_thread_pool()


lock = threading.Lock()
idx = 0


def job(i):
    global idx
    with lock:
        print "Entering critical section of job %d" % i
        if i != idx:
            print "i is different than idx"
        idx += 1
        time.sleep(0.5)
        print "Leaving critical section of job %d" % i


for k in xrange(1000):
    pool.add(job, None, k)

time.sleep(1000)

```

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -51,6 +51,7 @@ from taurus.core.util.log import Logger
 from taurus.core.util.user import USER_NAME
 from taurus.core.tango import FROM_TANGO_TO_STR_TYPE
 from taurus.core.util.enumeration import Enumeration
+from taurus.core.util.threadpool import ThreadPool
 
 from sardana.util.tree import BranchNode, LeafNode, Tree
 from sardana.util.motion import Motor as VMotor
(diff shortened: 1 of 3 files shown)


teresanunez:
Hi @reszelaz, I have tested your scripts, and I confirm that the problem that one can see in the first one is solved with the second implementation. The problem should be then also solved for Sardana with the changes you propose hier.

reszelaz:
Thanks @teresanunez for reviewing this!

I just realized that it requires some more modifications. Now the pool of threads is a singleton and we need one instance per continuous scan execution. There may be many scans running in parallel in different doors. I will make this modifications today.

reszelaz:
I have just pushed the change that I mentioned earlier. @sardana-org/integrators, now it is ready for final review and eventual integration.

teresanunez:
Hi @reszelaz, I have checked this and I have not found any problem. IMO it can be merged.

reszelaz:
Great! Go ahead with the merge please. Many thanks!

PR #801, opened 2018-07-03
Add configuration attribute to DiscretePseudoMotorController

jordiandreu:
Provide a more pythonic interface to the discrete pseudo motor
configuration.

The new configuration attribute encapsulates the values
from the label and calibration attributes as a dictionary.

The dictionary keys are extracted from the labels attribute.
and each dictionary entry is also a dictionary with four keys:
a `pos` key corresponding to the label enumerated position
integer value and three float values corresponding to the
calibration values and named as `min`, `set`, and `max`.

Three different macros are proposed to manage this configuration:
`def_dpm_pos`, `udef_dpm_pos` and `prdef_dpm` to add/remove a point
to the configuration and to print the configuration on screen.

### DIFF
--- a/src/sardana/macroserver/macros/discretepm.py
+++ b/src/sardana/macroserver/macros/discretepm.py
@@ -39,7 +39,7 @@ class DiscretePseudoMotorConfiguration(dict):
         self.update(conf)
 
     def get_configuration(self):
-        value = self.pseudo.read_attribute('configuration').value
+        value = self.pseudo.getAttribute('configuration').read().value
         fmt, data = self.json.decode(('json', value))
         return data
 
@@ -82,7 +82,7 @@ class DiscretePseudoMotorConfiguration(dict):
     def _update(self):
         try:
             fmt, value = self.json.encode(('', self))
-            self.pseudo.write_attribute('configuration', value)
+            self.pseudo.getAttribute('configuration').write(value)
             self.macro.debug('Updated configuration:\n{0}'.format(self))
         except Exception as e:
             msg = 'Cannot update configuration]\n{0}\{1}'.format(e, self)
--- a/src/sardana/macroserver/macros/discretepm.py
+++ b/src/sardana/macroserver/macros/discretepm.py
@@ -1,0 +1,124 @@
+from sardana.macroserver.macro import Macro, Type
+import json
+import math
+
+
+class DiscretePseudoMotorConfiguration(dict):
+    def __init__(self, pseudo_obj, macro):
+        self.pseudo = pseudo_obj
+        self.macro = macro
+        _physical_motor_name = self.pseudo.physical_elements[0]
+        self.motor = macro.getMoveable(_physical_motor_name)
+        conf = self.get_configuration()
+        super(DiscretePseudoMotorConfiguration, self).__init__(conf)
+
+    def get_configuration(self):
+        return json.loads(self.pseudo.read_attribute('configuration').value)
+
+    def has_calibration(self):
+        return all(['set' in self[x].keys() for x in self.keys()])
+
+    def add_point(self, label, pos, dmin, dmax, set):
+        point = dict()
+        point['pos'] = int(pos)
+        # Calculate point calibration if required
+        if self.has_calibration():
+            # Set to current physical position if no value supplied as argument
+            if math.isinf(set):
+                point['set'] = self.motor.position
+            else:
+                point['set'] = float(set)
+            # If point exists, we use current min, max values
+            if label in self.keys() and math.isinf(dmin) and math.isinf(dmax):
+                p = self[label]
+                min_pos = point['set'] - abs(p['set'] - p['min'])
+                max_pos = point['set'] + abs(p['set'] - p['max'])
+            # else, new point has new calibration,
+            else:
+                min_pos = point['set'] - dmin
+                max_pos = point['set'] + dmax
+
+            point['min'] = min_pos
+            point['max'] = max_pos
+
+        self[label] = point
+        self._update()
+
+    def remove_point(self, label):
+        try:
+            self.pop(label)
+            self._update()
+        except Exception as e:
+            self.macro.error('Cannot remove label {0}\n{1}'.format(label, e))
+
+    def _update(self):
+        try:
+            self.pseudo.write_attribute('configuration', json.dumps(self))
+            self.macro.debug('Updated configuration:\n{0}'.format(self))
+        except Exception as e:
+            msg = 'Cannot update configuration]\n{0}\{1}'.format(e, self)
+            self.macro.error(msg)
+
+    def __str__(self):
+        return json.dumps(self, indent=4, sort_keys=True)
+
+
+class def_dpm_pos(Macro):
+    """
+    Define a (calibrated) point for a discrete pseudomotor configuration.
+
+    The mandatory parameters to execute the macro are: pseudo, label and pos.
+
+    Two different scenarios exist: To define a new point or to modify an
+    existing one. The controller protects from uploading repeated pos values.
+
+    If the point is new, the default dmin and dmax parameters are used to
+    construct the calibration. If no set point is provided, the current
+    physical position is used instead.
+
+    If the point already exists, the values are updated as in the previous
+    case. However, if no dmin and dmax are provided, the previous
+    calibration values for dmin and dmax are calculated and used to rebuild
+    the calibration.
+    """
+    param_def = [
+        ['pseudo', Type.PseudoMotor, None, 'Discrete pseudomotor name'],
+        ['label', Type.String, None, 'Label name'],
+        ['pos', Type.Integer, None, 'Pseudo position'],
+        ['dmin', Type.Float, float('-inf'),
+         'Delta increment defining the minimum position'],
+        ['dmax', Type.Float, float('inf'),
+         'Delta increment defining the maximum position'],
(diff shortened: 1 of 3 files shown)


src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class DiscretePseudoMotorConfiguration(dict):
+    def __init__(self, pseudo_obj, macro):
+        self.pseudo = pseudo_obj
+        self.macro = macro
+        _physical_motor_name = self.pseudo.physical_elements[0]
+        self.motor = macro.getMoveable(_physical_motor_name)
+        conf = self.get_configuration()
+        super(DiscretePseudoMotorConfiguration, self).__init__(conf)
```
reszelaz: Maybe `dict.update` could be used instead of the late constructor call?

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def get_configuration(self):
+        return json.loads(self.pseudo.read_attribute('configuration').value)
```
reszelaz: `self.pseudo` is a Taurus device. Tt is better to use taurus API to read attribute.

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+    def remove_point(self, label):
+        try:
+            self.pop(label)
+            self._update()
+        except Exception as e:
+            self.macro.error('Cannot remove label {0}\n{1}'.format(label, e))
+
+    def _update(self):
+        try:
+            self.pseudo.write_attribute('configuration', json.dumps(self))
```
reszelaz: `self.pseudo` is a Taurus device. Tt is better to use taurus API to write attribute.

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+            self.macro.debug('Updated configuration:\n{0}'.format(self))
+        except Exception as e:
+            msg = 'Cannot update configuration]\n{0}\{1}'.format(e, self)
+            self.macro.error(msg)
+
+    def __str__(self):
+        return json.dumps(self, indent=4, sort_keys=True)
+
+
+class def_dpm_pos(Macro):
```
reszelaz: This applies to all new macros. Not 100 % sure about this..

What do you think to validate pseudo motor parameter if it is of `DiscratePseudoMotorController` class?
`pseudo.getControllerObj().klass == "DiscretePseudoMotorController"` this should do the job.

One disadvantage of this is that we won't be able to use different pseudo motors that implements the same API e.g. inheritance..

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Print discrete pseudomotor configuration.
+    """
+    param_def = [
+        ['pseudo', Type.PseudoMotor, None, 'Discrete pseudomotor name'],
+    ]
+
+    def run(self, pseudo):
+        conf = DiscretePseudoMotorConfiguration(pseudo, self)
+        self.output(conf)
```
reszelaz: This is just a matter of taste, but I like more a table like output. But no need to change this...
If you want to check what `from taurus.console.table.Table` can do see for example `wm` macro.

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+        conf = self.get_configuration()
+        super(DiscretePseudoMotorConfiguration, self).__init__(conf)
+
+    def get_configuration(self):
+        return json.loads(self.pseudo.read_attribute('configuration').value)
+
+    def has_calibration(self):
+        return all(['set' in self[x].keys() for x in self.keys()])
+
+    def add_point(self, label, pos, dmin, dmax, set):
```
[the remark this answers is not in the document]
jordiandreu: Done.

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
+from sardana.macroserver.macro import Macro, Type
```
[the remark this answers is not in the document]
jordiandreu: Done.

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+import json
```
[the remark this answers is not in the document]
jordiandreu: Done

src/sardana/macroserver/macros/discretepm.py
```diff
@@ -0,0 +1,124 @@
...
+import math
```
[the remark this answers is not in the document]
jordiandreu: Done.

src/sardana/macroserver/macros/discretepm.py
jordiandreu: Done.

src/sardana/macroserver/macros/discretepm.py
jordiandreu: I used the access to the attribute the dictionary form, however a discussion about the API still ongoing... https://github.com/taurus-org/taurus/issues/42

src/sardana/macroserver/macros/discretepm.py
jordiandreu: Done

src/sardana/macroserver/macros/discretepm.py
jordiandreu: ...which is the case.

reszelaz: Then in this case the check would not work. And what if we check if the pseudo motor has the Configuration attribute? We can go even further and check if its format satisfies the macro needs e.g. has certain keys, etc. Otherwise we could raise a ValueError("pseudo motor is not of discrete type ")

src/sardana/pool/poolcontrollers/DiscretePseudoMotorController.py
```diff
@@ -38,39 +38,56 @@
...
+    """
+    A discrete pseudo motor controller which converts physical motor
+    positions to discrete values"""
 
     gender = "DiscretePseudoMotorController"
     model = "PseudoMotor"
     organization = "Sardana team"
     image = ""
 
-    pseudo_motor_roles = ("OutputMotor",)
```
reszelaz: IMHO neither the previous names nor the new ones are self explanatory. And what about ContinuousMoveable and DiscreteMoveable?

reszelaz:
I have a more general comment about the `def_dpm_pos` macro:

According to the `def_dpm_pos` macro the dmin and dmax are described:

>         dmin : (Float) Delta increment defining the minimum position
>        dmax : (Float) Delta increment defining the maximum position

Then I was using the macro in the following way:

```
Door_zreszela_1 [6]: %def_dpm_pos dmot01 pos01 1 -1 1 0
```

And I get the following configuration:

```
Door_zreszela_1 [10]: dmot01.Configuration
       Result [10]: '{"pos01": {"max": 1.0, "set": 0.0, "pos": 1, "min": 1.0}}'
```

Personally I would expect: 

```
'{"pos01": {"max": 1.0, "set": 0.0, "pos": 1, "min": -1.0}}'
```

To get the above result I need to execute:

```
Door_zreszela_1 [12]: %def_dpm_pos dmot01 pos02 2 1 1 0
Door_zreszela_1 [13]: dmot01.Configuration
       Result [13]: '{"pos02": {"max": 1.0, "set": 0.0, "pos": 2, "min": -1.0}}'
```

Is it intentional?

src/sardana/pool/poolcontrollers/DiscretePseudoMotorController.py
jordiandreu: We have adopted the proposed names.

src/sardana/macroserver/macros/discretepm.py
jordiandreu: This is not the first time we need to face this situation (get the parent controller class from a derived controller). Maybe a better option is try to export some extra controller information (Gender, etc) to the Tango layer. Further investigation...

reszelaz: Good idea! I should be easy to implement. We just need to add a method `getClassObj` to the `Controller` taurus extension. The gender information is already serialized by the Pool and arrives to the MacroServer.

reszelaz:
Thanks for the changes! Now it looks ready to be integrated.

Just a final thought about the names.
* IMHO for the macro module name it is enough to call it `discrete.py`
* The proposed macro names are `def_dpm_pos`, `udef_dpm_pos` and `prdef_dpm_conf`. Some alternatives could be: 
    1. `def_disc_pos`, `udef_disc_pos` and `prdef_disc_pos` (or just `prdef_disc`)
    2. `def_discr_pos`, `udef_discr_pos` and `prdef_discr_pos` (or just `prdef_discr`)

...but I'm not good in finding catchy names. Anyone is invited to vote/propose names, thanks!

jordiandreu:
* Regarding the module name, _discrete.py_ is fine for me too.
* Regarding the macro's names, I like the second form within brackets:
`def_discr_pos`, `udef_discr_pos` and `prdef_discr`

reszelaz:
I renemed module and macros as agreed and marked the new macros as experimental. Merging it. Thanks for the contribution!

PR #808, opened 2018-07-05
add hooks to ct macro

dschick:
allow pre-acq and post acq-hooks for ct macro.
Add integ_time as property of ct macro class to make it accessible from within the hooked macros

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -642,14 +642,17 @@ class tw(iMacro):
 ##########################################################################
 
 
-class ct(Macro):
+class ct(Macro, Hookable):
     """Count for the specified time on the active measurement group"""
 
     env = ('ActiveMntGrp',)
-
+    hints = {'allowsHooks': ('pre-acq', 'post-acq')}
     param_def = [
         ['integ_time', Type.Float, 1.0, 'Integration time']
     ]
+    # integration time has to be accessible from with in the hooks
+    # so declare it also as a property of the ct class
+    integ_time = 0 
 
     def prepare(self, integ_time, **opts):
         mnt_grp_name = self.getEnv('ActiveMntGrp')
@@ -660,13 +663,20 @@ class ct(Macro):
         if self.mnt_grp is None:
             self.error('ActiveMntGrp is not defined or has invalid value')
             return
-
+        self.integ_time = integ_time # set integ_time property for hooks
         self.debug("Counting for %s sec", integ_time)
         self.outputDate()
         self.output('')
         self.flushOutput()
+        
+        for preAcqHook in  self.getHooks('pre-acq'):
+            preAcqHook()
 
         state, data = self.mnt_grp.count(integ_time)
+        
+        for postAcqHook in  self.getHooks('post-acq'):
+            postAcqHook()
+            
         names, counts = [], []
         for ch_info in self.mnt_grp.getChannelsEnabledInfo():
             names.append('  %s' % ch_info.label)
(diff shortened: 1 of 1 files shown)


dschick:
does this really make sense? could you explain your use case?

jangarrevoet:
I check in the pre-scan hooks several different states of the beamline before allowing the scan to proceed.
Other actions happen after a scan, such as playing sounds, sending out emails/text messages,...
Now that the data in ct is made available one can also use post processing steps that are already implemented for other scans in the post-scan hook in ct.

dschick:
okay I agree, but do you need to check your settings or want to sent an email for a single ct?
I totally agree for any kind of scan but I find it wired for a ct

reszelaz:
For me `pre-scan` and `post-scan` hook places sounds to be limited to scans. Placing them also in `ct` macro, which is not a scan, may be confusing.

However I see your point @jangarrevoet. Do you use the general hooks for this?

jangarrevoet:
Many users consider ct to be a scan of a single point. I have gotten a lot of complaints in the past that ct also does not store the data. That is why I have in my own installation a cnt macro that has all hooks implemented and that also saves the data.

teresanunez:
I see the two views ... but I also think that it can be dangerous to treat ct as a 'single point scan' ... some
 users can use ct only for making some test counting and not want to activate all the procedures/hooks that are used for an scan. I think the use of ct as 'single point scan' is a misuse, even if it could be useful for some people.

teresanunez:
We should then also start thinking in increasing the scanID when running ct ... if not we could have problems in the identification of the scans.

jangarrevoet:
@teresanunez I see your point.
I also see using ct as misusage of the macro. I will keep my more specialised cnt macro on the side to make point scans. BTW there I actually use a countID as an identifier.
I guess we then can keep the hooks just for pre-acq and post-acq.

src/sardana/macroserver/macros/standard.py
```diff
@@ -642,14 +642,17 @@ def run(self, motor, delta):
...
 
     env = ('ActiveMntGrp',)
-
+    hints = {'allowsHooks': ('pre-acq', 'post-acq')}
     param_def = [
         ['integ_time', Type.Float, 1.0, 'Integration time']
     ]
+    # integration time has to be accessible from with in the hooks
+    # so declare it also as a property of the ct class
+    integ_time = 0
```
reszelaz: I think there is no need to declare this on the class level. With setting the instance attribute `self.integ_time` in the run method it should be enough.

reszelaz:
Since we are planning to clarify the measurement concept in Sardana (see [SEP18-Terminology](https://github.com/reszelaz/sardana/blob/sep18/doc/source/sep/SEP18.md#terminology)). We could think of adding new hook place e.g. `pre-meas` and `post-meas` that could be added to `ct` and scans.

The ct would become:
1. pre-meas
2. pre-acq
3. count
4. post-acq
5. post-meas

The scan would become:
1. pre-scan
2. pre-meas
3. loop of pre-acq and post-acq or just one pre-acq and one pre-acq depending if this is a step or continuous scan
4. post-meas
5. post-scan

I think that this would fulfill with all the visions. If there is an interest in this I would treat it in another issue/PR and rename this one to "add acq hooks to ct". It is just to keep it simple and integrate it quickly.

Many thanks for all your points of view!

src/sardana/macroserver/macros/standard.py
dschick: okay, sure. I will change that in a bit.

dschick: I could get rid of it, by accessing the integ_time via the measurementGroup, however, it seems that at the time of the pre_acq hook, the integ_time of the measurementGroup (getIntegrationTime()) is not updated to the current value of the current acquisition.

reszelaz: Yes, for making it accessible already from the `pre-acq` hook it must be added to the instance. I think that's fine.

reszelaz:
Tested and works nicelly. I'm merging it. It would be nice to add the same for the `uct` macro.

PR #813, opened 2018-07-06
Join pool of threads in cont scan

reszelaz:
`CAcquisition` is creating a new pool of threads everytime it is instantiated.
These threads requires disposal, othewise the OS will complain that it can
not create more threads.

Add a method to join the pool of threads and call it from continuous scan
and timescan.

Can be tested by counting the number of threads in a process using:
`$> ps huH p <macroserver-pid> | wc -l`
Before it was increasing after every execution of the continuous scan, now it stays contant.

Can @sardana-org/integrators take look on this, please.

PR #816, opened 2018-07-07
Fix cyclic references between scan macros and GSF internals

reszelaz:
When stress testing #813 with a fast loop of continuous scans of 4000 points with ~10 channels we figured out that the memory of the MacroServer was continuously increasing ~3 MB per minute. At some time, when the MacroServer was beyond 1 GB a couple of `MemoryErrors` were logged:

```
ValueBufferTH.W001 ERROR    2018-07-05 20:37:04,350 ValueBufferTH.ValueBufferTH.W001: Uncaught exception running job 'addData' called from thread Dummy-4:
  File "/homelocal/sicilia/lib/python/site-packages/taurus/core/tango/tangoattribute.py", line 473, in push_event
    listeners=listeners)
  File "/homelocal/sicilia/lib/python/site-packages/taurus/core/taurusmodel.py", line 256, in fireEvent
    l.eventReceived(self, event_type, event_value)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/taurus/core/tango/sardana/pool.py", line 227, in eventReceived
    EventGenerator.fireEvent(self, v)
  File "/homelocal/sicilia/lib/python/site-packages/taurus/core/util/event.py", line 276, in fireEvent
    cb(data, event_val)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/taurus/core/tango/sardana/pool.py", line 1704, in valueBufferChanged
    self._value_buffer_cb(channel, value_buffer)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/scan/gscan.py", line 1889, in value_buffer_changed
    self._countdown_latch.count_down, info)
  File "/homelocal/sicilia/lib/python/site-packages/taurus/core/util/threadpool.py", line 132, in run
    callback(cmd(*args, **kw))
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/scan/scandata.py", line 465, in addData
    self.tryToAdd(idx, label)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/scan/scandata.py", line 479, in tryToAdd
    self.datahandler.addRecord(self, rc)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/scan/recorder/datarecorder.py", line 68, in addRecord
    recorder.writeRecord(record)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/scan/recorder/datarecorder.py", line 142, in writeRecord
    self._writeRecord(record)
  File "/homelocal/sicilia/lib/python/site-packages/sardana/macroserver/recorders/output.py", line 98, in _writeRecord
    data[name] = record.data[name]
MemoryError
```

but the scan continued. After a while a burst of `API_EventTimeout` killed the sequence.

In order to debug this problem I have placed some debug traces in the Macro class constructor and destructor, something like:

```python
    def __init__(self):
        [...]
        print("!!!!!!!!!!!!Creating macro: %s!!!!!!!!!!!!!" % id(self))

    def __del__(self):
        print("!!!!!!!!!!!!Deleting macro: %s!!!!!!!!!!!!!" % id(self))
```
When running different macros I realized that the scan macros are not being destructed, while the others yes. The destruction happens on the next macro execution (this has to do with the preserve macro data feature).

This pointed me to check the cyclic references between the scan macro objects and the GSF internal objects. I found that there are cyclic references. After removing them, the scan macros are also destructed.

Please @sardana-org/integrators can you take a look on this?


daneos:
I was able to reproduce something similar using Sardana 2.5.0 on my testing VM with dummy controllers.
For me, the memory usage increase was not continuous, but rather the MacroServer used up more memory when starting a scan.
After reaching 1GB of reserved memory there was a drop to ~900MB a few times, but then the memory usage stopped dropping (probably Python GC ran out of thing to clear).
I did not see any `MemoryError`s, after reaching ~1.5GB the MacroServer was just killed by the OS.

Tested on CentOS 7 VM with 2GB of RAM and no swap.
Tomorrow I'll run the same test with this PR and report the results.

daneos:
I repeated the tests with this PR and the result is the same, the MacroServer dies with ~1.5GB used memory, however I was doing the tests using [the `rep` macro](https://github.com/synchrotron-solaris/lib-solaris-facility-macros/blob/master/facility_sequence.py#L41) from SOLARIS macros.

When using Sardana's `repeat` macro I was not able to reproduce this issue neither on Sardana 2.5.0 nor with this PR, the MacroServer goes up to 1.3GB and then stays there forever.

Tomorrow I'll run the tests again, this time on a machine with much less RAM available...

reszelaz:
> however I was doing the tests using the rep macro from SOLARIS macros.

This could be due to the fact that you store a reference to the current macro. Then there is a cyclic reference between the `rep` macro and the current macro. Could you try to use weakref for the current macro and see if if behaves the same?

Also looking at the code of the `rep` macro, I think that the `on_abort` method is not necessary. I think that the current macro should be automatically stopped. Have you observed something different?

daneos:
I did the tests using CentOS 6.6 with Python 2.6, but still can't reproduce the issue.
I think it can be merged though, since it doesn't break anything and having weakrefs in GScan will probably be beneficial not only for this issue, but will also help to avoid similar memory issues in future.

reszelaz:
I found that this was breaking timescan and fixed this. I also merged the latest develop branch.
There are no public API changes that I was worried in https://github.com/sardana-org/sardana/issues/1095#issuecomment-482480417, so I'm merging. Thanks!

Issue #806, opened 2018-07-05
expconf is not automatically updating when configuration changes


dschick:
not sure if this is already issued, but when I add new counters to my pool and I have the expconf window open, I do not seem them in the list of counters to add after a refresh. I have to close+reopen the expconf window.

Moreover, when changing the counter which are plotted I also need to close/reopen the trend window.

Best,

Daniel

reszelaz:
Yes, this is a known limitation/bug. Actually I was thinking that it was already reported, but now I can not find any issue talking about it.

reszelaz:
If someone finds the original report please post the link here. Meanwhile I think it is better to keep this one open for other users.

dschick:
this is especially a bit annoying because of this #753 - so everytime you reopen the expconf you have to manually correct the FNDQ in the snapshot

reszelaz:
I agree, we will try to continue with #753 ASAP in order to have all the FQDN issues solved in Jul18 release.

dschick:
updating the expConf is also important in context of new macros like `newfile `or maybe like a `plotselect`.
It just happend to me several times that I run a macro which enables/disables counters while the expconf is open and when I then change a setting in the expconf, it overwrites all setting I made from within the macros. I guess a global refresh button or maybe even a periodic refresh would be nice - maybe one can trigger an update also via some events from within the macros.

reszelaz:
I totally agree.
The events are already emitted by the server. We just need to listen to them in the expconf.
It is just that we did not have time to work on it yet. But since we are now adding macros for manipulating the same configuration it will become more urgent..

rhomspuron:
Hi @dschick 

Could you review the PR #882? Any feedback is welcome. 

Best,
Roberto

reszelaz:
Solved in #882

PR #824, opened 2018-07-09
plotselect macro

dschick:
implementation of SPECs `plotselect `macro from the snippet of @droldan in #690 
The implementation is currently just a workaround until the measurement group API has changed.


dschick:
I guess to to #806 the `plotselect` changes the configuration, but the Trend just updates after closing and reopening, which is of course not the intention of the `plotselect` macro

reszelaz:
Yes, I confirm this. And indeed it has to due with #806.

droldan:
The macro could use the new MeasurementGroup Configuration API #867

dschick:
@droldan your right with the new API, I would wait until it is merged to the develop branch and then update the plotselect macro accordingly

reszelaz:
Hi Daniel,

It would be great to include this macro in the next release.
With the recenly added [MeasurementGroup configuration methods](https://sardana-controls.org/devel/api/sardana/taurus/core/tango/sardana/pool.html#sardana.taurus.core.tango.sardana.pool.MeasurementGroup) it should be now possible to make the refactoring.

One idea on implementing this could be (I have not tried it):
```python

enabled_channels = meas_grp.getEnabled()
plot_channels_ok = []
for channel in plot_channels:
    enabled = enabled_channels.get(channels)
    if enabled is None:
        self.warning("{} not in {}".format(channel, meas_grp.name))
    else:
        plot_channels_ok.append(channel)
        if not enabled:
            self.warning("{} is disabled".format(channel))
        plot_channels_ok.append(channel)
meas_grp.setPlotType("No", apply=False)
meas_grp.setPlotType("Spectrum", *plot_channels_ok, apply=False)
meas_grp.setPlotAxes(["<mov>"], *plot_channels_ok)
```

Just let me know if you want me to make the refactoring.

dschick:
I will try to make it till the next release. I was actually waiting for the MeasurementGroup API update.
Thanks for the snippet.

reszelaz:
Thanks! Sorry you had to wait so long.
If you find problems with the new API or you would like to suggest changes please share it with us.

dschick:
I have updated the macro to the new measurement API mostly following your snippet - thanks for that.
Let me know, if anything is missing/wrong

src/sardana/macroserver/macros/standard.py
```diff
@@ -1049,3 +1049,41 @@ def run(self, ScanFilePath_list, ScanID):
...
+          ['channel',
+           [['channel', Type.ExpChannel, 'None', ""], {'min': 0}],
+           None,
+           "List of channels to plot"],
+     ]
+
+    def run(self, channel):
+        try:
+            active_meas_grp = self.getEnv('ActiveMntGrp')
+        except UnknownEnv:
```
reszelaz, line 1066: In the `scanstats` macro in #880 I suggested to raise the exception to prevent the eventual subsequent macros in the sequence to proceed. I would recommend the same here. In this case it may be less critical because this macro does not produce any results that other macros could expect but still I'm in favor of ending with the exception.

I will make the necessary change and use the macro's `env` class member to indicate that the `ActiveMntGrp` is mandatory.

reszelaz: Done in 336bb46

dschick: I am pretty fine with that.

PR #827, opened 2018-07-10
Remove Sardana GUI screenshot

reszelaz:
Sardana GUI is not more maintained. Remove its screenshot from the docs.


amilan:
I think you should push the file containing the .png link

amilan:
Now looks better!!

Issue #826, opened 2018-07-10
Migrate docs from RTD to Travis


cpascual:
Migrate the doc auto-buid and publication from readthedocs to Travis.
See why and how Taurus did it in: taurus-org/taurus#572

Tasks:

- [x] build docs from travis and deploy them to github (implemented in #830)
- [x] change the sardana-controls.org domain to point to https://sardana-org.github.io/sardana-doc/ and configure the sardana-doc gihub pages accordingly
- [x] add links to other doc versions (implemented in #842)

cpascual:
I created the a PR implementing this: #830

cpascual:
I implemented the final steps of the move in PR #842

reszelaz:
I think this is done already. Closing...

PR #830, opened 2018-07-10
Build and deploy docs from travis

cpascual:
Build the docs in travis and deploy them to the sardana-doc repo using doctr (see how in https://github.com/taurus-org/taurus/pull/572 ). This is part of the work for #826 

The results can be seen, in:

https://sardana-org.github.io/sardana-doc/

... for the develop version (only after we merge this PR) and in

https://sardana-org.github.io/sardana-doc/v-BRANCH_OR_TAG_NAME/

... for other tags or branches in the sardana repo (for example, see https://sardana-org.github.io/sardana-doc/v-travisdocs )

Important: note that in this PR I am not disabling the parts needed by RTD, so the docs are still built by RTD in parallel and served in their usual location, while the docs built by travis get served by GH pages. This would allow us merge the PR without breaking RTD and test and compare before deciding to point our custom domain to GH pages instead of RTD.


cpascual:
The travis builds are failing due to missing deps in the sardana-test docker image.
I just added the mssing deps, so we should try to rebuild once the [updated image is ready in dockerhub](https://hub.docker.com/r/reszelaz/sardana-test/builds/)

cpascual:
It is working now. You can see the generated docs in 
 https://sardana-org.github.io/sardana-doc/v-travisdocs

reszelaz:
I just modified the script to build the docs always, also for PR and when other tests fails.

PR #831, opened 2018-07-10
Docs: Avoid duplication of macro docstings

cmft:
Documentation of the macros in the macro catalogue is duplicated.

Refactor the structure to be according with the Sardana API.
Use automodule sphinx directive to generate the macro docs.
Fix minor issues in the macro modules.


daneos:
Looks great. Is it ready to be merged (it's still in "In dev")?

Issue #756, opened 2018-05-16
documentation for defctrl and defelem


dschick:
There is currently no documentation about how to add a controller and motor/counter/etc to sardana via spock or a potential wizard.
Although the current documentation is very good in explaining the concepts of controllers it is hard to find how to finnaly add them to the pool.
It would be sufficient to add have a general part about the defctrl and defelem macros in the _Writing Controllers_ part of the doucmentation.
Special emphasis could be layed on roles for pseudo counters/motors.

PR #834, opened 2018-07-10
Adding elements documentation

daneos:


dschick:
reads good!

julianofjm:
I think that it is missing to reference it from "Writing Controllers", #756

daneos:
@julianofjm Done!

PR #854, opened 2018-07-12
Add documentation for expconf tabs

daneos:


cmft:
It looks like perfect

PR #857, opened 2018-07-12
[WIP] Add how to pseudomotors

amilan:
This commit adds the documentation about how to create pseudo motors.


amilan:
There are many things that need to be updated, so this PR should be considered a WIP.

PR #863, opened 2018-07-13
Clean code from Taurus3/Taurus4 FQDN compat.

reszelaz:
Sardana element identifiers were changed to FQDN URIs during the Jan18
release. Prior to that a big code base was adapted to be compatible both
with Taurus3 and Taurus4. Remove this code cause with Taurus 4 it is no
more necessary.

@sardana-org/integrators can you take a look on this?


teresanunez:
I have tested this and it works fine with Taurus4. IMO it could be merged.

reszelaz:
Great! Doing merge then! Many thanks!

PR #864, opened 2018-07-16
Adding min_value and max_value to column_desc for mesh scans

teresanunez:
In the Nscans the column_desc in the data record contains min_value and max_value with the scan limits. We need this information also for the mesh scan. This PR implement it.

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -708,7 +708,9 @@ class mesh(Macro, Hookable):
         generator = self._generator
         moveables = []
         for m, start, final in zip(self.motors, self.starts, self.finals):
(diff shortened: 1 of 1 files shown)


reszelaz:
Looks good to me. Exactly the same as it is done for other mesh scans and the `aNscan`.

I just fixed some PEP8 errors. Whenever the CI gets green it can be merged. 

I'm just curious where is it used? Did it give you any errors without this code?

teresanunez:
Hi @reszelaz, thanks a lot. It is used in one of our private Tools for display, which needs to read the limits of the scan for plotting the limits.

reszelaz:
Ok. Now it is clear.
I'm mergin it cause it is green now.

PR #865, opened 2018-07-16
Fix Pool element taurus extensions when Pool object ref was not filled (Fix Issue 57)

reszelaz:
This solves #57 and redistribute to the correct classes the `getPoolData` and `getPoolObj` fallback implementation. More details in the comments of the commits and in the code itself.

This requires #863.


reszelaz:
This is ready for review @sardana-org/integrators. Thanks!

teresanunez:
I have tested this and it solves the issue #57

reszelaz:
Great, then I'm merging it.

PR #866, opened 2018-07-16
Allow back. comp. for defining ctrl. property

cmft:
Some old controllers define its description property using a string
'Description' instead of using the
'sardana.pool.controller.Description' constant.

A recent modification were base on this constant and cause an
exception during the loading of the controllers.

Add back. comp. raising a deprecation warning.

### DIFF
--- a/src/sardana/pool/poolmetacontroller.py
+++ b/src/sardana/pool/poolmetacontroller.py
@@ -263,19 +263,31 @@ class ControllerClass(SardanaClass):
         self.types = []
         self.dict_extra = {}
         self.api_version = 1
-
         klass = self.klass
         # Generic controller information
         self.ctrl_features = tuple(klass.ctrl_features)
 
         self.ctrl_properties = props = CaselessDict()
         self.ctrl_properties_descriptions = []
+        dep_msg = ("Define the 'Description' controller property using a "
+                   + "string is deprecated use "
+                   + "sardana.pool.controller.Description constant instead")
         for k, v in klass.class_prop.items():  # old member
             props[k] = DataInfo.toDataInfo(k, v)
-            self.ctrl_properties_descriptions.append(v['description'])
+            try:
+                self.ctrl_properties_descriptions.append(v[Description])
+            except KeyError:
+                self.deprecated(dep_msg)
+                self.ctrl_properties_descriptions.append(v['Description'])
+
         for k, v in klass.ctrl_properties.items():
             props[k] = DataInfo.toDataInfo(k, v)
-            self.ctrl_properties_descriptions.append(v['description'])
+            try:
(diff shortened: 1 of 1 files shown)


src/sardana/pool/poolmetacontroller.py
```diff
@@ -263,19 +263,31 @@ def __init__(self, **kwargs):
...
+        dep_msg = ("Define the 'Description' controller property using a "
+                   + "string is deprecated use "
+                   + "sardana.pool.controller.Description constant instead")
         for k, v in klass.class_prop.items():  # old member
             props[k] = DataInfo.toDataInfo(k, v)
-            self.ctrl_properties_descriptions.append(v['description'])
+            try:
+                self.ctrl_properties_descriptions.append(v[Description])
+            except KeyError:
+                self.deprecated(dep_msg)
```
reszelaz: I would use `self.warning` instead.
Now, using Taurus3 it gives the following message:
```
MainThread     WARNING  2018-07-16 16:11:41,888 LinkamTST350TempMotorCtrl: /home/zreszela/workspace/sardana/src/sardana/pool/poolcontrollermanager.py:463: DeprecationWarning: Define the 'Description' controller property using a string is deprecated use sardana.pool.controller.Description constant instead
  lib=controller_lib, klass=klass)
```
Where information about the poolcontrollermanager.py does not add information and may be even confusing.

I would us the following message:
"Definining the controller property description using a string is deprecated, use sardana.pool.controller.Description constant instead.

Using Taurus4 it is even worse. Since now Sardana is still Taurus3/Taurus4 compatible it must use the old API which is deprecated. In this case it is very convenient to run Sardan with `tauruscustomsettings._MAX_DEPRECATIONS_LOGGED` set to 1. In this case, if you have more than one controller class defining its properties descriptions with a string instead of the constant the deprecation warning will be set for only the first one.

src/sardana/pool/poolmetacontroller.py
```diff
@@ -263,19 +263,31 @@ def __init__(self, **kwargs):
...
+                self.ctrl_properties_descriptions.append(v['Description'])
+
         for k, v in klass.ctrl_properties.items():
             props[k] = DataInfo.toDataInfo(k, v)
-            self.ctrl_properties_descriptions.append(v['description'])
+            try:
+                self.ctrl_properties_descriptions.append(v[Description])
+            except KeyError:
+                self.deprecated(dep_msg)
```
reszelaz: The same as for the previous usage of `self.deprecated`.

src/sardana/pool/poolmetacontroller.py
reszelaz: After an internal talk to @cmft he agrees with the changes. I will implemen it.

PR #867, opened 2018-07-17
Implement MGConfiguration new API

droldan:
Implement a new API for the MGConfiguration class to set/get the configuration parameter.


teresanunez:
Hi, I wonder if it could be added to this API the methods for adding/removing channels from a measurement group. Perhaps I missed, but I think the implemented methods only modify the current configuration of the channels but does not allow to add or remove them. At DESY we have implemented some macros for adding/removing counters/timers, so we can live with that, but perhaps it could be also added here.

rhomspuron:
Hi @teresanunez, this is a good idea, we did not implement  methods to add/remove channels of the MG, we can implement them.

droldan:
Hi @teresanunez , 
Yes, we could these method to add/remove counters/timers, should be easy to do it.

Maybe, we could wait until the pull request will be accepted and create a new pull request with the add/remote counters/timers methods.

In the other hand, we should to create a new issue to adapt the ExpConf to use this new API and for this is necessary create missing methods.

Method available in this pull request:

- getEnabledChannels
- setEnabledChannels
- getOutputChannels
- setOutputChannels
- getPlotTypeChannels
- setPlotTypeChannels
- getPlotAxesChannels
- setPlotAxesChannels
- getCtrlsTimer
- setCtrlsTimer
- getCtrlsMonitor
- setCtrlsMonitor
- getCtrlsSynchronization
- setCtrlsSynchronization
- getCtrlsSynchronizer
- setCtrlsSynchronizer

Missing methods to full management:

- getConditionChannels
- setConditionChannels
- getNormalizationchannels
- setNormalizationchannels
- getNexusPathChannel
- setNexusPathChannel

- addChannels
- removechannels

teresanunez:
Thanks @droldan and @rhomspuron. Yes, I agree, one can accept this PR first and create a second one with the missing methods. Many thanks.

teresanunez:
I have been testing this new API and IMO it can be merged.

dschick:
hej, I would like to finish the PR #824 which should use the new MHConfiguration API.
Can anyone resolve the conflicts and merge this PR please?

Thanks

Daniel

reszelaz:
I'll take care of updating this branch with the recent develop changes. But I remember that there was some design idea to be clarified still before merging back into develop. Tomorrow I will try to recall what was that..

reszelaz:
Thanks for this PR! It is already functional and works quite well (I think that the use_fullname keyword argument does not work correctly, at least when I use the `getCtrlsSynchronization` method. But before intergrating, I propose a little changes in the API (the internal code would be pretty much reusable, this is just in order to provide IMHO clearer API).

**First, the type of the API which applies to configuration parameters set on the channel level.**

```python
def setOutput(self, output, *elements, apply=True):
    """Set the output configuration for the given elements.
    
    Channels and controllers are accepted as elements. Setting the output
    on the controller means setting it to all channels of this controller
    present in this measurement group.
    
    Configuration by default is directly applied on the server. Since setting
    the configuration means passing to the server all the configuration 
    paramters of the measurement group at once this behavior can be 
    changed with the *apply* argument and we can keep the configuration
    changes only locally. This is useful when we want to change more then 
    one parameter, in this case only the setting of the last parameter should
    use `apply=True`.
    
    :param output: `True` - output enabled, `False` - output disabled
    :type output: bool
    :param elements: sequence of element names or full names, no elements 
        means set to all
    :type elements: list(str)
    :param apply: `True` - apply on the server, `False` - do not apply yet 
        on the server and keep locally (default: `True`)
    :type apply: bool
    """

def getOutput(self, *elements, ret_full_name=False):
    """Get the output configuration of the given elements.
    
    Channels and controllers are accepted as elements. Getting the output
    from the controller means getting it from all channels of this controller
    present in this measurement group.
    
    :param elements: sequence of element names or full names, no elements 
        means get from all
    :type elements: list(str)
    :param ret_full_name: whether keys in the returned dictionary are
        full names or names (default: `False` means return names)
    :type ret_full_name: bool
    :return: ordered dictionary where keys are **channel** names (or full 
        names if `ret_full_name=True`) and values are their output
        configurations. Note that even if the *elements* contained
        controllers, the returned configuration will always contain
        only channels.
    :rtype: dict(str, bool)
    """
```

I also attach some example of usage (could be later included in the docsting of the class):

1. Enable output on channels "ct01" and "ct02" of "mntgrp01".

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
meas_grp.setOutput(True, "ct01", "ct02")
```
2. Enable output **only** on channels "ct01" and "ct02" of "mntgrp01" and ensure it is disabled on the others.

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
meas_grp.setOutput(False, apply=False)
meas_grp.setOutput(True, "ct01", "ct02")
```

3. Print output configuration of all channels of "mntgrp01":
```python
meas_grp = self.getMeasurementGroup("mntgrp01")
out_conf = meas_grp.getOutput()
for channel, output in out_conf.items():
    self.print("{0} - {1}".format(channel, output))
```

Almost the same API applies to the following configuration paramters:
* enabled
* plot_type
* plot_axes
* value_ref_pattern
* value_ref_enabled

The last two were missing in the PR but after SEP2 are possible.

**Second, the type of the API which applies to configuration parameters currently set on the controller level.**

```python
def setSynchronization(self, synchronization, *elements, apply=True):
    """Set the synchronization configuration for the given elements.
    
    Channels and controllers are accepted as elements. Setting the     
    synchronizer on the channel means setting it also to all the other 
    channels proceeding from the same controller present in this measurement 
    group.
    
    Configuration by default is directly applied on the server. Since setting
    the configuration means passing to the server all the configuration 
    paramters of the measurement group at once this behavior can be 
    changed with the *apply* argument and we can keep the configuration
    changes only locally. This is useful when we want to change more then
    one parameter, in this case only the setting of the last parameter should
    use `apply=True`.
    
    :param synchronization: synchronization type to be used during 
        acquisition
    :type synchronization: :class:`~sardana.pool.AcqSynchType`
    :param elements: sequence of element names or full names, no elements 
        means set to all
    :type elements: list(str)
    :param apply: `True` - apply on the server, `False` - do not apply yet 
        on the server and keep locally (default: `True`)
    :type apply: bool
    """

def getSynchronization(self, *elements, ret_full_name=False,
                       ret_by_ctrl=False):
    """Get the synchronization configuration of the given elements.
    
    Channels and controllers are accepted as elements. Getting the output
    from the controller means getting it from all channels of this controller
    present in this measurement group, unless `ret_by_ctrl=True`.

    :param elements: sequence of element names or full names, no elements 
        means get from all
    :type elements: list(str)
    :param ret_full_name: whether keys in the returned dictionary are
        full names or names (default: `False` means return names)
    :type ret_full_name: bool
    :param ret_by_ctrl: whether keys in the returned dictionary are
        controllers or channels (default: `False` means return channels)
    :type ret_by_ctrl: bool
    :return: ordered dictionary where keys are **channel** names (or full 
        names if `ret_full_name=True`) and values are their synchronization  
        configurations
    :rtype: dict(str, :class:`~sardana.pool.AcqSynchType`)
    """
```

Almost the same API applies to the following configuration paramters:
* timer
* monitor
* synchronizer

I also attach some example of usage (could be later included in the docsting of the class):

1. Set synchronization to `AcqSynchType.Trigger` for "ct01" and "ct02" (belong to the same controller "ctctrl01" of "mntgrp01".

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
meas_grp.setSynchronizer(AcqSynchType.Trigger, "ct01", "ct01")
```

It is also possible to set it directly on the controller:

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
meas_grp.setSynchronizer(AcqSynchType.Trigger, "ctctrl01")
```

2. Print synchronization configuration of all channels of "mntgrp01":
```python
meas_grp = self.getMeasurementGroup("mntgrp01")
synch_conf = meas_grp.getSynchronization()
for channel, synch in synch_conf.items():
    self.print("{0} - {1}".format(channel, synch))
```

3. Print synchronization configuration of all controllers of "mntgrp01":

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
synch_conf = meas_grp.getSynchronization(ret_by_ctrl=True)
for ctrl, synch in synch_conf.items():
    self.print("{0} - {1}".format(ctrl, synch))
```

In the second group of methods, which applies to configuration parameters currently only possible on the controller level, the `ret_by_ctrl` keyword argument of the getters may be a little bit confusing.
The getters accepts both channels and controllers, even if currently the configuration is 
only possible on the controller level. IMO the following argumets are in favor of this API: 
* users mostly operate on channels and not on the controllers
* expconf visualizes these configuration parameters per channel
* the default return value is coherent with the other group of methods (configuration parameters on the channel level)
* if in the future we add the feature to allow to configure these parameters on the channel level the API would be stil valid)

So the `ret_by_ctrl=True` allows to obtain the configuration per controller, as seen in the above example 3.

reszelaz:
I encourage you to post your opinion about the API proposed in https://github.com/sardana-org/sardana/pull/867#issuecomment-504478326. Many thanks in advance!

reszelaz:
I have just realized the the proposal from https://github.com/sardana-org/sardana/pull/867#issuecomment-504478326 has a small conflict with the already existing methods for the measurement group's Tango attribute `Synchronization`. So, on the measurement group Taurus extensions we have the following methods:
* `getSynchronizationObj`
* `getSynchronization`
* `setSynchronization`

The above proposal also plans to add the `getSynchronization` and `setSynchronization` methods for another purpose and here is the conflict.

I propose that we rename the Tango attribute `Synchronization` to `SynchronizationPlan`. Then on the Taurus extension to the measurement group we will have:

* `getSynchronizationPlanObj`
* `getSynchronizationPlan`
* `setSynchronizationPlan`
* `getSynchronization`
* `setSynchronization`

Another alternative name, a little bit longer, for this attribute is `SynchronizationDescription` and this would give the following names:

* `getSynchronizationDescriptionObj`
* `getSynchronizationDescription`
* `setSynchronizationDescription`
* `getSynchronization`
* `setSynchronization`


I think it should be fine to introduce this change due to the following reasons:
* I doubt that there are many usages of this methods outside of Sardana code itself - this could be used if you program your own synchronized measurements different then scans.
* It is relativelly new attribute (introduced with SEP6)
* We are going to break the backwards compatibility in the upcoming release anyway so it may be a good moment.

It would be great if you could express your preferences about the name (or propose a different one) and if you are agains this renaming also raise your hand now:)

teresanunez:
Hi @reszelaz, I agree with renaming the attribute of the MG. Perhaps de name SynchronizationDescription is more clear ... but IMO both could be fine.

rhomspuron:
Hi I agree too with the name change of the synchronization to avoid conflicts. Both cases are fine for me, but I prefer SynchronizationPlan because is shorter :).

droldan:
Dear,
IMO (out of the scope), I understand what  getSynchronizationDescription means,
 unlike  getSynchronizationPlan that is more complicated to understand what
it means.
But well, both are fine!

rhomspuron:
I did not implement the renaming of the Synchronization to SynchronizationDescription. We can do it in another PR. For that  I do not implement the method to set/get the synchronization, however it is possible to do it by the configuration:
```
mg.getConfiguration()._setCtrlsSynchronization(self, synchronization, ctrls=None, apply_cfg=True)
mg.getConfiguration()._getCtrlsSynchronizer(self, ctrls=None, use_fullname=False)

```
When we implement the PR to change the name, we should export this two method to the MeasurementGroup API

reszelaz:
While working with @ByRellex on the tests for this PR, I stopped once again to see if we could improve the API even more. Below I explain you the excercise which I did but, to whom is not interested in all the details I anticipate that IMO the current proposal is a good trade-off between the API-friendliness and the cost of implementation.

So, I thought of making the getter methods e.g. `getOutput`, `getEnabled` return `BaseSardanaElement` objects. This way we could remove the `ret_full_name` kwarg (this objects already have `name` and `full_name` members). 

An example of usage could be to print all channel names and their output configurations:

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
out_conf = meas_grp.getOutput()
for channel, output in out_conf.items():
    self.print("{0} - {1}".format(channel.name, output))
```

Another example of usage could be to print all channel full names and their output configurations:

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
out_conf = meas_grp.getOutput()
for channel, output in out_conf.items():
    self.print("{0} - {1}".format(channel.full_name, output))
```

This change would make code more complicated when searching in the returned dictionary based on the name:

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
out_conf = meas_grp.getOutput()
channel_name = "ct01"
for channel, output in out_conf.items():
    if channel.name == channel_name:
        self.print(output))
```

instead of (as with the current proposal):

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
out_conf = meas_grp.getOutput()
channel_name = "ct01"
self.print(out_conf[channel_name]))
```

but I think this won't be a typical use of this API. If one knows the name beforehand it is better to use it as the input argument of the getter:

```python
meas_grp = self.getMeasurementGroup("mntgrp01")
channel_name = "ct01"
out_conf = meas_grp.getOutput(channel_name)
self.print(list(out_conf.values())[0])
```

Similarly we could think of extending the input API and allow to pass `BaseSardanaElement` objects as `elements`.

Up to here this propsal looks very interesting - we get rid of `ret_full_name` kwarg and we get returned objects. But what if we have external attribute in the measurement group? We don't have a class which represent the external attributes e.g. Tango attribute. We would need to create one and fill its objects with at least name and full_name. Actually the more elegant solution would be to have a common configuration class for all channels (proceeding from Sardana or the external ones) and return always objects of this class. And here I thought that this will require a big design process followed by a refactoring. I think it is better for the project to integrate this API, which at the end is pretty alright.
Well, this was just to share with you these thoughts, if someone has an opinion about it we would be glad to hear it! Many thanks!

rhomspuron:
@reszelaz  I totally agree with:

> Actually the more elegant solution would be to have a common configuration class for all channels (proceeding from Sardana or the external ones) and return always objects of this class. And here I thought that this will require a big design process followed by a refactoring. I think it is better for the project to integrate this API, which at the end is pretty alright.

When we start to use more this API, we can evaluate better the way to use it and improve it. 

I will check the errors reported by @ByRellex.

ByRellex:
Updated the tests and added more complex tests. Found two main problems while running the new tests:
- [ ] Using get method with controller name returns unordered dict. See test_PlotAxes.
- [ ] Methods that contain get_by_ctrl flag return controller ret_full_name while get_by_ctrl=False and ret_full_name=True, this should return channels full names.

Added TODOs on the test code areas where these problems happen.

rhomspuron:
Thanks @ByRellex, I will check it. 

On the other hang, after to add protections for the referable parameters, I modified the expconf to not generate exception. Now the server is responsible to configure the this defaults values for the channels, it generates a pop up message in expconf after to apply the configuration when is added a new 1D or 2D channel on the measurement group. The same situation is happened when is added a tango attribute now. The issue #1046 should solve it.

reszelaz:
I just pushed three commits directly to develop. These are:
* 1ac1bc0 - fix notes in the docstrings
* 28940b86 - remove duplicated info in the docstrings
* efd2ea4 - re-introduce accidentally removed method in 28940b86

PR #871, opened 2018-07-17
Fix incorrect conversion to boolean

rhomspuron:
The boolean type is not protected and it does not allow to use 0/1 as values. 

Macro to test it:
```python
class test_boolean(Macro):
    param_def = [['val', Type.Boolean, None, 'value']]

    def run(self, value):
        self.info(value)

````


teresanunez:
I have reviewed this and IMO it can be merged.

reszelaz:
Thanks to both! I have also review the extra commit that implemented the macro completer in spock.
Everything ok. Merging..

PR #872, opened 2018-07-18
Fix Controller taurus extension

reszelaz:
Fix two methods in Controller taurus extension:
* `getElementByAxis`
* `getElementByName`


teresanunez:
IMO it can be merged.

reszelaz:
Thanks! Mergin it..

Issue #833, opened 2018-07-10
reconfig implementation


dschick:
Very night we are switching off some of our controllers but not necessarily the spock and Sardana terminals.
So currently I need to restart Sardana after I reconnected/restarted the hardware in the morning.

In SPEC there is a reconfig command which reconfigures all the hardware which would be equivalent to a restart of the PoolSever from within spock I guess.

reszelaz:
Just to add some information gathered some time ago by @guifrecuni:

From spec documentation [understanding errors](https://www.certif.com/guide/Solving.html#understanding-errors):

> reconfig not only will close/open connection to devices but it will also execute all the initialization sequences as it will in many ways constiuture the equivalent of a full reset

...and a guide what to do when [motor does not move](https://www.certif.com/guide/Solving.html#my-motor-does-not-move).

reszelaz:
I think that this feature should reinitialize the controller and/or elements, basically call controller constructor and call the `DeleteDevice` and `AddDevice` methods for the given axes. Furthermore it should restore values of some attributes, I propose the memorized ones. There should be two flavors of `reconfig` - the macro could be parametrized. The difference comes from the values to be restored. In some cases it would be enough to restore the last set values, let’s call this *reconfig last*. In other cases it would be better to restore values from a snapshot done at some point, let’s call this *reconfig snap*. For the second one we would need another macro, let’s call it `saveconfig`. This macro would save the set values of the given attributes to some kind of persistence medium, it could be a file in a configurable directory.

I already implemented the *reconfig last* option. It was very easy, it just calls the Tango `Init` command of the object to be reconfigured. In case we want to reconfigure a controller it also iterates over its elements and call the `Init` command for them as well. If someone is interested in this I can make a PR.

dschick:
Please do the PR.

Many thanks

reszelaz <notifications@github.com> schrieb am Mi., 18. Juli 2018, 11:03:

reszelaz:
#873 adds this macro. Any feedback is more than welcome:)

PR #873, opened 2018-07-18
Add reconfig macro

reszelaz:
This adds #833. It requires #872 to work correctly.
Please could you test it and feedback.

### DIFF
--- a/src/sardana/macroserver/macros/expert.py
+++ b/src/sardana/macroserver/macros/expert.py
@@ -541,3 +541,30 @@ class sar_info(Macro):
         self.output("-----------")
         for line in table.genOutput():
             self.output(line)
+
+
+class reconfig(Macro):
+    """Reconfigure a pool element.
+
+    This implies its full initialization which includes re-write of its
+    memorized attributes with the last set values.
+
+    Controller initialization implies initialization of all its elements."""
+
+    param_def = [
+        ['obj', Type.PoolElement, None, 'Pool element to reconfigure']
+    ]
+
+    def run(self, obj):
+        type_ = obj.getType().lower()
+        if type_ == "instrument":
+            raise ValueError("Instruments can not be reconfigured")
+        obj.Init()
+        self.info("%s was reconfigured" % obj.getName())
+        if type_ == "controller":
+            ctrl = obj
+            used_axes = ctrl.getUsedAxes()
+            for axis in used_axes:
+                elem = ctrl.getElementByAxis(axis)
+                elem.Init()
+                self.info("%s was reconfigured" % elem.getName())
--- a/src/sardana/taurus/core/tango/sardana/pool.py
+++ b/src/sardana/taurus/core/tango/sardana/pool.py
@@ -612,7 +612,8 @@ class Controller(PoolElement):
     def getElementByAxis(self, axis):
         pool = self.getPoolObj()
         for _, elem in pool.getElementsOfType(self.getMainType()).items():
-            if elem.controller != self.getName() or elem.getAxis() != axis:
+            if (elem.controller != self.getFullName() or
+                    elem.getAxis() != axis):
                 continue
             return elem
 
--- a/src/sardana/taurus/core/tango/sardana/pool.py
+++ b/src/sardana/taurus/core/tango/sardana/pool.py
@@ -619,8 +619,9 @@ class Controller(PoolElement):
 
     def getElementByName(self, name):
         pool = self.getPoolObj()
-        for name, elem in pool.getElementsOfType(self.getMainType()).items():
-            if elem.controller != self.getName() or elem.getName() != name:
+        for _, elem in pool.getElementsOfType(self.getMainType()).items():
+            if (elem.controller != self.getFullName() or
+                    elem.getName() != name):
                 continue
             return elem
 


teresanunez:
HI @reszelaz, the argument of the macro can only be the alias name?

reszelaz:
@teresanunez many thanks for your feedback!

I used the `PoolElement` parameter type. This includes motors, experimental channels, controllers, measurement groups, instruments, etc. You refer to them with names (and indeed in Tango this are aliases at the same time). 

Were you thinkin in a different way of passsing parameters?

Actually I was thinking in defining one repeat paramter with `min=0`. This would allow to reconfig more elements at once. When no parameter values are passed this could reconfg all the elements. I don't know if this is interesting? For sure the implementation would be quite more difficult...

teresanunez:
Hi @reszelaz, thanks, it is ok for me to pass the parameters like this, but perhaps we could add some
 note about the 'name' that has to be used (sardana element name, tango alias). I was first trying with
 the device name and getting an error.
I am not sure that the reconfig of all the elements at once would be useful, I think it could be a risk.
I think in case of being implemented one should rise a confirmation question ...
But well, let's see what the other think about it.

reszelaz:
>but perhaps we could add some note about the 'name' that has to be used

I see your point, I will describe it as following: "Pool element (or its name) to reconfigure"

Actually it is not a name (string) but a PoolElement object - see the parameter type. The tab completion in spock could be helpful here. But it is just that the clients (spock, macroexecutor, ...) always pass strings to the server which later on are docoded to the objects. When using this macro with `Macro.execMacro` method you could actually pass the PoolElement object e.g. Motor object and not its name.

> I am not sure that the reconfig of all the elements at once would be useful, I think it could be a risk.
I think in case of being implemented one should rise a confirmation question ...

I agree with you that it may be risky. For me there is no need to have this possibility if noone see a benefit of this.

teresanunez:
Thanks, @reszelaz, it is perfect for me if you add the '(or its name)' ... but yes, you are right, it is actually
not the name.

dschick:
Hej @reszelaz & @teresanunez , thanks very much for pushing this!
It is actually cool feature to only reconfigure/restart a single pool element by its "name".
This is actually a drawback of SPEC's reconfigure, that one always have to reconfigure everything.

However, this reconfig all is also quite helpfull an many case.
Most often at the beamline if there is any problem in SPEC with driving motors or acquiring data, you just do a reconfig and normally you are done.
If there is really a controller down, the reconfig macro will give an error message, so in the Sardana case it should somehow print some log messages from the pool into the spock CLI.
So this reconfig all normally saves you to go through all devices manually if you do not know, where the problem is (think of a netwokr issue, power cut-off, etc.) Especially users do not even know about all the available names of the controllers...

But I of course agree that regonfiguring everything might be dangerous especially if some controllers have more complicated startup procedures. I guess this also applies for situations where spock is not running on the same machine as the pool and/or macro servers and the users (should not) have no access to restart both of them.

Do you have pools which are actually used by more than one client at the same time?

reszelaz:
Thanks for the feedback @dschick!

> If there is really a controller down, the reconfig macro will give an error message, so in the Sardana case it should somehow print some log messages from the pool into the spock CLI.

It would be great if you could test how the current macro behaves in this situation.

> So this reconfig all normally saves you to go through all devices manually if you do not know, where the problem is (think of a netwokr issue, power cut-off, etc.) Especially users do not even know about all the available names of the controllers...

Yes, I was also thinking about a similar scenario. In this case the `reconfig` macro should be very similar to restarting the server itself. So this way it would save the user the process of openning the astor and make some clicks...

> Do you have pools which are actually used by more than one client at the same time?

Yes, we do. Maybe it is no the most common architecture but we have a couple of them where different spock sessions may be running in parallel doing totally unrelated tasks. In this case `reconfig all` could be very dangerous.

dschick:
I just tested the `reconfig` macro.
I guess I need to implement a warning or error message inside the controller logic in order to catch a missing connection to the controller, right?

However, somehow the `reconfig` just worked partly:
1. I started Sardana and spock with on of the controllers switched off.
2. I did a` reconfig controller `and got no error message (although the controller was still off) but this is also not implemented yet in the controller class, see above.
3. I switched the controller on and did another `reconfig`.
4. I got some messages in the Sardana pool/macro server terminal which proved the actual reconnection via gpib to the controller
5. however, the motor of the controller was still not readable nor writable from within spock

teresanunez:
This kind of issues (except to point 5. that seems to be something general) are the reason why I 
don't like the idea of a 'reconfig all', one has to really take care how your controller and your hardware will behave in such cases. Make it in a general case with all controllers could cause problems that you would not expect, and for the controllers you know you want to have it you would have to really investigate them one by one. Well, it is only my opinion.

rhomspuron:
>   Actually I was thinking in defining one repeat parameter with min=0. This would allow to reconfig more elements at once. When no parameter values are passed this could reconfg all the elements. I don't know if this is interesting? For sure the implementation would be quite more difficult...

This is a good idea. IMHO, after a reconfig, the system should work with the default configuration, and should show which element will not do it.

reszelaz:
Thanks to all for your comments!
Thanks @dschick for testing it!

> I guess I need to implement a warning or error message inside the controller logic in order to catch a missing connection to the controller, right?

No, passing these kind of logging information between the Pool and MacroServer is not yet possible. But I see it very interesting from the point of view of debugging problems with the controllers.

For the moment I have modified the macro so it prints this information from the controller status. Below some exemplary macro output, first in the case of the successful reconfig and later in the case of failure (note that `sfmotctrl01` has one motor `sfmot01`, this one is reconfigured only if the controller reconfiguration succeeded):

```
Door_zreszela_1 [62]: reconfig sfmotctrl01
sfmotctrl01 was successfully reconfigured
sfmot01 was successfully reconfigured

Door_zreszela_1 [63]: reconfig sfmotctrl01
Failure while reconfiguring sfmotctrl01
sfmotctrl01:
       State: Fault
      Status: Traceback (most recent call last):
                File "/home/zreszela/workspace/sardana/src/sardana/pool/poolcontroller.py", line 346, in _init
                  self._ctrl = self._create_controller()
                File "/home/zreszela/workspace/sardana/src/sardana/pool/poolcontroller.py", line 337, in _create_controller
                  ctrl = klass(name, props, *args, **kwargs)
                File "/home/zreszela/workspace/sardana/doc/source/devel/howto_controllers/sf_motor_ctrl.py", line 109, in __init__
                  raise Exception("Constructor exception")
              Exception: Constructor exception
```

This should already give you the communication error messages.

> 5. however, the motor of the controller was still not readable nor writable from within spock

I was able to reproduce it. I will need more time to fix this.

dschick:
hi  @reszelaz I confirm that the communication error message works as expected

```
%reconfig kepcoCtrl
Failure while reconfiguring kepcoCtrl
kepcoCtrl:
       State: Fault
      Status: Traceback (most recent call last):
                File "/home/moke/Repos/sardana/src/sardana/pool/poolcontroller.py", line 346, in _init
                  self._ctrl = self._create_controller()
                File "/home/moke/Repos/sardana/src/sardana/pool/poolcontroller.py", line 337, in _create_controller
                  ctrl = klass(name, props, *args, **kwargs)
                File "/home/moke/Repos/sardanaControllers/kepcoController.py", line 46, in __init__
                  print self.inst.query('*IDN?')
                File "/usr/local/lib/python2.7/dist-packages/pyvisa/resources/messagebased.py", line 558, in query
                  self.write(message)
                File "/usr/local/lib/python2.7/dist-packages/pyvisa/resources/messagebased.py", line 223, in write
                  count = self.write_raw(message.encode(enco))
                File "/usr/local/lib/python2.7/dist-packages/pyvisa/resources/messagebased.py", line 201, in write_raw
                  return self.visalib.write(self.session, message)
                File "/home/moke/Repos/pyvisa-py/pyvisa-py/highlevel.py", line 346, in write
                  raise errors.VisaIOError(ret[1])
              VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.

```
 and vanishes when I switch on the controller.#
 ```
%reconfig kepcoCtrl
kepcoCtrl was successfully reconfigured
Failure while reconfiguring kepco
kepco:
       State: Fault
      Status: kepco is in Fault
              TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
    Position: position attribute not found

```

As you mentioned, the motor stays in Fault state. So my controller is obviously named `kepcoCtrl` and the motor is `kepco`

Thanks for the efforts

reszelaz:
>> however, the motor of the controller was still not readable nor writable from within spock

> I was able to reproduce it. I will need more time to fix this.

I've investigated this a little bit and it is not so straightforward. It has to do with the dynamic attributes initialization. This one is done by `tango.DeviceClass.dyn_attr` implementation. I will come back to this after the Jul18 release.

PR #874, opened 2018-07-20
Configure ORBendpoint omniORB argument from free properties

droldan:
Allow to configure the ORBendPoint omniORB parameter from tango DB free properties. 
The priority order is:
1. Arguments
2. Environment Variable
3. Tango DB free property

### DIFF
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -1228,7 +1228,7 @@ def run(prepare_func, args=None, tango_util=None, start_time=None, mode=None,
     except KeyboardInterrupt:
         pass
 
-    log_messages.extend(prepare_ORBendPoint(args))
+    log_messages.extend(prepare_ORBendPoint(args, tango_args))
     # Tango versions < 8.0.5 are affected by
     # https://sourceforge.net/p/tango-cs/bugs/495/
     tango_version = get_tango_version_number()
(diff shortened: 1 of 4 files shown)


src/sardana/tango/core/util.py
```diff
@@ -620,6 +675,13 @@ def prepare_environment(args, tango_args, ORB_args):
             env_name = arg[1:]
             env_val = ORB_args[i + 1]
             os.environ[env_name] = env_val
+
+            # In Tango > 7 the omniORB parameters has a conflict if are
```
reszelaz: This workaround is too agressive. It removes all ORB arguments.
Seems like from Tango 8 we don't need this workaround, see [tango-cs#495](https://sourceforge.net/p/tango-cs/bugs/495/)

reszelaz:
@droldan I addes a new chapter to in the User's guide: Server configuration and explained this configuration possibilities there. I'm merging this one. Many thanks!

reszelaz:
I just pushed dd869c7 directly to develop cause there were problems the the free property was not set and the server was running with python 2.6,

reszelaz:
I just had to change the order when this preparation is done and add some protection when the server instance name is unknown (I also protected the logstash preparation).

This is because there are two special cases of the server startups, one without the instance name and the other with `-?` option e.g. 
```
$> Pool
$> Pool -?
```
I pushed it directly to develop.

PR #875, opened 2018-07-20
Docs: ViewOptions

teresanunez:


reszelaz:
I have just changed some formatting to be more clear (some bulletpoints, newlines, etc..).
I'm integrating this. One idea for future would be to try to maintain only one list of view options (instead of three), maybe in `setvo` macro only?

teresanunez:
HI @reszelaz, many thanks, I also think that it would be better to maintain only one list ... but I didn't want to put the whole explanation of each variable inside of the setvo macro ... but well, no problem if you want to have it there.

reszelaz:
I see your point. If we don't want to have a long macro docstring in this case we could document the list of options just in the spock documentation refering from the `setvo` to that chapter.

I'm just affraid that havind three lists adds an overhead and could possible lead to inconsistencies between them..

teresanunez:
The problem that I see if you have the list only in the documentation and not in the macros is that
the users have to go to the docu to see the name of the variables, they can not see them directly
in Spock. I am not sure if they would prefer to have it in the docstring. Any opinion about that from other people?.

PR #876, opened 2018-07-20
Optional param

rhomspuron:
Implement optional parameter for the macros. Adds #285.

### DIFF
--- a/doc/source/devel/howto_macros/macros_general.rst
+++ b/doc/source/devel/howto_macros/macros_general.rst
@@ -300,7 +300,7 @@ being a composed of four elements:
     - parameter type
     - parameter default value:
         - ``None`` means no default value
-        - ``OptionalParam`` means that
+        - ``Optional`` means that
           :ref:`the parameter value is optional <sardana-macro-optional-parameters>`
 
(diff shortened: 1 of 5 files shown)


rhomspuron:
Hi @dschick, thanks for test it!!! 

I just tested with the next code:
```python
class ch(Macro):
    param_def = [['ch', Type.ExpChannel, OptionalParam, 'name of counter']]

    def run(self, ch):
        if ch is OptionalParam:
            ch = self.getExpChannel('ct29')
        self.info(ch)
``` 
And it worked. Output:
```python
Door_rhoms_1 [4]: ch
ct29

Door_rhoms_1 [5]: ch ct30
ct30
```

Could you send your macro?

dschick:
works like a charm, my if statement was wrong like
`if counter:`
instead of
`if counter is OptionalParam`

Thanks so much

Daniel

rhomspuron:
Hi @dschick 

After seeing your code, we think it's more intuitive and pythonic to pass a None instead of OptionalParam. I will change the code.

dschick:
gracias! I guess this is more convinent

PR #878, opened 2018-07-23
Docu: users macro logging

teresanunez:


teresanunez:
The report feature is also added in this PR. 
I was very short in the description since it was already described in the developers part, I put mainly links for not having the same information in different places.

doc/source/users/environment_variable_catalog.rst
```diff
@@ -112,6 +112,19 @@ LogMacroFormat
 
 Definition of the format to be used to save the file.
 
+.. _logmacrofilter:
+
+LogMacroFilter
```
reszelaz: This is not an environment variable but sardana custom setting. I would put this documentation as a docstring sardanacustomsettings module.

doc/source/users/index.rst
```diff
@@ -16,6 +16,8 @@ User's Guide
     Adding real elements <adding_elements>
     Macro Hooks <macro_hooks>
     Macro Environment <macro_environment>
+    Macro Logging <macro_logging>
```
reszelaz: I have no strong opinion about this. But maybe a single chapter describing both features would be enough? At the end they are very similar.

doc/source/users/macro_logging.rst
```diff
@@ -0,0 +1,21 @@
...
+.. currentmodule:: sardana.macroserver.macros.logging
+
+
+.. _sardana-macros-logging:
+
+=============
+Macro Logging
+=============
+
+The MacroServer allows to write into a log file messages from the class responsible for
```
reszelaz: This part about "the class responsible running for running the macros" is too specific for the user. Especially as the first sentence describing this feature. I would eventually mention something about additional log entries that enclose each macro execution e.g. when it starts and when it ends. What do you think?

doc/source/users/macro_report.rst
```diff
@@ -0,0 +1,14 @@
...
+.. currentmodule:: sardana.macroserver.macros.logging
+
+
+.. _sardana-macros-reports:
+
+=============
+Macro Reports
+=============
+
+:ref:`Report messages <sardana-macro-reporting>` can be send by the macros to a report file. This report feature is configured as described :ref:`here <sardana-configuration-macroserver>`.
```
reszelaz: We should [avoid hyperlinks like *here*](http://www.writethedocs.org/guide/writing/docs-principles/#skimmable)

doc/source/users/environment_variable_catalog.rst
reszelaz: You can reference a sardana custom setting with [data Python domain markup](http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py:data) (see [example](https://sardana-controls.org/users/configuration/spock.html))

teresanunez: Sorry, it is true ... I implemented and I forgot about it !

teresanunez:
Sorry, I have some problems with this branch, I create a new PR.

Issue #825, opened 2018-07-09
Docs: Taurus extensions are not documented


reszelaz:
Taurus extensions are very important part of the Sardana API. For example these are interfaces to the Sardana objects when used wihin macros. We should autodocument this part of the API.

Furthermore we could add a short section to "Writing macros" chapter. It could be called "How to access to pool elements". This section should explain about the possible getters e.g. getMotor, getController, about the returned objects i.e. Taurus extensions. and finally point to the autogenerated API

teresanunez:
Continue in PR #881

PR #881, opened 2018-07-27
[WIP] Docu: Taurusextensions

teresanunez:
Implements issue #825. Up to now only the autodocumentation for the taurus extensions is implented.
Still missed the chapter explaining how to access pool elements in the macros.


teresanunez:
Hi @reszelaz, actually the access of the pool elements inside the macro is already partially described in the chapter 'Macro Context'. Should this be repeated?. Perhaps one could add a chapter called
'Accessing Pool elements' (keeping it consistent with the title of the chapter 'Accessing Environment',
and there point out to the 'Macro Context' chapter ... Do you want to add more descriptions, details,
examples?. At the end of the chapter one could add the link to the Taurus Extensions API, even if
 this could be too much for the users.

reszelaz:
Thanks! This looks very good!

> access of the pool elements inside the macro is already partially described in the chapter 'Macro Context'. Should this be repeated?

I think this chapter is enough. I have moved the second sentence from the "Accessing Pool elements" there and I hve removed the new chapter.

Also I made the link point to the `sardana.taurus.core.tango.sardana` which should also contain the `macroserver` module since the chapter talks about the sardana elements and not pool elements only.

I have also added the inheritance diagrams.

I have removed the `ComChannel` class from the API which is not ready (it is not implemented on the Pool side). I will create an issue for that.

PR #885, opened 2018-07-31
Allows hooks in timescan

reszelaz:
Expose allowsHooks hint of timescan macro.
Implement pre-acq and post-acq hooks in TScan class.
@sardana-org/integrators can you take a look, please?


teresanunez:
IMO it can be merged

PR #619, opened 2017-09-28
Fix taurusgui with instruments (instead of #612)

reszelaz:
This PR solves #122 and substitutes PR #612. More information in the said PR.

PR #887, opened 2018-08-01
Docu: Instrument

teresanunez:
@reszelaz, i am sorry, I do not find in the documentation de [Howto-GUI_creation] where you
said that the configuration of the instruments is described. This part is missed in the PR, if you
tell me where to find it (for using a link to that) I will add it. Thanks.

### DIFF
--- a/doc/source/devel/overview/index.rst
+++ b/doc/source/devel/overview/index.rst
@@ -22,4 +22,5 @@ Overview
     2D experiment channel <overview_2D>
     Pseudo counter <overview_pseudocounter>
     Measurement Group <overview_measurementgroup>
+    Instrument <overview_instrument>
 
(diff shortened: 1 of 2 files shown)


reszelaz:
Many thanks!

I've not built it yet but the overview looks good to me.

The link mentioned in the issue is this one: https://sourceforge.net/p/sardana/wiki/Howto-GUI_creation

Could you crosscheck if the last paragraph is right? #619 should had fixed it? At least partially.

teresanunez:
Thanks @reszelaz. I took out the last paragraph, I had not realized that it was solved. I put the way how
to configure it, I have not used the link because it doesn't below to the docu and also because there
is still mentioned that Sardana has to be restarted after configuring the instrument. I have also not
 given too many details how to create them because it is based on Tango devices ... so I put a general
 sentence and them the use of tango devices is as an example. Well, you can have a look, I can go
 deeper is needed. Many thanks.

PR #888, opened 2018-08-01
genv macro

dschick:
get env variable


teresanunez:
Hi @dschick, the way you use for reading the environment if it belongs to a macro or to a door is 
wrong (I tested it and it doesn's work), have a look at the macro lsenv, there you see how to do it.

dschick:
@teresanunez thanks for the advise.
Is this correct way of doing it?
How about the lsenv macro? what means there the macro_list input?

teresanunez:
Hi @dschick, it is still not right, it should be something like this:

def run(self, var):
        pars = var.split(".")
        if len(pars) == 1:
            env = self.getAllDoorEnv()
            line = '%s = %s' % (str(var), env[var])
        else:
            env = self.getEnv(key=None, macro_name=pars[0])
            names_list = list(env.keys())
            names_list.sort(key=str.lower)
            line = '%s = %s' % (str(var), str(env[pars[1]]))
        self.output(line)
    

The macro_list input is the list of the macros for which the environment variables would be displayed.
The environment variables can be defined as global or for an specific door or macro.

dschick:
@teresanunez thanks again, hope everything is fine now.

I was wondering if one should prettifiy the output e.g. of dicts.
There is a json.dump method available doing that.

teresanunez:
Hi @dschick, thanks for the changes. I put only the example for cases of the type macro_name.variable
 (or door_name.variable) but it could be also cases like door_name.macro_name.variable, that are
 still not taken into account. The macro should also be extended for those cases.
About the output, you can put it nicer if you prefer, no problem at all.

dschick:
Hi @teresanunez, how about handing the Door and/or Macro names as Optional parameters to the macro?
I will change the PR accrodingly.

dschick:
Hi @teresanunez the macro is now accepting also Door and/or Macro names.
I have not prettified the output, because it would be easiest to use some existing libs for that, such as `json.dump` but I fear this would be an extra dependency for Sardana.
Otherwise it is a bit of overkill to catch all possible cases for pretty printing the output, but maybe anyone has a simple solution for that.

reszelaz:
Hi @dschick and @teresanunez,
I have tested the macro and it works perfectly. I made a small change to use the taurus validator instead of searching for "/" slash in the the door name. I have also added this macro to the catalogue in the docs. I will proceed with the merge. Many thanks!

Issue #889, opened 2018-08-01
access the trend window from a macro


dschick:
Hej all,

I would like to draw a simple vertical line into the trend window using a macro.
The macro should indicate the current position of the motor of the last scan.

Is this possible from Sardana via the taurus API?

reszelaz:
It sounds very attactive. But I'm affraid that currently it is not possible and won't be easy to implement.

We could implement something similar to [Macro Plotting](https://sardana-controls.org/devel/howto_macros/macros_general.html#plotting). If I remember correctly this works in the following way: when a macro requests to plot, this request is sent with a Tango event to the client which does the plotting on the client machine.

But first we need to overcome this [TODO](https://github.com/sardana-org/sardana/blob/develop/src/sardana/spock/magic.py#L54) which is explained with more details in #166.

dschick:
Okay, so macroplotting would then just replot the current data in the trend window, which would require to store that data always until the next scan happens.
And in addition to the data a horizontal line is plotted where the scan motor is or was.

I guess using matplotlib this is quite esay, despite of the fact, that the scan data needs to be stored for all counters. This would then also be questionable if one wants to store the statistics or the scan data in the env, see discussion in #290. In the later case, the stats are only calculated on demand and not always.

dschick:
I am just wondering if plotting extra data (lines) in the `showscan online` window could be implemented with the new` taurus_pyqtgraph` backend?

reszelaz:
I just came accross this issue again.
I think it won't be actually so difficult. Let's forget about the previous approach explained by me earlier.

In the showscan we receive the `RecordData` events. One option would be to use this attribute and pass the stats after the `record_end` data. The stats are calculated in the `post-scan` hook. Not sure where and how exactly we could make this push event there in the `GScan` - this will require further investigation.

Another option is to listen in the showscan also to the environment events. Then filter only the [ScanStats](https://sardana-controls.org/users/environment_variable_catalog.html#scanstats) events.

@tiagocoutinho how difficult would be then to plot the horizontal line in the pyqtgraph?

dschick:
Hi @reszelaz , thanks for remembering it!
I was actually asking to plot **vertical** lines in the pyqtgraph in order to visualize the position of the motor of the last scan, when you type `where` or show the new position after doing the `pic` or `cen` macros.
This saves a lot of time when aligning your experment as you do not have to read the values by hand.

So in this scenario you could always throw an event whithin these three macros.

[...]

PR #890, opened 2018-08-02
pic cen where

dschick:
add macros pic, cen, edge, and wher to move the motor of the last scan to the according position calculated from the statistics of a given counter or by default from the first plotted counter.
This PR relies on OptinalParameters #876 and on scan statistics #880 

All macros should be able to print their goto position in the trend window as well, see #889


dschick:
the `wher` macro is not `where` because numpy has such a macro already in the sardana namespace populated

dschick:
This PR now depends on #880 
Is there any guide how you tread stacked PRs?

I guess the `pic` and `cen` logic is quite simple and should be pretty similar to what is done at SOLARIS.

The `where` macro is not too usefull now, as it would ideally also mark the current motor positon in the `showscan online` window, see #889

Issue #802, opened 2018-07-04
Extend repeat macro to run it easily from spock


reszelaz:
Currently, it is not user friendly at all to use the `repeat` macro from spock. This macro works by executing another macros hooked in the `body` hook place. This can be done in the following ways:
* programmatically
* using sequencer
* using general hooks

We could extend this macro to accept also parameters. I propose to add a param repeat of strings to accept passing macro name and parameters e.g.

```python
    param_def = [
        ['nr', Type.Integer, None, 'Nr of iterations'],
        ['macro_name_params', [['token', Type.String, None, 'Macro name or macro parameter'], {'min':0}], None, "List with macro name and parameters (if any)"]
    ]
```

This repeat parameter is not mandatory (min=0). One could use the macro as it is now i.e. with hooks.
If one runs it with parameters, this should internally call: `self.execMacro`.

The usage from spock would be (using normal spock syntax):
`Door> repeat 100 ascan mot01 0 10 10 0.1`
or (using advanced spock syntax):
`Door> repeat 100 [ascan mot01 0 10 10 0.1]`

What do you think?

reszelaz:
This is inspired on the [macros](https://github.com/synchrotron-solaris/lib-solaris-facility-macros/blob/develop/facility_sequence.py) developed by Solaris. The idea is to merge both solutions and have only one macro for this.

teresanunez:
HI @reszelaz, I agree with what you propose.

reszelaz:
Great! For the moment I was thinking to restrict it and allow to repeat only one macro. In the future we could think of extendind this if necessary. I don't know what do you think?

teresanunez:
I was also thinking on that, allowing more than one macro, I think it would be interesting to have.

dschick:
this would be really great! especially repeating multiple macros

teresanunez <notifications@github.com> schrieb am Mi., 4. Juli 2018 um
14:53 Uhr:

reszelaz:
In this case I see two options for passing multiple macros with their parameters.

Either use param repeat of param repeats of strings. This would require to use the advanced spock syntax e.g.
```
Door> repeat 100 [[ascan mot01 0 10 10 0.1] [ct 10] [dscan mot02 -20 20 6 0.2]]
```

Or just use repeat parameter of strings, and if the macro require passing parameters enclose them in quotes e.g. 
```
[...]

PR #892, opened 2018-08-03
Extend repeat macro to run it easily from spock

teresanunez:
According to the discussion in #802, the syntax is implemented as:

Door> repeat 100 "ascan mot01 0 10 10 0.1" "ct 10" "dscan mot02 -20 20 6 0.2"

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -822,30 +822,45 @@ class logmacro(Macro):
         else:
             self.setEnv('LogMacro', False)
 
-
 class repeat(Hookable, Macro):
-    """This macro executes as many repetitions of it's body hook macros as
-    specified by nr parameter. If nr parameter has negative value,
-    repetitions will be executed until you stop repeat macro."""
+    """This macro executes as many repetitions of a set of macros as
+    specified by nr parameter. The macros to be repeated can be 
+    given as parameters or as body hooks. 
+    If the macros are given as parameters the body hooks are ignored. 
+    If nr has negative value, repetitions will be executed until you 
+    stop repeat macro.
 
+    .. note::
+        The repeat macro has been included in Sardana
+        on a provisional basis. Backwards incompatible changes
+        (up to and including removal of the macro) may occur if
+        deemed necessary by the core developers."""
+    
     # hints = { 'allowsHooks': ('body', 'break', 'continue') }
     hints = {'allowsHooks': ('body',)}
 
     param_def = [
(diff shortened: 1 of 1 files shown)


dschick:
works as expected for me

reszelaz:
Many thanks!

I do not know why but in my spock I have `repeat` name already reserved for the `numpy.repeat` function. In normal ipython I don't have it. This causes troubles when executing the macro. I need to proceed its name with `%` (explicitly call ipython's magic) I don't need to do it for other macros. Anyone else with this behavior?

Then, just to clarify, in case both parameters and hooks are set is it fine to execute only macros passed as parameters? I am not very decided about this and would like to know your opinions.

reszelaz:
Thanks for the hint! I will check that.

reszelaz:
This is the guilty one: https://github.com/sardana-org/sardana/blob/d8e9fbf28bf03d437f4fefd4c5e63b55ecb1f221/src/sardana/spock/ipython_01_00/genutils.py#L1124
When this line is commented the numpy object disappears from the namespace.

reszelaz:
From IPython's documentation:

> TerminalIPythonApp.pylab : ‘auto’|’gtk’|’gtk3’|’inline’|’nbagg’|’notebook’|’osx’|’qt’|’qt4’|’qt5’|’tk’|’wx’
Pre-load matplotlib and numpy for interactive use, selecting a particular matplotlib backend and loop integration.

> TerminalIPythonApp.pylab_import_all : Bool
Default: True

> If true, IPython will populate the user namespace with numpy, pylab, etc. and an import * is done from numpy and pylab, when using pylab mode.

> When False, pylab mode should not import any names into the user namespace.

I think we need to set the second one to False. At least I think we don't need all these in the namespace. What do you think?

teresanunez:
Hi @reszelaz, I also had this problem with the name repeat ... I made my tests calling it myrepeat.
Sorry, I thought it was working normally for all of you and it was a problem in my configuration.

teresanunez:
About this:

""Then, just to clarify, in case both parameters and hooks are set is it fine to execute only macros passed as parameters? I am not very decided about this and would like to know your opinions.""

I am also not sure how to proceed. The advantage of making it like it is now:

-> if parameters are given the body hooks are ignored

is that one can reuse the macro without having to remove the body hooks. This could save time in some cases.

teresanunez:
It is fine for me. Should I make the change or do you want to know more opinions before?

reszelaz:
Maybe let's wait until tomorrow. Thanks a lot!

PR #894, opened 2018-08-07
Store FQDN URI for Door and MacroServer in spock conf file

reszelaz:
When spock is used with Taurus4 store in the spock configuration file
Door and MacroServer FQDN URI references.

This is resting part of #668.

@sardana-org/integrators can you take a look? I've tested it with Taurus 4 but now with Taurus 3.

Note this does not upgrade the already exiting configuration file. I would not block the release for that cause there is a backwards compatibility layer so old configuration files basically works.. If at some point we want to upgrade them we can add it a posteriori.

Also I've not implemented it for IPython < 1.0. It is not probable that someone will use Taurus 4 together with such version of IPython.

### DIFF
--- a/src/sardana/spock/ipython_01_00/genutils.py
+++ b/src/sardana/spock/ipython_01_00/genutils.py
@@ -304,7 +304,8 @@ def get_macroserver_for_door(door_name):
             if dev.lower() == door_name:
                 for i, klass in enumerate(klasses):
                     if klass == 'MacroServer':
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi @reszelaz, I have tested it in taurus4 and taurus3. It works as you said. In Taurus3 is not making any
change, in taurus4 generates the config file with FQDN URI references. Old config files are not updated and works.

reszelaz:
Many thanks for the review! I'm merging then.

PR #895, opened 2018-08-07
Asynchronous Logstash communication

daneos:
This is a follow-up to #699 and an attempt to make Logstash communication asynchronous.

Things changed so far:
* use `python-logstash-async` instead of `python-logstash`
* add `LogstashCacheDbPath` property for both Pool and MacroServer

There is no need to define two Logstash endpoints with tagging (see [example configuration](https://github.com/reszelaz/sardana-elastic/blob/master/logstash/conf.d/sardana.conf)) since the `python-logstash-async` automatically adds `program` field. It's value is taken from the executable name, which in case of single Sardana instance is correctly recognized as `Sardana` and not `Pool`.

The asynchronous library uses caching in local SQLite database, path of which is configured by `LogstashCacheDbPath` property. This caching however causes problems when logs are produced at high speed (e.g. during a scan). Since all messages are logged first to the cache and then from cache to Logstash, it seems that with high message rate it is impossible to push messages to Logstash and record new ones in the cache simultaneously. The messages will end up in Logstash after some time, but a 1000 point scan is able to completely prevent any message from reaching Logstash for scan duration + a few minutes.
Other issue is the database file size, it needs to be manually cleaned (SQLite `VACUUM` command) and can grow significantly even when empty.

### DIFF
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -936,12 +936,7 @@ def prepare_logstash(args):
             props = db.get_device_property(dev_name, "LogstashCacheDbPath")
             cache_db_path = props["LogstashCacheDbPath"][0]
         except IndexError:
-            if class_name == "Pool":
-                cache_db_path = "/tmp/sardana-pool-logstash-cache.db"
-            elif class_name == "MacroServer":
-                cache_db_path = "/tmp/sardana-ms-logstash-cache.db"
-            else:
-                cache_db_path = "/tmp/sardana-logstash-cache.db"
+            cache_db_path = None
         return host, port, cache_db_path
 
     db = Database()
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -914,7 +914,7 @@ def prepare_logstash(args):
     log_messages = []
 
     try:
-        import logstash
+        from logstash_async.handler import AsynchronousLogstashHandler
     except ImportError:
         msg = ("Unable to import logstash. Skipping logstash "
                + "configuration...", )
@@ -953,7 +953,8 @@ def prepare_logstash(args):
 
     if host is not None:
         root = Logger.getRootLog()
-        handler = logstash.TCPLogstashHandler(host, port, version=1)
+        handler = AsynchronousLogstashHandler(host, port,
+                  database_path="/tmp/sardana-logstash-cache.db")
         root.addHandler(handler)
         msg = ("Log is being sent to logstash listening on %s:%d",
                host, port)
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -947,7 +947,7 @@ def prepare_logstash(args):
     if bin_name in ["Pool", "MacroServer"]:
         class_name = bin_name
         dev_name = get_dev_from_class_server(db, class_name, server_name)[0]
-        host, port, cache = get_logstash_conf(dev_name, class_name)
+        host, port, cache = get_logstash_conf(dev_name)
     else:
         dev_name = get_dev_from_class_server(db, "Pool", server_name)[0]
         host, port, cache = get_logstash_conf(dev_name)
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -916,7 +916,7 @@ def prepare_logstash(args):
     try:
         from logstash_async.handler import AsynchronousLogstashHandler
     except ImportError:
-        msg = ("Unable to import logstash. Skipping logstash "
+        msg = ("Unable to import logstash_async. Skipping logstash "
                + "configuration...", )
         log_messages.append(msg,)
         return log_messages
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -931,7 +931,7 @@ def prepare_logstash(args):
             props = db.get_device_property(dev_name, "LogstashPort")
             port = int(props["LogstashPort"][0])
         except IndexError:
-            port = 12345
+            port = None
         try:
             props = db.get_device_property(dev_name, "LogstashCacheDbPath")
             cache_db_path = props["LogstashCacheDbPath"][0]
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -921,7 +921,7 @@ def prepare_logstash(args):
         log_messages.append(msg,)
         return log_messages
 
-    def get_logstash_conf(dev_name, class_name=None):
+    def get_logstash_conf(dev_name):
         try:
             props = db.get_device_property(dev_name, "LogstashHost")
             host = props["LogstashHost"][0]
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -921,7 +921,7 @@ def prepare_logstash(args):
         log_messages.append(msg,)
         return log_messages
 
-    def get_logstash_conf(dev_name):
+    def get_logstash_conf(dev_name, class_name=None):
         try:
             props = db.get_device_property(dev_name, "LogstashHost")
             host = props["LogstashHost"][0]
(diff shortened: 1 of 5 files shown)


reszelaz:
Thanks for this work @daneos!

At the first glance I do not like the idea of SQL lite caching and if you say that it gives problmes maybe we should investigate another possibilities...

In the docs of [python-logstash-asynch persistance](https://python-logstash-async.readthedocs.io/en/latest/persistence.html#persistence) they say:

> By default, you do not need to provide a database_path to the AsynchronousLogstashHandler. There are a couple of things you should keep in mind if you choose to go down this path. With no database, the backend is a simple in-memory cache. Because it is in memory, your messages will not be kept across process restarts. This means it is possible to lose messages. If you cannot lose messages, then you should set the database_path option.

There are two scenarios that I imagine that the sardana server can stop:
1. server stop e.g. stopped from Astor or Ctrl+C (SIGTERM signal)...
2. server crash
 
In the docs of [python-logstash-async asynchronous processing](https://python-logstash-async.readthedocs.io/en/latest/about.html#asynchronous-processing) they say:

> Whenever the application stops, to be more exact whenever Python’ logging subsystem is shutdown, the worker thread is signaled to send any queued events and clean up itself before shutdown.

So, I imagine that this should cover the 1 scenario. IMHO it is fine if we sacrify logs from the server crash scenario. Furthermore, playing with the constants.QUEUE_CHECK_INTERVAL and constants.QUEUED_EVENTS_FLUSH_COUNT we could minimize the logs eventually being lost.

Also we need to keep in mind that #699 introduced the feature for direct logging into the Logstash for legacy OS. In a modern OS there is the possibility file beats.

@daneos will you continue with the tests without the SQL lite persistance?

daneos:
I'll see how this works without the on-disk database, but I think that the in-memory cache is also done using SQLite as a backend, so this may not solve the problem.

Regarding losing logs in the event of server crash, I would say that the logs gathered immediately before the crash would be probably the most valuable for debugging, so we should try hard to save them.

I know about Beats, but I think I'd prefer to be able to send logs directly from Sardana without the need for additional software on the control system machines, what's your opinion on that?

I'll continue with this, probably next week.

reszelaz:
Ok, so at least we are not the only ones who noticed this problem. Also, there is no information about the technical solution for SQL lite. Just a hint to avoid using it. I would go then for testing the *in memory cache*.

daneos:
I tested this without on-disk database, and it works better, the messages are constantly appearing in Logstash, but with scan of 40000 points I have now over 1h of delay in processing messages. This means if something breaks, I'd loose 1 hour worth of logs...
I'm not sure this is the way to go.

reszelaz:
Thanks for the tests @daneos!
> The messages are constantly appearing in Logstash, but with scan of 40000 points I have now over 1h of delay in processing messages.

Wow, thats really a lot of time. I was not expecting that. If I understands correctly this also implies that if we were using the synchronous Logstash handler the scan itself would gain 1 h of deadtime?

From the default values of [options for the asynchronous processing](https://python-logstash-async.readthedocs.io/en/latest/config.html#options-for-the-asynchronous-processing-and-formatting) I would expect to flush the data much faster.

I will check it as well whenever I have some time and see if I can reproduce it.

reszelaz:
I have tried this as well on my setup and without the db-cache (change line 954 of sardana.tango.core.util to return `None` instead of `db_cache_path`).

I run a step scan of 10k points: `ascan mot29 0 1 10000 0.001` and I was constantly monitoring the elasticsearch data with kibana.

All the data were correctly indexed almost instantaneously when the scan finished (I have not measured but for sure it was less than 5 seconds - the time that I moved my eyes from one window to the other:)). 

So I have not observed such big delays as you comment. The delay of 1h is only when you run 40k scan? Or maybe with shorter scans you see proportionally smaller delays?

daneos:
I tried with the 10000 point scan on single VM (which I suppose should be faster).
I still got the delay of about 12min... I'm planning to make two more tests: set the entire thing up on Debian (currently using CentOS) and if this still doesn't work well I'll try with the real hardware instead of VMs.

daneos:
I repeated the tests and noticed that changing the filter settings in LogTrail is enough to load all missing data points. For example, when the scan is around point 2000 I can see point 300 in LogTrail. Then, I change filter to, for example, `level:OUTPUT`, and suddenly I can see all scan points up to 2000.

I will modify the code, so it doesn't use default database when none is specified, and I think it will be ready to merge then.

reszelaz:
Great news @daneos!

Yes, for this kind of tests it is better to look with the "discover" view of Kibana (this is how I tested it) or even directly querying the Elasticsearch. Logtrail has some configuration parameters which could affect the refresh.

src/sardana/tango/core/util.py
```diff
@@ -942,18 +947,21 @@ def get_logstash_conf(dev_name):
...
                                                  server_name)[0]
-            host, port = get_logstash_conf(dev_name)
+            host, port, cache = get_logstash_conf(dev_name)
 
     if host is not None:
         root = Logger.getRootLog()
-        handler = logstash.TCPLogstashHandler(host, port, version=1)
+        handler = AsynchronousLogstashHandler(host, port, database_path=cache)
+        # don't use full path for program_name
+        handler._create_formatter_if_necessary()
```
reszelaz: Could you remind me what is it for?

src/sardana/tango/macroserver/MacroServer.py
```diff
@@ -397,7 +397,17 @@ class MacroServerClass(SardanaDeviceClass):
...
+        'LogstashCacheDbPath':
+            [DevString,
+             "Path to the Logstash cache database [default: None]. "
+             "It is advised not to use the database cache, as it may "
+             "have negative effects on logging performance. See #895. "
+             "This property has been included in Sardana on a provisional "
+             "basis. Backwards incompatible changes (up to and including "
+             "its removal) may occur if deemed necessary by the "
+             "core developers.",
+             "/tmp/sardana-ms-logstash-cache.db"]
```
reszelaz: I think here is the default value? Isn't it?

src/sardana/tango/pool/Pool.py
```diff
@@ -1380,7 +1380,17 @@ class PoolClass(PyTango.DeviceClass):
...
+        'LogstashCacheDbPath':
+            [PyTango.DevString,
+             "Path to the Logstash cache database [default: None]. "
+             "It is advised not to use the database cache, as it may "
+             "have negative effects on logging performance. See #895. "
+             "This property has been included in Sardana on a provisional "
+             "basis. Backwards incompatible changes (up to and including "
+             "its removal) may occur if deemed necessary by the "
+             "core developers.",
+             "/tmp/sardana-pool-logstash-cache.db"]
```
reszelaz: I think here is the default value? Isn't it?

src/sardana/tango/core/util.py
```diff
@@ -932,7 +932,12 @@ def get_logstash_conf(dev_name):
             port = int(props["LogstashPort"][0])
         except IndexError:
             port = 12345
```
reszelaz: It is not part of this PR, but I just found it. I think this is not correct. I think it is beter to move write it as the default property value in its declaration and documente it accordingly.

src/sardana/tango/core/util.py
daneos: This, and the line below is to get only `Pool` and not `/usr/bin/Pool` as `program` keyword.
This worked differently in the synchronous Logstash library.

src/sardana/tango/macroserver/MacroServer.py
daneos: Right, I removed it from the code, but forgot here.

src/sardana/tango/pool/Pool.py
daneos: Right, I removed it from the code, but forgot here.

reszelaz:
I just found some minor issues and commented them inline. Apart from that, could you please add a sentence about the new property in: [MacroServer configuration](https://sardana-controls.org/users/configuration/macroserver.html) and [Device Pool configuration](https://sardana-controls.org/users/configuration/pool.html)

src/sardana/tango/core/util.py
daneos: OK, I'll update this.

src/sardana/tango/core/util.py
reszelaz: Thanks! Then we need it:)

src/sardana/tango/core/util.py
```diff
@@ -914,14 +914,14 @@ def prepare_logstash(args):
...
+        from logstash_async.handler import AsynchronousLogstashHandler
     except ImportError:
-        msg = ("Unable to import logstash. Skipping logstash "
+        msg = ("Unable to import logstash_async. Skipping logstash "
                + "configuration...", )
         log_messages.append(msg,)
         return log_messages
 
-    def get_logstash_conf(dev_name):
+    def get_logstash_conf(dev_name, class_name=None):
```
reszelaz: I think this `class_name` is not used anymore?

daneos: You're right. It was used to determine correct default DB path, but there is no default DB now.

reszelaz:
Thanks @daneos, then I'm merging this!

PR #908, opened 2018-08-09
Move param parser out of Spock

### LINKED ISSUE #907: MacroServer requires Qt bindings
This issue was found during Jul18 release testing on CentOS (#903).

The MacroServer imports from spock which tries to import Qt from Taurus.
Tracebacks can be found here:
https://gist.github.com/daneos/fd88c73af209c1bdabb9aa71b482c15f
https://gist.github.com/daneos/b6f08014e579e20cbe99dc9a1e5d7a1d

Quick git bisect shows that the bug was introduced in commit 399c2c0b, pull request #646 (General Hooks).

daneos:
As the macro parameter parser is used also by MacroServer, it should be placed outside of Spock, so MacroServer doesn't need to import whole Spock and Qt with it.

This PR fixes #907.

@sardana-org/integrators, could you look at this so we can incorporate this into Jul18 release?


reszelaz:
> But I anticipate that in order to properly implemt #781 we will need to add more inteligence to the parser (let it know the macro definition).

I was wrong here. After properly implementing #781 the parser is still sardana independent. So the place proposed in this PR where to put it: `sardana.util` is correct from my point of view.

Now, the only problem is that the #781, that has major changes to the parser, would become hard to merge if we accept this PR. I propose to first integrate #781 and then propose another PR that would move the parser to `sardana.util` again. Sorry..

daneos:
OK, fortunately the changes here are really trivial, so we can redo this after merging #781.
I'm closing this PR then.

PR #921, opened 2018-08-30
Adapting for 4c diffrac

teresanunez:
After testing with a new beamline for a new diffractometer type, we need to adapt the hkl macros.


reszelaz:
I've just made some tiny correction. Merging it. Thanks!

reszelaz:
If I understand it correctly the changes in the macros were for adapting them for "PETRA3 P23 6C" diffractomter. This is just to put it in the CHANGELOG. Thanks!

Issue #925, opened 2018-09-04
Expconf does not check the scan path


gjover:
Expconf does not warn the user if the scan path is a valid path.
Would be very useful if it checks it

reszelaz:
See discussion from #926.

HEnquist:
The path must be checked by the MacroServer (likely not on the machine that runs expconf). Check could be implemented as a macro. Expconf runs the macro with a new path to see if valid. LineEdit edit finished signal should trigger the macro. Red background if bad, disable "Apply". A popup warning instead?

PR #926, opened 2018-09-05
Solving issue #925

gjover:


src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
```diff
@@ -321,6 +322,21 @@ def writeExperimentConfiguration(self, ask=True):
                 self.changeActiveMntGrp(mgname)
                 return False
 
+        # check path
+        path = self._localConfig['ScanDir']
+        if not os.access(path, os.F_OK):
+            try:
+                os.mkdir(path)
```
cmft: On the one hand, using `os.mkdir` is not  a good option to check if the given path is writable or not since it only check one path level, I mean it will fail with subpath. e.g.  when you set up /foo/bar as path and neither foo and bar exist. On the other hand `spock` and `sardana` could be launched with different users which could have different permissions so you can not  advise the user about if sardana can write o no the data with this kind of checking.

.bumpversion.cfg
```diff
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 2.5.1-alpha
+current_version = 2.5.2-alpha
```
cmft: For the moment you don't have to bump the sardana version, but update the changelog is a must.

src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
gjover: I understand.
Then it would be better to move the check to the door.

gjover: Should the EnvironmentManager validate the values?
Which would be the right place to do the check?

reszelaz:
I agree with the review of @cmft:

>  On the other hand spock and sardana could be launched with different users which could have different permissions so you can not advise the user about if sardana can write o no the data with this kind of checking.

The solution would rather need to be implemented on the server side. 
I proceed to close this PR. If someone does not agree we could reopen.

PR #933, opened 2018-09-10
Move pre-scan and post-scan hooks out of scan_loop (issue-920; previously pr-922)

reszelaz:
pre-scan and post-scan hooks are executed within the scan_loop. Before
the scan loop, for example, are started the scan recorders making
them to start the scan entry. It makes more sense to not start them
until the pre-scan hooks are executed - they may want discard the scan
or simply may not want to mix the scan header output with the pre-scan
hooks output.

This solution allows to avoid the code duplication in all the scan
specific implementation.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -944,6 +944,10 @@ class GScan(Logger):
             pass
 
     def step_scan(self):
+        macro = self.macro
+        if hasattr(macro, 'getHooks'):
+            for hook in macro.getHooks('pre-scan'):
+                hook()
         self.start()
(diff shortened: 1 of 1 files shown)


dschick:
if the result is fine, I am fine too.
I had no time to look into the code so far.
I hope I can test it this week, after that I will be off for two weeks.

Thanks for taking care!

reszelaz:
The result is exactly the same is in #922 which you already tested. The only additional thing is that the same policy was applied to the post-scan hook places. But I think we need to be coherent with both hook places.

I'm merging this one. Feel free to feedback after tests from the develop branch alredy. Thanks a lot!

PR #936, opened 2018-09-12
Correct continuous scan hooks

reszelaz:
Continuous scans e.g. `ascanct`, `dscanct` or `meshct` does not allow to attach all hook places even if in the code they are implemented. Fix it by correctly defining `allowHook` hint. Do it in newly created base clases for these macros in order to avoid duplication of code.

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -1549,7 +1549,7 @@ class dNscanct(dNscan):
                              "post-scan")}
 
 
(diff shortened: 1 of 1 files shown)


reszelaz:
@sardana-org/integrators this is ready for review. Thanks!

teresanunez:
IMO it can be merged. Thanks @reszelaz

reszelaz:
Thanks for the review! I'm integrating it

PR #940, opened 2018-09-18
Provide MeasurementGrp as a parameter in ct macro

droldan:
Add measurement group as a optional parameter using the Option parameter feature

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -659,11 +659,10 @@ class ct(Macro, Hookable):
         if mnt_grp is None:
             self.mnt_grp_name = self.getEnv('ActiveMntGrp')
             self.mnt_grp = self.getObj(self.mnt_grp_name,
-                                   type_class=Type.MeasurementGroup)
+                                       type_class=Type.MeasurementGroup)
         else:
             self.mnt_grp_name = mnt_grp.name
             self.mnt_grp = mnt_grp
-
 
     def run(self, integ_time, mnt_grp):
         if self.mnt_grp is None:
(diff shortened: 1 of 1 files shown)


src/sardana/macroserver/macros/standard.py
```diff
@@ -41,6 +41,8 @@
     ViewOption, iMacro, Hookable
 from sardana.macroserver.msexception import StopException
 from sardana.macroserver.scan.scandata import Record
+from sardana.macroserver.msparameter import Optional
```
reszelaz: The idea is to imprort it from sardana.macroserver.macro as in the example.

src/sardana/macroserver/macros/standard.py
```diff
@@ -648,17 +650,24 @@ class ct(Macro, Hookable):
     env = ('ActiveMntGrp',)
     hints = {'allowsHooks': ('pre-acq', 'post-acq')}
     param_def = [
-        ['integ_time', Type.Float, 1.0, 'Integration time']
+        ['integ_time', Type.Float, 1.0, 'Integration time'],
+        ['mnt_grp', Type.MeasurementGroup, Optional, 'MntGrp to use']
```
reszelaz: I would not use accronyms in the description.

src/sardana/macroserver/macros/standard.py
droldan: Provided in [43eabb4](https://github.com/sardana-org/sardana/pull/940/commits/43eabb45148e4ea0512a146f7b7cb7577bf0b237) commit.

src/sardana/macroserver/macros/standard.py
droldan: Provided in [43eabb4](https://github.com/sardana-org/sardana/pull/940/commits/43eabb45148e4ea0512a146f7b7cb7577bf0b237) commit.

reszelaz:
Great! Thanks for the changes. Merging..

Issue #931, opened 2018-09-07
Macrobutton should try to stop macro instead of direct aborting


reszelaz:
Currently macro button calls abortMacro command instead of stopMacro. It should at least try to stop it.

reszelaz:
Solved in #943

PR #943, opened 2018-09-21
MacroButton Stop macro instead direct Aborting

droldan:
Pull Request to solve the [issue 931](https://github.com/sardana-org/sardana/issues/931) (Macrobutton should try to stop macro instead of direct aborting #931)


reszelaz:
This PR forgot to change the macro button text. I just pushed it directly to the develop branch: cd4320f.

Issue #788, opened 2018-06-22
MeasurementGroup gets corrupted if you disable all channels


reszelaz:
If you disable all channels in the measurement group, for example in expconf, then your measurement groups is not operable anymore - you can not reenable them back. One has to manually modify the configuration `__value` attribute (memorized value)  and restart the Pool.

reszelaz:
This issue was fixed:

\<disable all channels in expconf\>

```
Door_py3_1 [3]: ct 
Tue Sep 10 09:45:03 2019

An error occurred while running Macro 'ct(1.0, None) -> e65d23d6-d39e-11e9-a544-f8b156a48f79':
RuntimeError: all channels in measurement group are disabled
```

\<Enable all channels in expconf\>

```
Door_py3_1 [4]: ct 
Tue Sep 10 09:45:17 2019

  ct05  =         1.0
  ct06  =         2.0
  ct07  =         3.0
  ct08  =         4.0
```

PR #952, opened 2018-09-26
Add protection when all timers are disabled

rhomspuron:
it solves #788


teresanunez:
I have tested this and for me there is not change in the behaviour observed after disabling all channels.
#788 is not solved. It could be that I misunderstood the conditions at which this should be tested, but in
 any case my MG get corrupted with or without the change if I disable all channels.

reszelaz:
@rhomspuron do you think that this is still necessary after hotfix-2.7.2?

rhomspuron:
@reszelaz I will close it, the hotfilx protects this case.

PR #955, opened 2018-10-05
Fix optional param for python 2.6

reszelaz:


reszelaz:
This is a trivial bug fix for python 2.6 so I don't bother the integrators and just merge it.

Issue #967, opened 2018-10-19
readone after stateone works in different way in 2.5.0


pjb671:
My first issue submitted here. At one beamline at MAX IV we have upgraded to sardana 2.5.0.  It seems as though the behaviour of the experimental channels has changed - a piece of hardware that returns a 1d array gets read during the acquisition, which results in incomplete data being returned. To confirm I made a super simple 1d controller with the following methods:

    def StartOne(self, axis, position=None):
        self.state = State.Moving
        print("----------------------------------STARTED -> MOVING")

    def StateOne(self,ind):  
        self.status = "test"
        #self.state = State.On   SO IT IS ALWAYS MOVING AFTER DOING STARTONE
        tup = (self.state,self.status)
        print("----------------------------------IN STATEONE", ind, self.state)
        return tup

    def ReadOne(self,ind):
        if self.state == PyTango.DevState.MOVING:
            print("------------------- STATE IS MOVING : SHOULD NEVER CALL READONE!")
            return np.zeros(MAX_BUF_SIZE)
        else:
            return np.zeros(MAX_BUF_SIZE)

In 2.3.1, it works, or rather fails, as expected. The state is always moving so a step scan hangs waiting forever, ReadOne is never called:

----------------------------------STARTED -> MOVING
----------------------------------IN STATEONE 2 6
----------------------------------IN STATEONE 2 6
...this is printed forever

In 2.4.0 and 2.5.0, the readone is called while we are still in moving state:

----------------------------------STARTED -> MOVING
----------------------------------IN STATEONE 2 6
------------------- STATE IS MOVING : SHOULD NEVER CALL READONE!
----------------------------------IN STATEONE 2 6
----------------------------------IN STATEONE 2 6
...and readOne is repeatedly called every second or so.

Is this an expected new behaviour? It fundamentally changes how we are used to things working.  I would not expect Sardana to try and read the hardware while stateone returns moving.

reszelaz:
It looks like a bug. The C/T channels are read while acquiring to report count updates. But the 1D/2D should not be. I will check carefully and let you know.

reszelaz:
I confirm it's a bug. Next week I will provide a fix.

reszelaz:
You mean to not read the channel selected as Timer in the measurement group configuration?

Just loud thinking... Maybe a long `uct` execution could see how much time elapsed from the start? Of course if the controller knows how to report this time.

dschick:
[...]

PR #969, opened 2018-10-22
Hotfix 2.5.1: do not read 1D and 2D in software acquisition loop.

reszelaz:
This is a PR to the master branch in order to hotfix #967. @pjb671 could you please verify it? and @amilan eventually integrate it? Of course anyone is more then welcome to review it:)

The integration would require tag of the master branch with 2.5.1 and opening a similar PR to the develop branch where version should be bumped after integration to 2.5.2-alpha.

Many thanks in advance!


amilan:
Hi Zibi, 

Many thanks for providing that solution, we will evaluate it soon and share the results.

teresanunez:
Hi @reszelaz, sorry, I have tested this because we have problems with our 1D detectors due to this bug. I can confirm that this solves the problem (at least in my case tested in 1D experimental channels).

reszelaz:
Thanks for the review @teresanunez !

If there is an interest in the hotfix release please integrate this PR. @amilan do you need it in MaxIV? 

Otherwise, just push the changes from `src/sardana/pool/poolacquisition.py` to develop branch. This way we will save time on the release "paperwork".. Soon we will make Jan19 release anyway..

PR #972, opened 2018-10-24
Flyingbot bugfix pool page

flyingbot91:
When `len(sys.argv) > 1` the second condition is never met


reszelaz:
AFAIK this part of the code base is not used, but merging anyway cause it seems correct.

flyingbot91:
Thanks @reszelaz

PR #983, opened 2018-11-15
Fix general hooks broken by #781

reszelaz:
#781 made `ParamParser` aware of the parameters definition. General hooks use the `ParamParser`. Adapt geral hooks implementation to pass the parameters definition to the `ParamParser`.

Also add some unit tests.

@sardana-org/integrators it is ready fore review. Thanks!


teresanunez:
i have tested the general hooks, and after this fix they work again. Thanks.

reszelaz:
Great! Thanks @teresanunez for the review. I'm integrating this.

PR #984, opened 2018-11-19
Document RConsolePort

reszelaz:
This PR complements the MacroServer configuration docs. Eventually, we could add in the future a dedicated chapter for Sardana debugging and a part of this information could be moved there.
@sardana-org/integrators could you please take a look and eventually integrate it? Thansk!


teresanunez:
Hi @reszelaz, I have tried it for first time following what you document here and iMO it can be merged. Thanks.

reszelaz:
Thanks to both! Merging.
BTW this is a very usefull feature that could help in debugging problems with MacroServer.

Issue #982, opened 2018-11-14
ROI pseudocounters from 2Dcounter


dschick:
Hej all,

I am going to implement a 2D counter acquiring images from a CCD.
The actual program controlling the CCD (LabView) not only provides access to the full image, but also to a 1D integration of the image and up to 10 ROIs which are an array of 10 floats. Since LabView already provides a GUI for all this, I do not want to transfer the ROI definition and calculation to Sardana but keep it all in LabView.
The CCD program is interfaced from Sardana by an TCP/IP server which I can modify as I wish.

Now I would like realize the 2D image, the 1D integration as well as the 10 ROIs as 2Dcounter, 1Dcounter, as well as 10 Timer/Counters respectively. I can get all the data with one TCP/IP command from the server, but of course I need to distribute it between the different counter types.

Are there already concepts around for such a situation?

Best

Daniel

teresanunez:
Hi Daniel,
                   we also have the case to send information from a detector as 2d or as counters (for the
 RoIs), we simply create controllers for the different types and take care about the synchronization, but
 in principle there is no problem.

                                   Regards,
                                                       Teresa

dschick:
Hej @teresanunez , thanks for the quick reply.

Okay, but how do you do the synchronization exactly? Is your code online available?

I guess you have to open a separate connection to the camera server (I guess you have implemented the camera in Tango?) in each Sardana controller?

Best

Daniel

rhomspuron:
Hi @dschick, 

As @teresanunez commented, in ALBA we also use a controller `CounterTimerController` for the ROIs. We base this controller on the Lima ROI Device Server.  There is another controller to configure the detector. The SEP2 should solve the ROI integration on Sardana. 

ROI controller: https://sourceforge.net/p/sardana/controllers.git/ci/master/tree/python/countertimer/LimaRoICoTiCtrl.py
Lima controller: https://sourceforge.net/p/sardana/controllers.git/ci/master/tree/python/countertimer/LimaCoTiCtrl.py

Cheers,
Roberto

reszelaz:
SEP2 should answer to all this questions. 

Meanwhile mutliple contollers seems to workaround it quite well.
You could develop an interface with the hardware as a [singleton](https://en.wikipedia.org/wiki/Singleton_pattern) class together with some serialization primitives (if necessary) and them use it in all controllers.

[...]

PR #986, opened 2018-11-23
Allow changes of pseudo counter data format (shape)

reszelaz:
This feature was initiated by the #982 discussion. Now it is possible to define pseudo counters which returns a scalar, a spectrum or an image.

I tested it with the following trivial example:

```python

class ROI(PseudoCounterController):
    """A simple pseudo counter which receives an image from a 2D experimental
       channel and returns 1D of its first row"""

    counter_roles = "2D",

        axis_attrs = PseudoCounterController.GetAxisAttributes(self, axis)
        axis_attrs = dict(axis_attrs)
        axis_attrs['Value'][Type] = (float,)
        axis_attrs['Value'][MaxDimSize] = (1024,)
        return axis_attrs

    def Calc(self, axis, counter_values):
        i = counter_values[0]
        s = i[0]
        return s
```

It is very trivial, and assumes image of 1024 x 1024 pixels (for example the one provided by the `DummyTwoDController`).

This PR also includes the necessary documentation.

@sardana-org/integrators it is ready for the review and eventual integration.
Don't hesitate to ask if something does not work at first try. I made it in rush and tested it just with this simple example. Thanks!

### DIFF
--- a/doc/source/devel/howto_controllers/howto_controller.rst
+++ b/doc/source/devel/howto_controllers/howto_controller.rst
@@ -404,6 +404,33 @@ spock), sardana assignes the default value
 has no default value, if it is not specified by the user, sardana will complain
 and fail to create and instance of SpringfieldMotorController.
 
+.. _sardana-controller-howto-change-default-interface:
+
+Changing default interface
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Elements instantiated from your controller will have a default interface
+corresponding to the controller's type. For example a moveable will have a
+*position* attribute or an experimental channel will have a *value*
+attribute. However this default interface can be changed if necessary.
+
+For example, the default type of a moveable's *position* attribute ``float``
+can be changed to ``long`` if the given axis only allows discrete positions.
+To do that simple override the
+:class:`~sardana.pool.controller.Controller.GetAxisAttributes` where you can
+apply the necessary changes.
+
+Here is an example of how to change motor's *position* attribute to ``long``:
+
+.. code-block:: python
+
+    def GetAxisAttributes(self, axis):
+        axis_attrs = MotorController.GetAxisAttributes(self, axis)
+        axis_attrs = dict(axis_attrs)
+        axis_attrs['Position']['type'] = float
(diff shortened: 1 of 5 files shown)


teresanunez:
Hi @reszelaz, many thanks for this implementation. I have started to test it and I have some comments:
1. perhaps you assume everybody would know ... but in the example the pseudo counter role is missed.
2. for me it is working if the counter is 2d and pseudo counter 1d, but with pseudo counter 2d is not working, it appears like 1d. I did it as you described in the documentation file of this PR ... perhaps
 I am setting something wrong. Have you tried it?.

Many thanks.

reszelaz:
There was a bug in the proposed documentation, and actually another one in the example povided in the initial PR comment.
The first one is that the `counter_roles` is a tuple - there was a comma missing.
The second one, is that in order to set a 2D value type you must use a typle with an intenal tuple containing the actual type: `((float, ), )`. The complete controller example could look like this:

```python
class ROI2D(PseudoCounterController):
    """A simple pseudo counter which receives an image from a 2D experimental
       channel and returns a 2D ROI after cropping one pixel frame"""

    counter_roles = "2D",

        axis_attrs = PseudoCounterController.GetAxisAttributes(self, axis)
        axis_attrs = dict(axis_attrs)
        axis_attrs['Value'][Type] = ((float, ), )
        axis_attrs['Value'][MaxDimSize] = (1024, 1024)
        return axis_attrs

    def Calc(self, axis, counter_values):
        i = counter_values[0]
        roi = i[1:-1, 1:-1]
        return roi
```

> perhaps you assume everybody would know ... but in the example the pseudo counter role is missed.

I thought that this was documented, but actually only the case of omitting the `counter_roles` is documented. I will add it to the documentation right now and will push it directly to the develop branch.

reszelaz:
> I will add it to the documentation right now and will push it directly to the develop branch.

Done in: 23798d8

teresanunez:
Many thanks, @reszelaz, I have tested it now again and IMO it can be merged.

reszelaz:
Thanks! Merging it.

PR #988, opened 2018-11-28
Fix expconf for Qt 4.4 (signal style compatibility)

reszelaz:
#882 made expconf incompatible with Qt 4.4 (old style signals). This PR fixes it.

@sardana-org/integrators this is ready for review and eventual integration. Thanks in advance!
I have tested it both on Qt 4.4 and Qt 4.8 and for me both works.


cpascual:
Everything looks ok to me. Merging.

PR #990, opened 2018-11-28
Fix limits protection in ascanct when Taurus 4 is used

### LINKED ISSUE #989: ascanct gets wrong motor range when position unit is configured
I have done some tests with continuous scan using the ascanct macro and I noted that when we have the motor limits different of -inf inf and position unit configured, sardana gets wrong motor range.

For example:
```
set_lim mot01 -500 500
ascanct mot01 0 10 10 0.01 0
```
If the position unit of mot01 is configured to "degrees" sardana will get wrong motor range.
I have tested it with sardana 2.5.0 and taurus 4.4.0.

reszelaz:
Taurus 4 uses quantities as attribute values, range, etc. Sardana for
the moment does not support them, for example it is not possible to
pass macro parameters as quantities. In the continuous scan, when motor
limits are checked, it is better to use the position attribute limit
magnitudes. Fixes #989.

@sardana-org/integrators this is ready for review and eventual integration. Thanks!

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1482,6 +1482,12 @@ class CScan(GScan):
         pos_obj = motor.getPositionObj()
         min_pos, _ = pos_obj.getRange()
         try:
+            # Taurus 4 uses quantities however Sardana does not support them
+            # yet - use magnitude for the moment.
+            min_pos = min_pos.magnitude
+        except AttributeError:
+            pass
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi @reszelaz, I have tested and the problem described in #989 is solved. IMO it can be merged.

reszelaz:
Great! Thanks for the review! Merging..
At some point we should also support quantities as Taurus does.

PR #997, opened 2018-12-19
Independent acquisition of timerable channels (#185)

teresanunez:


src/sardana/pool/poolbasechannel.py
```diff
@@ -228,6 +229,37 @@ def _set_value(self, value):
...
+        self._integration_time = integration_time
+        if not propagate:
+            return
+        if integration_time == self._integration_time:
+            # current state is equal to last state_event. Skip event
+            return
+        self.fire_event(EventType("integration_time", priority=propagate),
+                        integration_time)
+
+    def put_integration_time(self, integration_time):
```
reszelaz: Not sure if this method is really necessary.

src/sardana/pool/poolbasechannel.py
```diff
@@ -228,6 +229,37 @@ def _set_value(self, value):
...
+        :return: the current integration time
+        :rtype: bool"""
+        return self._integration_time
+
+    def set_integration_time(self, integration_time, propagate=1):
+        self._integration_time = integration_time
+        if not propagate:
+            return
+        if integration_time == self._integration_time:
+            # current state is equal to last state_event. Skip event
```
reszelaz: Here the comments are about the state.

src/sardana/pool/poolonedexpchannel.py
```diff
@@ -66,3 +67,34 @@ def read_data_source(self):
 
     data_source = property(get_data_source,
                            doc="source identifier for the 1D data")
+
```
reszelaz: What about adding a `PoolTimerableChannel` in `sardana.pool.poolbasechannel`.
It would iherit from `PoolBaseChannel` and would be a base class for `PoolCounterTimer`, `Pool1DExpChannel` 
 and `Pool2DExpChannel`. This way we eliminate the code duplication.

src/sardana/tango/pool/CTExpChannel.py
```diff
@@ -215,6 +215,20 @@ def is_Value_allowed(self, req_type):
             return False
         return True
 
+    def read_Timer(self, attr):
```
reszelaz: What about adding a `PoolTimerableDevice` in `sardana.tango.pool.PoolDevice`.
It would iherit from `PoolExpChannelDevice` and would be a base class for `CTExpChannel`, `OneDExpChannel` 
 and `TwoDExpChannel`. This way we eliminate the code duplication.

teresanunez:
Hi @reszelaz, I have implemented what we have been talking about. You can have a look when you have time. I leave it like this before I get some feedback. Many thanks.

reszelaz:
Thanks @teresanunez for working on that! I will do my best but I'm affraid I won't have time to review this one before the ABCAD workshop.

reszelaz:
@teresanunez finally I could review this PR. I fixed some things - see my commits. It would be nice if you could recheck this now if it still works as expected.

Meanwhile I will try to document this feature and maybe even allow to pass an experimental channel as the `ct` macro parameter.

reszelaz:
I have added documentation and the taurus extension `count` method. Regarding the `ct` macro parameter I will treat it as a separate PR cause it requires changes not only affecting the independent acquisition. With this I remove the WIP tag and in my opionion this is ready to be integrated. Maybe this will even get into the Jan19 release.

teresanunez:
Hi @reszelaz, many thanks for the review and the changes. I have tested it now and it works as
expected.

reszelaz:
Great! I merget it then. Soon I will start discussion obout the `ct` macro parameter.

PR #1001, opened 2019-01-07
Setting Moveable to None if 'None' is written in the Moveable tango attribute

teresanunez:


src/sardana/pool/poolmeasurementgroup.py
```diff
@@ -1070,12 +1070,16 @@ def get_moveable(self):
 
     def set_moveable(self, moveable, propagate=1, to_fqdn=True):
         self._moveable = moveable
-        if self._moveable != 'None' and self._moveable is not None:
+        if self._moveable is not None:
             if to_fqdn:
                 moveable = _to_fqdn(moveable, logger=self)
             self._moveable_obj = self.pool.get_element_by_full_name(moveable)
+
+        if moveable is None:
```
reszelaz: The moveable should travel still as `None` within the Sardana kernel (core) events.
One should cast it to str in the Tango device callback. See a similar commit: https://github.com/sardana-org/sardana/pull/997/commits/a9b6d17e8c6102dd36adf5bc3d19a447bf19e14a for reference

teresanunez: Hi @reszelaz, many thanks.  I hope I understood the change that should be done.

reszelaz: Exactly. The change is correct. You just used the wrong variable name (`value` instead of `event_value`. I fixed this and the flake8 errors. I will merge whenever travis-ci accepts it. Thanks!

Issue #962, opened 2018-10-15
General Hooks are not called if a scan is called with createMacro/runMacro


teresanunez:
if a scan is started using createMacro/runMacro the general hooks are not called. One
 can test it defining general hooks and running this macro:

class escan_test(Macro):
    """Scan energy"""
    
    param_def = [ 
        ['start_energy',  Type.Float,  -999, 'Scan start energy'],
        ['end_energy',  Type.Float,   -999, 'Scan final energy'],
        ['nr_interv',  Type.Integer, -999, 'Number of scan intervals'],
        ['integ_time', Type.Float,   -999, 'Integration time'],
        ['fixq', Type.String, "Not", 'Add fixq  as argument if q has to be kept fixed' ],
        ['return_flag', Type.Integer, 1, 'If 1, motor returns to start position' ]
        ]
    
    def energy_pre_move(self):
        self.info("\tCalling energy_pre_move hook")

    def hkl_pre_move(self):
        self.info("\tCalling hkl_pre_move hook")
        
    def hkl_post_move(self):  
        self.info("\tCalling hkl_post_move hook")   
        
    def run(self,  start_energy, end_energy, nr_interv, integ_time, fixq, return_flag):

        dummy_motor = self.getObj("exp_dmy01")
        
        macro,pars = self.createMacro("ascan", dummy_motor, start_energy, end_energy, nr_interv, integ_time)

        macro.hooks = [ (self.hkl_pre_move, ["pre-move"]), (self.hkl_post_move, ["post-move"])]

        self.runMacro(macro)

Only the programmatic hooks are called.
Using execMacro instead of createMacro/runMacro:

        self.execMacro("ascan", dummy_motor, start_energy, end_energy, nr_interv, integ_time)

the general hooks are called, but we need to use createMacro/runMacro for adding additional programmatic hooks.

reszelaz:
I've not tested it, but I think I know what is going on...

Before introducing *general hooks* the use of the `Hookable.hooks` property was not clear. There was no way of appending hooks, only a way of setting all hooks at once. Furthermore this did not work 100% correctly - see #655.

Then, when working on #646, in 3d24295, we added `Hookable.appendHook` method. Which was intended to substitute the *old* way of adding hooks programatically. If you continue using the `Hookable.hooks` property directly you override the general hooks (and I guess the sequencer hooks as well). The corresponding [examples](https://sardana-controls.org/devel/examples/macro_hooks_examples.html) are still not updated...

[...]

PR #1013, opened 2019-01-15
Documenting how to add programmatic hooks keeping the general ones (solving #962)

teresanunez:


reszelaz:
Thanks for this docs @teresanunez!
I have moved the note about difference between the `hooks` property and the `appendHook` method from the "Users Guide" to the `Hookable class` API docstring. I consider it too specific as a first information that the user finds in the docs about the hooks. This required a tiny change on the `hooks` property implementation: c4d02b7. I hope it won't break anything..

Wheneve travis-ci gets green I proceed with the merge.

Issue #1012, opened 2019-01-14
update input of scan trend when changed outside of expconf


dschick:
I am not sure if this a duplicate, but when I open a scan trend from a expconf window, the plotted counters in the trend do not change, when I change the selection in the expconf window.

I need to close the trend window and reopen it, to let the changes take effect.

dschick:
Hej @cpascual ,

actually I have to check, if I use the "Apply" button.

What I am pretty sure of is, that when I am using the propose plotselect macro #824 with expconf in auto-update mode, then it is also not changing.

cpascual:
> What I am pretty sure of is, that when I am using the propose plotselect macro #824 with expconf in auto-update mode, then it is also not changing.

About that, I am not aware of the exact current state of those developments, so I cannot say if the behaviour is to be expected... but I think that it is an independent issue. 

What I can say fo sure is that the expconf trend is supposed to be updated when new settings are applied  **from the same expconf widget**

dschick:
hej @cpascual , indeed I forgot to hit the apply button when changing the plotted channels in the expconf.
However, all changes I make from spock to the measurement group via macros are not applied to the trend window.
Therefor I am changing the title of the issue to:

> update input of scan trend when changed outside of expconf

droldan:
Hi @dschick @cpascual,

I have checked what you say and yes, it seems like the expconf doesn't send a 'config changed' event to the plots opened, is necessary to close and reopen the plot.
The good news is that the expconf is aware of any changes made isn the measurement groups even if they are made using macros.

I will try to review the event connection between the expconf and plot.

I have reproduced the issue using the proposed plotselect macro #824

droldan:
Hi @dschick ,
I've created a PR #1016 to solves this issue.
The event 'experimentConfigurationChanged' was only sent to scan plot when the changes were made with expconf widget.

With this fix, the event is also propagated when the measurement group is modified from external client, in this case Macros.

dschick:
Thanks a lot @droldan ! it's working perfectly fine!

PR #1009, opened 2019-01-09
Qt5

mrosanes:
This PR adds support for Qt5 in sardana (and drops support for Qt<4.6)

It fixes most of the items required for #1006 (except those related to trend support under Qt5, wich need replacement of the qwt5 dependency)

PR #1016, opened 2019-01-17
ExpConf -  Emit 'experimentConfigurationChanged' on external changes

droldan:
This solves the issue #1012.
The scan plot only received 'experimentConfigurationChanged' when the changes were made in the expconf widget.
With this fix, the event is also propagated when the measurement group is modified from external client, in this case Macros.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
@@ -441,6 +441,8 @@ class ExpDescriptionEditor(Qt.QWidget, TaurusBaseWidget):
         for tg_info in tg_elements.values():
             avail_triggers[tg_info.full_name] = tg_info.getData()
         self.ui.channelEditor.getQModel().setAvailableTriggers(avail_triggers)
+        self.emit(Qt.SIGNAL('experimentConfigurationChanged'),
+                  copy.deepcopy(conf))
 
     def _setDirty(self, dirty):
         self._dirty = dirty


dschick:
it is working for me - also with the --auto-update flag of %expconf

reszelaz:
I just reviewed this changes. @droldan what do you think to change a little bit the strategy how the scan trends gets updated. Now these react on the "experimentConfigurationChanged" Qt signal comming from the `ExpDescriptionEditor` widget. What about connecting the trends to the "experimentConfigurationChanged" Qt signal comming from the `QDoor` directly? Previously this was not possible cause the `QDoor` was not emitting them. But now this should be theoretically possible. This way we could decouple the scan trends from the expconf. Initially these two were decoupled and the scan trends were launched with the `showscan online` spock magic command.

guifrecuni:
And if I remember correctly, it was also possible to get the scan trends when running `taurustrend <door_alias>` from console or any other process.

dschick:
I would appreciate it, to decouple the trend from the config.
Would this also mean, that one could manipulate the trend from within spock
or any client connected the same door?

Am Di., 22. Jan. 2019, 17:03 hat Guifre Cuni <notifications@github.com>
geschrieben:

cpascual:
> Would this also mean, that one could manipulate the trend from within spock or any client connected the same door?

Yes, and that is the case already (at least in the qt5 branch). The plots are not yet separated from the expconf, but they already react to external changes via spock.

Separating the plots from the expconf should not change this.

**IMPORTANT** note that in the qt5 branch we changed the signature of the `QDoor.experimentConfigurationChanged`signal from `()` to `object` (now it passes a copy of the expconf dictionary to be consistent with `ExpDescriptionEditor.experimentConfigurationChanged`). This change was done in d3f60003b , and will be propagated to develop as soon as we merge #1009 (which should be imminent). I suggest to base any new implementation on the new signature (and use new-style signals)

droldan:
The #1021 PR Separates DynamiPlotManager from the expconf widget.

Take care that this PR is based on qt5, using the new signatures and new-style signals.
Could be merged as soon #1009 will be merged.

PR #1021, opened 2019-01-23
Separate DynamiPlotManager from the expconf

droldan:
Before this PR, the scan trends react to "experimentConfigurationChanged" Qt signal comming from the ExpDescriptionEditor widget.

With the changes in this PR the scan trends (DynamiPlotManager) are connected to the "experimentConfigurationChanged" Qt signal coming from the QDoor directly, emitting signals regardless of the client who made the changes.

Therefore, the scan trends are completely separated from the expconf widget.

**NOTE:**
This PR has been tested with Taurus TEP18 version using the **new QT signals.**

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
@@ -441,8 +441,6 @@ class ExpDescriptionEditor(Qt.QWidget, TaurusBaseWidget):
         for tg_info in tg_elements.values():
             avail_triggers[tg_info.full_name] = tg_info.getData()
         self.ui.channelEditor.getQModel().setAvailableTriggers(avail_triggers)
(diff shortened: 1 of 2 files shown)


droldan:
> IMHO, it is better to partially revert [630b75b](https://github.com/sardana-org/sardana/commit/630b75b28b26113fc00880804c47ad36278d30c6) so that only the connects are removed but the emits remain in place.
I agree with you.
Anyway, I've checked the Sardana and taurus code and I haven't found any widgets that use this signal, but well, your suggestion seems reasonable to me. Thanks @cpascual !

cpascual:
> I've checked the Sardana and taurus code and I haven't found any widgets that use this signal, 

I can confirm that this was the only usage of this signal within taurus & sardana... that's why I explicitly mentioned **3rd party** widgets  () ;)

droldan:
Continuing PR in develop branch #1034

PR #1022, opened 2019-01-23
Add backwards compatibility for measurement group plot configuration (fixes #1017)

### LINKED ISSUE #1017: empty Trend1D after upgrade of Sardana
Hi,

After upgrade to Taurus 4.4.0 and Sardana 2.5.0 we don't have Trend1D plots for Sardana scans.
I posted it to Taurus issues (https://github.com/taurus-org/taurus/issues/841), Carlos P suggested to move it here.

In meantime we have discovered that re-creation of measurement groups has solved the issue.

Whether this behavior is intended/expected or not, maybe there should be a messagebox that suggests to make all measurement groups anew?

reszelaz:
This PR fixes #1017 by adding back-comp for measurement group plotting configurations.

Measurement group plotting configurations created with sardana < 2.4.0 and taurus < 4.3.0 are not compatible after the upgrade. Fix it by using channel's full name URIs (tango scheme + FQDN).

@sardana-org/integrators it is ready to be reviewed, thanks!

@julianofjm, @kklmn your confirmation of the fix would speed up the integration process, thanks!
Do you need a hotfix release for sardana 2.4.0 or 2.5.0? Otherwise, soon we will release Jan19 milestone and the fix will be there anyway.

### DIFF
--- a/src/sardana/pool/poolmeasurementgroup.py
+++ b/src/sardana/pool/poolmeasurementgroup.py
@@ -717,6 +717,7 @@ class MeasurementConfiguration(object):
                 else:
                     if to_fqdn:
                         ch_name = _to_fqdn(ch_name, logger=self._parent)
+                        ch_data['full_name'] = ch_name
                     channel = pool.get_element_by_full_name(ch_name)
                 ch_data = self._fill_channel_data(channel, ch_data)
                 user_config_channel[ch_name] = ch_data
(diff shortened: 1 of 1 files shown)


julianofjm:
Hi @reszelaz , Thanks a lot. 
Sorry for the late reply, we don't need a hotfix because we simple recreated the measurement groups and it was enough to work around this issue.
I finally had time to review that and the fix worked. I reproduced one more time the issue described in #1017 
Then I tried the same using your branch related to this PR and the taurusgui Trend1D worked fine.
Regards,
Juliano

reszelaz:
Thanks a lot for your tests. I merge it then!
Sorry about the problem and that you had to recreate your measurement groups.

Issue #961, opened 2018-10-10
Add autoupdate checkbox to expconf 


reszelaz:
Now the expconf by default warns with a pop-up about the configuration changes coming from the server.
User may want to autoupdate the expconf with this changes.

This is now possible from spock if the expconf was started with the `--auto-update` flag.

However it would be more convenient to be albe to enable/disable auto update feature at runtime, for example, with a checkbox.

dschick:
I totally agree, especially when the expconf is started from the taurus gui, it is a bit harder to add the --auto-update flag, I guess

cpascual:
Also, this option should be a "configurable property" so that its value is remembered between runs of the GUI

reszelaz:
Solved in #1028

PR #1028, opened 2019-01-30
Add auto-update action (context menu) in expconf widget.

reszelaz:
As stated in #961 an access to the auto-update feature of expconf from the graphical interface is missing. Note that it is available when starting expconf in spock (--auto-update).

This PR adds a context menu to the expconf (also fixing some context menu issues of the measurement group tab) with an action to enable/disable auto update. Also a short documentation is added.

The autoupdate feature is saved in the applications settings, so in case of setting it in a TaurusGUI this setting will persist (it will be reapplied on the GUI restart). It is pending to add the same feature when expconf is started from spock.

I tested it both on Qt4 and Qt5 and it worked fine for me and I see it ready for review and eventual integration. I see it quite important to be included in the Debian10 release and the Jan19 milestone. @sardana-org/integrators any volunteers? Thanks!

### DIFF
--- a/doc/source/users/taurus/experimentconfiguration.rst
+++ b/doc/source/users/taurus/experimentconfiguration.rst
@@ -9,7 +9,7 @@ Experiment Configuration user interface
 
 .. contents::
 
-Experiment Configuration widget a.k.a. expconf is a complete interface to
+Experiment Configuration widget a.k.a. ``expconf`` is a complete interface to
 define the experiment configuration. It consists of three main groups of
 parameters organized in tabs:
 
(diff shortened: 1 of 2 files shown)


teresanunez:
Hi @reszelaz, I have tested this (debian9, qt4) and it works as described. The only point, but perhaps it is a problem with my display/GUI, is that it is not easy to know where the box to activate it is ... I search
 for it becaus I knew that it should be somewhere ... but if not I would never know that this possibility
 exist. But as I said, it can be something wrong with my display ... I don't know.

reszelaz:
> perhaps it is a problem with my display/GUI, is that it is not easy to know where the box to activate it is

It is not a problem of you display/GUI. When you open expconf there are three areas:
* expconf area
* measurement group area
* channels editor area

expconf-areas

Each of them have different actions available in the context menu. So respecting the same order:
* enable/disable auto update
* enabled/disable online plotting
* enabled/disable simple view

So the "auto update" feature is accessible in the expconf area since it affects/refers the whole expconf widget.

We have different options to improve the user experience here.
1. Rearange the options in the context menu.
2. Add a special tab for expconf settings/configuration.

But I would treat it in a separate issue.

PR #1033, opened 2019-02-01
Make auto update action correctly restore its settings

reszelaz:
Register config property with action's setter/getter so it is correctly
restored from the setting.


cpascual:
I didnt test the feature itself, but I did a code review and it all seemed ok to me.

reszelaz:
Thanks for the review. I'm mergint then..

PR #1041, opened 2019-02-06
Warn when hooks gets overridden

reszelaz:
This is related to #962 and #1013.
On one of the internal meetings at Alba @srgblnch commented that the `Hookable.hooks` property setter could warn about the fact of overriding general hooks. What is your opinion to add it?

I already created a PR for that, the output looks like this:

```
Door_zreszela_1 [12]: defgh ct pre-acq
Defining general hook
ct

Door_zreszela_1 [13]: lsgh
   Hook place   Hook(s)
 ------------ ---------
      pre-acq        ct

Door_zreszela_1 [14]: captain_hook 0 10 1
previously set hooks, these may include general hooks, are being overridden now
Starting loop
At step 0
running hook with hints=['pre-acq']
        hook execution
At step 1
running hook with hints=['pre-acq']
        hook execution
At step 2
running hook with hints=['pre-acq']
        hook execution
At step 3
running hook with hints=['pre-acq']
        hook execution
At step 4
running hook with hints=['pre-acq']
        hook execution
At step 5
running hook with hints=['pre-acq']
        hook execution
At step 6
running hook with hints=['pre-acq']
        hook execution
At step 7
running hook with hints=['pre-acq']
        hook execution
At step 8
running hook with hints=['pre-acq']
        hook execution
At step 9
running hook with hints=['pre-acq']
        hook execution
Finished loop
```

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -265,7 +265,7 @@ class Hookable(Logger):
             return
 
         if len(self.hooks) > 0:
-            self.warning("previously set hooks, these may include general"
+            self.warning("previously set hooks, these may include general "
                          "hooks, are being overridden now")
         # store self._hooks, making sure it is of type:
         # list<callable,list<str>>
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -264,6 +264,9 @@ class Hookable(Logger):
                 'the hooks must be passed as a list<callable,list<str>>')
             return
 
+        if len(self.hooks) > 0:
+            self.warning("previously set hooks, these may include general"
+                         "hooks, are being overridden now")
         # store self._hooks, making sure it is of type:
         # list<callable,list<str>>
         self._hooks = []
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -265,8 +265,11 @@ class Hookable(Logger):
             return
 
         if len(self.hooks) > 0:
-            self.warning("previously set hooks, these may include general "
-                         "hooks, are being overridden now")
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi, I agree with the idea of printing a warning message when the hooks are overridden, but I would change the output message. I think the one proposed here could be confusing for the users (not clear
who has overridden the already defined hooks - that in fact, i think can only be general hooks,
because any other one is only affecting the macro defining it - why and if they are overridden in general or only for the current running that is called. I don't know if it would be too much explanation but I would put something like: 'This macro define its own hooks. Any other previously defined hooks, including the general ones, would not be called. Use the method appendHook - see documentation - if you want to keep them'. What do you think?

reszelaz:
I'm fine with your suggestion.

But now thinking, in general, there is one special case when this message may be misleading. For example if one program like this:

```python

def my_hook():
    pass

[in a macro code]
hooks = self.hooks
hooks.append((my_hook, ["pre-acq"]))
self.hooks = hooks
```

teresanunez:
Hi @reszelaz, you are right, and if we put something like:

'This macro defines its own hooks. Previously defined hooks, including the general ones, would be only called if these own hooks were added used the appendHook method or appended to the self.hooks.'

By the way, is this other way of appending hooks documented?

reszelaz:
I'm fine with the proposed name. Quite long but indeed more explicative. I'll change it now.

> By the way, is this other way of appending hooks documented?

No, it is not. Do you think it is necessary to document this? It comes from python itself.

teresanunez:
Hi @reszelaz, many thanks. No, perhaps it is not necessary to document it.

reszelaz:
I'm merging this since it was accepted by @teresanunez.

PR #1049, opened 2019-02-15
Adapt ct and uct to accept timerable channels

reszelaz:
#997 introduced independent acquisition of experimental channels.
This PR allows to count them with `ct` and `uct` macros. Now one can do `ct 0.1 ct01` or `uct 0.4 twod01 `

An alternative solution would be to add a dedicated macro e.g. `cti` to be able to count in a manner similar to the `mv` macros e.g. `cti ct01 0.1` or `ucti twod01 0.4` (`i` from independent). But maybe this may introduce too many macros for similar purposes? What are your preferences?

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -648,6 +648,15 @@ class tw(iMacro):
 ##########################################################################
 
 
+def _value_to_repr(data):
+    if data is None:
+        return "<nodata>"
+    elif np.rank(data) > 0:
+        return list(np.shape(data))
+    else:
+        return data
+
+
 class ct(Macro, Hookable):
     """Count for the specified time on the active measurement group"""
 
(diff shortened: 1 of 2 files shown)


teresanunez:
Hi @reszelaz, thanks for the implentation, I have just test it and it works nicely. The only problem I see is that if ActiveMntGrp is not defined one gets the error:

Error: The macro ct requires the following missing environment to be defined: ['ActiveMntGrp']

what in fact is not true if one gives the countable argument. I think this condition should be relaxed, and an error message should appear if not countable is given and the ActiveMntGrp is not defined.
I also think that the text in the error:

if countable_elem is None:
        self.error('The MntGrp {} is not defined or has invalid '
                       'value'.format(self.countable_elem_name))

should be changed and use Contable instead of MntGrp.

About creating the macro 'count' instead of using 'ct' with a counter as argument ... I also think that it is defining a new macro for similar purpose and even could be that the people doesn't realised that it
 exists.
Thanks again.
                                 Regards,
                                                      Teresa

reszelaz:
Thanks @teresanunez for the review. I just fixed it and now it is not necessary to specify the `ActiveMntGrp` environment variable. I also corrected the error message.

Regarding a dedicated macro, I think that the `count` name is not the best for exactly the reason you point:
> even could be that the people doesn't realised that it exists

So, I changed the initial proposal to `cti` (`i` from independent) so one could execute: `cti ct01 0.1` or `ucti twod01 0.4`.

guifrecuni:
> What are your preferences?

I think it is better to use the same macros for counting so:
ct 0.1 (using the active measurement group)
ct 0.1 mntgrp02 (providing a different measurement group)
ct 0.1 ct02 (provinding just a specific counter)

reszelaz:
Seems like there is a consensus in the proposal already implemented in this PR and it was accepted by @teresanunez so I proceed with the merge. If you have comments on that or still prefer the second option of dedicated macros you are always on time to rise your hand. Thanks to all!

Issue #1058, opened 2019-02-18
Jan19 manual release tests: Windows 10


daneos:
### Installation
- [x] Install Sardana (on Linux from the tar.gz : `pip install <tarball_artifact_URL>`
  and on Windows from MSI)
  **Note:** On openSuse 11.1 there are problems with pip, try `python setup
  .py install`

### Create testing environment and run testsuite
- [x] Start Pool demo2. In a console do `Pool demo2`.
- [x] Start MacroServer demo2 and connect to the Pool demo2.
  In another console do: `MacroServer demo2`
- [x] Set MacroServer's MacroPath to point to the macro examples.
  In another IPython console do:
  `PyTango.DeviceProxy('macroserver/demo2/1').put_property({'MacroPath':'<path_to_sardana_installation_dir>/macroserver/macros/examples'})`
  **Note:** Remember to use OS path separator e.g. '/' on Linux and '\' on 
  Windows
- [x] Restart MacroServer e.g. Ctrl+C in the MacroServer's console and
  start it again.
- [x] Create spock profile demo2. In another console do `spock --profile=demo2`
**- Problem1: traceback appearing when creating new spock profile:**
https://gist.github.com/mrosanes/e8473156e350ba2d2425a089d0d19f15
**Bug also existed in previous releases:** https://github.com/sardana-org/sardana/issues/536
**- Problem2: Traceback related to matplotlib and PyQt5 (even if PyQt5 is not installed)**
https://gist.github.com/mrosanes/e87b970edb6b6fac23384477a7773d24
**Similar bug existed in previous releases:** https://github.com/sardana-org/sardana/issues/536
**- Problem3: cannot exit spock doing 'exit()', even if after I use 'Ctrl+C'**

- [x] In spock run `sar_demo` macro.
- [x] Edit `<path_to_sardana_installation_dir>/sardanacustomsettings.py`
  to point to the demo2 door e.g. `UNITTEST_DOOR_NAME = "door/demo2/1"`
- [x] Run testsuite. In another console do `sardanatestsuite`
  **Note:** On openSuse 11.1 and Windows there are known problems with 
  testsuite. Check previous release comments.
```
Ran 228 tests in 1415.110s
FAILED (failures=1, skipped=7, expected failures=1)
```
[Full testsuite output](https://gist.github.com/mrosanes/05392d12e9c61bbc7364b6a12c3372ff)

### Test Sardana using Spock and expconf
[...]

PR #1060, opened 2019-02-19
Remove ms properties file in tests correctly on Windows

reszelaz:
@mrosanes discovered this in #1058. @mrosanes could you verify if this works on Windows and still on Linux, plese?


mrosanes:
Tested PR and it works well in Windows and Linux running: 
The macroserver.properties files is now removed correctly after this PR.

testing using tests from:
python -m unittest sardana.macroserver.macros.test.test_gh

reszelaz:
Thanks @mrosanes!
@daneos , @teresanunez I'm merging. I think it is a minor change and should not affect the release process. But at least it let us better understand the testsuite output on Windows. Feel free to revert...

reszelaz:
This is not the end of the problems on windows. We discovered that the MacroServer processes started for the needs of the tests does not die correctly. This prevents the said file to be deleted. We will investigate it tomorrow but if we won't manage to get it fix, I propose to skip this file deletion on windows.

PR #1061, opened 2019-02-20
Prepare channel in independent acquisition

reszelaz:
Measurement preparation (introduced in SEP18) is not executed
in independent acquisition of experimental channels. Do it.

One can test it by adding a dummy `PrepareOne` methodn to a controller, executing the independent acquisition and observing if it was called:

```python
def PrepareOne(self, axis, value, repetitions, latency, nb_starts):
    self._log.debug("PrepareOne")
```
@sardana-org/integrators can you take a look please?


reszelaz:
Thanks @teresanunez for the review. Travis-ci is green. So I'm merging.

PR #1069, opened 2019-02-28
clear the event set, before macro commnads

mrosanes:
Clear the event set before running execution and before
macro stop/abort.

This is solving bugs found in Windows, where the spock prompt was
returned before expected, as events arrived with the same
timestamp, cause of poor time resolution of python on windows.


daneos:
I tested this on Linux and works OK. If it's fixing the issue on Windows, then I'm merging this.

reszelaz:
Great! Thanks!

PR #1071, opened 2019-03-01
Avoid memory errors on testsuite execution

### LINKED ISSUE #1067: Test suite memory errors
This is an issue found during manual tests of Jan19 release on CentOS 7 (#1057).

The test suite fails with many `Cannot allocate memory` errors, despite having ~3GB of free RAM.
I tried to isolate the problem with a smaller subset of tests, but it seems it only shows up when the full suite is run.
You can find the full output [here](https://gist.github.com/daneos/c04ca418d5567a17b06d2713cdf86bab).

It doesn't look like a problem with Sardana, more like with the test framework.

@amilan, could you check if the problem is also present on your CentOS environment?

reszelaz:
Dummy 2D controller generates an array of zeros (1024x1024) for each
axis, many axes may increase the VM memory consumption of testsuite and
this apparently causes OSError: [Errno 12] Cannot allocate memory on
the os.fork() call. Avoid this by just creating one axis of dummy 2D.

Fixes #1067.

### DIFF
--- a/src/sardana/pool/test/base.py
+++ b/src/sardana/pool/test/base.py
@@ -139,7 +139,7 @@ class BasePoolTestCase(object):
             ctrl_obj = self.createController(name,
                                              'DummyCounterTimerController',
                                              'DummyCounterTimerController.py')
-            # Create nelems CT elements for each ctrl
+            # Create nctelems CT elements for each ctrl
             for axis in range(1, self.nctelems + 1):
                 name = '_test_ct_%s_%s' % (ctrl, axis)
                 self.createCTElement(ctrl_obj, name, axis)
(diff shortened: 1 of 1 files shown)


reszelaz:
One can check how the virtual memory raises before and after this fix with:
`python -m unittest sardana.pool.test.test_measurementgroup &>/dev/null & top -p $!`

src/sardana/pool/test/base.py
```diff
@@ -171,13 +175,13 @@ def setUp(self):
...
         # Create nctrls MOT ctrls
-        for ctrl in range(1, self.nctctrls + 1):
+        for ctrl in range(1, self.nmotctrls + 1):
             name = '_test_mot_ctrl_%s' % ctrl
             ctrl_obj = self.createController(name,
                                              'DummyMotorController',
                                              'DummyMotorController.py')
             # Create nelems CT elements for each ctrl
-            for axis in range(1, self.nctelems + 1):
+            for axis in range(1, self.nmotelems + 1):
```
cpascual: Hi, I was analysing the changes and this one surprised me. Did you change `self.nctctrls` by `self.nmotelems` intentionally **in this case**?
I see a change before where the same change makes sense according to the comment above, but in this case the comment refers to "CT elements"

reszelaz: Thanks! The code is correct. Te comment also needs to be changed to "create nelems MOT elements...".
And the one above regarding the trigger gate element needs to be changed similarly.

cpascual: ok. I'll do the changes and merge then

PR #1072, opened 2019-03-04
Fix typo in test exception strings

mrosanes:


daneos:
There is some strange problem with cloning in Travis, but I doubt it has anything to do with changes in this PR. @sardana-org/integrators if you agree I would merge this.

PR #1075, opened 2019-03-06
Get ms_properties from db properties

mrosanes:
Removing macroserver.properties file sometimes is giving problems in Windows; and because of that,  errors in the sardanatestsuite increases a lot.
Add try/except to remove macroserver.properties file.


reszelaz:
Thanks @mrosanes! The changes looks ok!
I merge them. @daneos, @teresanunez feel free to revert if you consider it problematic for the release.

PR #1076, opened 2019-03-08
Fix expconf when empty (unspecified) DataType

### LINKED ISSUE #1070: HDF file does not save channel if data type column is empty in measurement group.
If the measurement group has channel with the "Data Type" column empty and the file is configured to be HDF (with .h5 extension) the channel won't be saved in the file. 
It appears as a column in spock (Door output) but not in the file.
This issue doesn't happen with .dat files.

reszelaz:
Empty _DataType_ field in expconf is considered as unspecified `data_type` configuration. Pop this configuration key so the _downstream_ software can consider it as unspecified and eventually use default values. Setting it to `None` could be reserved for unknown or erroneous conditions.

Fixes #1070.

@julianofjm could confirm that it fixes your problem? How important is this bug to you? Do you need it in an official release? Or merging into develop would be enough?

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py
@@ -330,7 +330,9 @@ class MntGrpChannelItem(BaseMntGrpChannelItem):
         return ret
 
     def setData(self, index, qvalue):
+        ch_name, ch_data = self.itemData()
         taurus_role = index.model().role(index.column())
+        key = self.itemdata_keys_map[taurus_role]
         if taurus_role in (ChannelView.Channel, ChannelView.Conditioning,
                            ChannelView.NXPath, ChannelView.DataType,
                            ChannelView.Enabled, ChannelView.Output):
@@ -353,8 +355,6 @@ class MntGrpChannelItem(BaseMntGrpChannelItem):
                 data = ()
         else:
             raise NotImplementedError('Unknown role')
-        ch_name, ch_data = self.itemData()
-        key = self.itemdata_keys_map[taurus_role]
         ch_data[key] = data
 
     def role(self):
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py
@@ -340,7 +340,10 @@ class MntGrpChannelItem(BaseMntGrpChannelItem):
         elif taurus_role == ChannelView.DataType:
             if len(qvalue.strip()) == 0:
                 # empty strings are considered as unspecified data type
-                ch_data.pop(key)
+                try:
+                    ch_data.pop(key)
+                except KeyError:
+                    pass  # data_type key may not be there if not specified
                 return
         elif taurus_role == ChannelView.PlotType:
             data = PlotType[qvalue]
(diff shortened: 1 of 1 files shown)


julianofjm:
Hi @reszelaz . Thanks a lot for the reply. It is not a critical issue for us. We found it out by chance when one of the beamlines configured a measurement group with empty DataType column. Filling in the column solves the issue temporarily, so that is fine to just merge into develop and have this fix in a future release.

I tried to confirm it but I get an error when trying to set the dataType column via expconf:

`Traceback (most recent call last):

  File "/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py", line 595, in setData
    item.setData(index, qvalue)
  File "/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py", line 363, in setData
    ch_data[key] = data
UnboundLocalError: local variable 'data' referenced before assignment`

Apart from that, using the old measurement group configured before containing a channel with empty dataType seems that the problem still persists, I couldn't see the channel with empty dataType into the HDF file.

One more information is that I could only reproduce this error when creating new measurement groups. Using measurement groups created by sar_demo ("mntgrp01", for example) the issue doesn't appear and the channel is saved, even with dataType column empty.

reszelaz:
Thanks for the feedback @julianofjm !

> I get an error when trying to set the dataType column via expconf:

I just fixed this.

I also fixed another bug that was happening when entering to "Data Type" editor where no data type was set and then leaving the editor without setting any data type.

> Apart from that, using the old measurement group configured before containing a channel with empty dataType seems that the problem still persists, I couldn't see the channel with empty dataType into the HDF file.

Yes, it is the case. Before, when a user was leaving empty "Data Type" field in expconf, behind the scene this was setting the `data_type` channel's configuration parameter to `None`. This PR changes this behavior, so now the expconf simply removes this configuration parameter indicating that it is unknown.

This way, if the configuration parameter is not there, the recorder (or ony other part of the system relying on this configuration) may apply its own choice to assign one. I propose to reserve `data_type=None` for indicating abnormal/erroneuous cases.

Based on that, the measurement groups where one explicitly removed the "Data Type" configuration using the expconf before this change are in king of _broken_ state. One would need to set a data type by hand to recover them. But, this is only for the cases where one explicitally removed this configuration parameters. By default, when channel was (and is) added to the expconf it contains the corresponding data type configuration parameter. So I think it is not that bad and we could eventually leave it like this. I suspect that there are not that many users touching this configuration parameter.. What do you think?

> One more information is that I could only reproduce this error when creating new measurement groups. Using measurement groups created by sar_demo ("mntgrp01", for example) the issue doesn't appear and the channel is saved, even with dataType column empty.

The measurement groups created with `sar_demo` are created using the `defmeas` macro. The measurement group created in this way is directly created on the server without applying measurement group configuration. Then it misses the `data_type` configuration paramter. In contrary, when a measurement group is created in expconf, first, it is created on the server and the the expconf applies a full measurement group configuration which contains also the `data_type` parameter. Maybe at some point we should think of having this unified...

julianofjm:
Hi @reszelaz 
Thanks! 
I tested your new modification and it works fine. I didn't get the error any more (to set the dataType) and when forcing the dataType column to be empty the channel is still saved into the HDF file, so I think we can consider this issue solved.
I also agree that is fine if old measurements stay in a broken state because of this empty configuration, I saw only once this parameter changed/removed in our measurement group configurations and now we are aware of this.
As I said, since it is not very critical, it is okay for us to wait the next official release to have this fix.
Thanks again and regards,
Juliano

reszelaz:
Since @julianofjm confirms that the bug is fixed and this is a minor change I proceed with the merge directly.

Issue #1081, opened 2019-03-21
set the significant digits of a counter value


dschick:
Hej,

I was wondering how many significant digits sardana handles for counter values?
I am trying to store long integer into the .spec file and I miss the last three digits.
Is there a way how to adjust that?

Best

Daniel

reszelaz:
Sorry for late reply @dschick!

First of all, I think there is no way to parametrize the way values are written by  SPEC recorder.

I still did not have time to try to reproduce this problem - I will try it next week.

I just quickly looked at the code and just in case you want to investigate on your own in parallel I think that this line https://github.com/sardana-org/sardana/blob/aac4c20090b350fe94f5a9563c4e21c262c7cbab/src/sardana/macroserver/recorders/storage.py#L476 is responsible for formatting the channel values.

dschick:
Thanks  @reszelaz for the comment.

I am actually trying to save a timestamp of a 2d detector within the SPEC file as integer as some kind of option 3, see #991 .
So I am trying to do something like 20190326125038999 where the last three digits are the miliseconds of the timestamp. Unfortunately, these three last digits are cut off.

I guess I could change the `str(data)` to some kind of `"".format()` but this would only make sense with have some general parameter available to set the formatting in the spec file.

Best

Daniel

reszelaz:
@dschick thanks for further details and the example!

To solve this problem, first, we need to remember that the counter/timer values are floats and not integers.
Second, we would need to find a way of printing this value with pure Python. I tried the following and we lose the miliseconds precission:

```python
In [15]: i = 20190326125038999.0
In [16]: "{:17.0f}".format(i)
Out[16]: '20190326125039000'
```

I'm a little bit busy these days so it would be great if you could investigate this.

Whenever we find a way of doing it with Python, you could use a custom recorder (inheriting from SPEC recorder) where you could implement this formatting change. Adding a parametrization to SPEC recorder is another option but will certainly require more work.

reszelaz:
> To solve this problem, first, we need to remember that the counter/timer values are floats and not integers.

I completelly forgot about the possibility of [changing the default interface](https://sardana-controls.org/devel/howto_controllers/howto_controller.html#changing-default-interface).

So, you simply need to add this method in your counter/timer controller:

```python
    def GetAxisAttributes(self, axis):
        axis_attrs = CounterTimerController.GetAxisAttributes(self, axis)
        axis_attrs = dict(axis_attrs)
[...]

PR #1083, opened 2019-03-27
Change mapping of Integer data type to Tango DevLong64

reszelaz:
Integer mapped to DevLong allows to pass integer numbers of 2^31 - 1.
Change mapping to DevLong64 to allow passing numbers of 2^63-1.

This change does have a slight memory footprint effect on py2 32 bits.
There DevLong is simply int and DevLong64 is long. On py2 64 bits
both are int. On py3, regardless of 32 or 64 bits arch, there are only int,
so there is no effect.

This change does have a slight network footprint effect. Tango read
of DevLong attribute requires 160 bytes frame to be transfered while
DevLong64 attribute requries 168 bytes frame. Tango event (ZMQ) of
DevLong attribute requires 226 bytes frame to be transfered and
DevLong64 requires 230 bytes frame to be transfered.

---

One could also think of parametrizing this in `sardanacustomsettings` or even adding new data type to sardana... But I think that the network transfer oeverheads in range of 5% in case of read and 2% in case of event are very small and it is not worth. But if you have another opinion just comment! Thanks in advance!

This is needed for #1081

### DIFF
--- a/src/sardana/tango/core/util.py
+++ b/src/sardana/tango/core/util.py
@@ -51,7 +51,7 @@ import itertools
 
 import PyTango
 from PyTango import Util, Database, WAttribute, DbDevInfo, DevFailed, \
-    DevVoid, DevLong, DevBoolean, DevString, DevDouble, \
+    DevVoid, DevLong64, DevBoolean, DevString, DevDouble, \
     DevState, SCALAR, SPECTRUM, IMAGE, FMT_UNKNOWN, \
     READ_WRITE, READ, Attr, SpectrumAttr, ImageAttr, \
     DeviceClass, Except
@@ -384,7 +384,7 @@ def from_tango_state_to_state(state):
 
 #: dictionary dict<:class:`sardana.DataType`, :class:`PyTango.CmdArgType`>
 TTYPE_MAP = {
-    DataType.Integer: DevLong,
+    DataType.Integer: DevLong64,
     DataType.Double: DevDouble,
     DataType.String: DevString,
     DataType.Boolean: DevBoolean,


reszelaz:
@sardana-org/integrators can someone from you take look on this? Thanks!

teresanunez:
Hi @reszelaz, I have tested it and IMO it can be merged. Thanks.

jairomoldes:
I have experienced a problem with this change. A controller with a memorized attribute of type DataType.Integer will fail when setting the memorized value. It may be a side effect of other changes or maybe the machine where sardana is running (32 bits python2.6 tango 7.26)

I detected the problem when switching from sardana 3.24 to 3.25 bliss versions (this is internal to alba). I think that the diff between these 2 versions can be retrieved using:
git diff 2.8.2 1f8ebb3

reszelaz:
Thanks @jairomoldes for the report. Have you tried it with a normal DS to see if a DevLong64 memorized attribute works correctly on the same machine?

> It seems that the problem does not happen with a debian 9 with tango 9 and python3 sardana3.0.2a0+34.9f6895-1~bpo9+1

This was on a 64 bits machine, right?

jairomoldes:
> Thanks @jairomoldes for the report. Have you tried it with a normal DS to see if a DevLong64 memorized attribute works correctly on the same machine?

You are right. The problem is in [py]tango. A normal pytango DS also fails to recover memorized values for a DevLong64 attribute. Probably this is specific to tango 7.2.6. It should be investigated, but seems clear that the problem is not in sardana but in tango or pytango

> 
> > It seems that the problem does not happen with a debian 9 with tango 9 and python3 sardana3.0.2a0+34.9f6895-1~bpo9+1
> 
> This was on a 64 bits machine, right?

Yes.

reszelaz:
I have just patched ALBA bliss package to revert this PR. I would not do to much about it in Sardana upstream, at least for the moment. Could you please report it to Tango?

PR #1090, opened 2019-04-04
Hotfix for the measurement group configuration

rhomspuron:


reszelaz:
Another minor thing: according to [gitflow](https://nvie.com/posts/a-successful-git-branching-model/) and how we have done it so far the branch name should be hotfix-2.7.2.
Also the hotfix branch must bump the version number, recreate manpages, etc. but don't worry about it. I can do it whenever we agree on the hotfix scope and accept it.

rhomspuron:
I can change the synchronizer without problem, but it is not possible to disable the channels.

rhomspuron:
I found another problem, it is possible to disable all channels. Should we protect it?

reszelaz:
> I found another problem, it is possible to disable all channels. Should we protect it?

Yes, but maybe on the server side when starting the measurement group. If there are no enabled channels it should give an exception.

Any other opinions?

rhomspuron:
It's a good idea.

reszelaz:
There is a problem when one tries to set a timer/monitor of a controller to a channel which is not the first in the list of channels proceeding from the said controller, e.g.

channel | timer |
-----------|--------|
... | ...
ct02 | ct01
ct01 | ct01
...|...

I have found that this can be workaround by copying timer/monitor configuration in `set_configuration_from_user`:
```python
            [...]
            if external:
                ctrl_item = ExternalControllerConfiguration(ctrl)
            elif ctrl.is_timerable():
                is_software = synchronizer == 'software'
                acq_synch = AcqSynch.from_synch_type(is_software,
                                                     synchronization)
                ctrl_acq_synch[ctrl] = acq_synch
                ctrl_conf["timer"] = ctrl_data.get("timer")
                ctrl_conf["monitor"] = ctrl_data.get("monitor")
                ctrl_item = TimerableControllerConfiguration(ctrl, ctrl_conf)
            else:
                ctrl_item = ControllerConfiguration(ctrl, ctrl_conf)
            [...]
```

rhomspuron:
I tested on the next conditions:
* Two dummy counter controllers with one channel each one and the same synchronization hardware. Do not use software synchronization and run %ct.
* Three dummy counter of the same controller ct01, ct02 and ct03. ct01is the timer and ct03 is the monitor, disabled ct01 and run %ct
* Try to disable the channel used as timer and monitor, without disabled the other channels of the controller. 
* All channels disabled and run %ct.
* Two dummy counter controllers with two channels each one and use one of the channel as timer and the another as monitor. Disabled the timer channel and run %ct.

reszelaz:
Thanks @rhomspuron, this looks very promising. I just modified some messages and string formatting {} to {0} (we still need to support python 2.6).

I found one problem.
I set a controller's timer to a channel and applied this configuration. Then I removed the said channel from the measurement group and applied again the configuration. This raises the following pop-up: _Measurement Group mntgrp07: AttributeError: 'str' object has no attribute 'full_name'_
which is not very descriptive. Also what worries me is that there is no traceback logged in any of the processes so I don't know from where it comes...

reszelaz:
Now everything looks ok. However I close this PR cause I opened another one https://github.com/sardana-org/sardana/pull/1118 to comply with the branching naming as commented in https://github.com/sardana-org/sardana/pull/1090#issuecomment-481187779

PR #1094, opened 2019-04-10
Modify expconf popup messages

mrosanes:
Proposal of a pull request to modify messages in expconf configuration modification popup, to
improve user experience. Simplified the messages by removing technical concepts as 'Door', etc.


mrosanes:
I agree with the proposed changed. Changes applied after review.

reszelaz:
Great, merging!

PR #1099, opened 2019-04-17
Show external changes pop-up in expconf when last measurement group is deleted remotelly

reszelaz:
This fixes case when the last measurement group is removed remotely and no pop-up with external
changes appears in expconf. It also clears the measusrement group configuration table when active meas group is deleted.

@sardana-org/integrators can you take a look on this? Thanks!


reszelaz:
Thanks for the review @teresanunez! I'm merging..

PR #1100, opened 2019-04-17
Emulate hardware triggering between dummy elements

reszelaz:
Up to now hardware synchronization modes (trigger/gate/start) of dummy counter/timer works but these types of acquisitions starts already on `StartAll` method call. In case of the continuous scan this makes the acquisitions finish too early - not respecting the acceleration time of the moveable. This can be observed on this continuous scan trend of motor position and counter/timer state.
This was ascanct of dummy motor (vel: 100, acc = 2, dec = 2, base_rate = 0) from 0 to 10 with 10 intervals and 0.1 second integration time. The measurement group contains just a dummy counter/timer synchronized by a dummy trigger/gate in either of trigger/gate/start synchronization modes.

Different approaches exist to emulate the medium of propagation of hardware synchronization events between the dummy trigger/gate element and dummy counter/timer element e.g. socket, Linux signal, or Sardana kernel events. The Linux signalling looks like not platform independent. The socket connection may require quite a lot of development if we would like to support multiple acquisitions running at the same time (managing the cache of port numbers for connections, etc). The Sardana kernel events would do the job equally good and these are the simplest to implement even if these require some hacks in the code. So I have chosen this last option and this PR proposes the implementation. The results can be observed on 
the equivalent continuous scan trend - see how the counter/timer state changes back to on when the scan active range finishes.

Of course in the future we could change the medium of propagation of events and this will be totally transparent.

Summary of changes:
* add `Synchronizer` attribute to dummy counter/timer controller
* implement synchronization (real start of acquisition) on first active event (trigger)
* automatic connection of dummy counter/timer controller with dummy trigger/gate element in the `sar_demo` macro.

@sardana-org/integrators could you please take a look. This feature will help in SEP2 development.

### DIFF
--- a/src/sardana/macroserver/macros/demo.py
+++ b/src/sardana/macroserver/macros/demo.py
@@ -172,6 +172,11 @@ def sar_demo(self):
         self.print("Creating trigger element", tg_name, "...")
         self.defelem(tg_name, tg_ctrl_name, axis)
 
+    ct_ctrl = self.getController(ct_ctrl_name)
+    ct_ctrl.getAttribute("synchronizer").write(tg_name)
+    self.print("Connecting trigger/gate element", tg_name,
+               "with counter/timer controller", ct_ctrl_name)
+
     self.print("Creating IORegister controller", ior_ctrl_name, "...")
     self.defctrl("DummyIORController", ior_ctrl_name)
     for axis, ior_name in enumerate(ior_names, 1):
--- a/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py
+++ b/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py
@@ -336,10 +336,10 @@ class DummyCounterTimerController(CounterTimerController):
         pool = pool_ctrl.pool
         try:
             synchronizer_obj = pool.get_element_by_name(synchronizer)
-        except:
+        except Exception:
             try:
                 synchronizer_obj = pool.get_element_by_full_name(synchronizer)
-            except:
+            except Exception:
                 msg = "Unknown synchronizer {0}".format(synchronizer)
                 raise ValueError(msg)
         self.__synchronizer_obj = synchronizer_obj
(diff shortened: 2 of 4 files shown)


reszelaz:
Travis shows problems with docs:

```
Warning, treated as error:
WARNING: intersphinx inventory 'https://pythonhosted.org/guiqwt/objects.inv' not fetchable due to <class 'urllib2.HTTPError'>: HTTP Error 504: Gateway Time-out```

This has nothing to do with this PR.

teresanunez:
Hi @reszelaz, I have some question, is the scan from -10 to 10 or from 0 to 10 as you wrote?, in this
 last case, why the counter is going to MOVING already at position -10?. I think I got something wrong.

reszelaz:
The scan indeed is starting at 10. However the experimental channels gets to Moving state already on call to `StartAll`. The measurement group is started before the moveable. So already at Position - 10 the experimental channels are in Moving. From then up to the first trigger these are armed - I reflected that with the status. We must do it like this, the channels must be moving - involved in an acquisition, otherwise someone else could start another acquisition with them - this could happen if we wait with setting the Moving state until the first trigger. The real hardware controllers are programmed the seme way.
I think this is confusing, and could be improved in the future if we develop more states, for example _armed_ . I already have some ideas for that. But this would be a major development.

teresanunez:
Hi @reszelaz, thanks, it is now clear to me. I think it is not a problem to keep the  MOVING state like it is, this is also how we are doing in our hardware triggered tango devices. Thanks for the implementation.

reszelaz:
Thanks for the review. I will proceed with the merge..

> I think it is not a problem to keep the MOVING state like it is..

Yes, I agree, the Tango State will change to MOVING as soon as the device is armed and waiting for external triggers. I was thinking in adding something similar to `MacroStatus` attribute of the Door device. There we export all the additional information which does not fit into the Tango State. But as I said this will require more development. For the moment let's keep this additionatl information in the Status.

PR #1102, opened 2019-05-06
avoid deepcopy of DeviceProxy

jkotan:
I should fix one of the reason of sardana hanging #1023. 
I replaces `deepcopy` of `DeviceProxy` by a simple reference copy.


jkotan:
it looks like TEST='doc' hangs.

PR #1103, opened 2019-05-09
Remove setting of global measurement group synchronizer

reszelaz:
Global measurement group synchronizer (even if not a bad idea) is
not used. Remove its random setting (to the last selected synchronizer)
so it does not raise a pop-up with external changes in expconf.

@sardana-org/integrators can you take a look, please?


teresanunez:
I have checked it and the change avoids the popup window.

PR #1105, opened 2019-05-14
Add scan macro example with a custom column with motor positions shifted to the middle of interval

reszelaz:
This macro demonstrates how to add an extra scan column with the shifted positions of the motor corresponding to the middle of the space interval.

Be aware that the space interval does not necessarily correspond to the middle of the acquisition interval
(remember about the latency time).

This macro does not export all the ascanct features e.g. hooks, ...


teresanunez:
I have tested it and it works as described. iMO it can be merged.

reszelaz:
Thanks @teresanunez ! Merging..

Issue #1107, opened 2019-05-16
Length of description displayed with lsdef


teresanunez:
One of our scientist wants to have the possibility of displaying the full description of the macro
with the macro lsdef, Up to now the description displayed is restricted to 60 characters (default
value of the argument max_chars in the function get_brief_description defined in sardanameta.py). But it could be easily increased adding an argument to the call:

m.get_brief_description()

done in the macro lsdef. Would that be a problem  for someone?. 
I also thouhgt to put the length as optional argument of the macro, but I think our users would prefer not put any extra argument for getting the right display.

reszelaz:
Looks ok to me to add this feature.

Maybe it fits better into to the [ViewOptions](https://sardana-controls.org/users/spock.html#changing-appearance-with-view-options)? What do you think?

Do you also see a possibility to show the full description but breaking it into multiple lines? This however may be a little more tricky to implement with the table like output...

teresanunez:
Hi @reszelaz, thanks. Yes, the ViewOptions could fit better, it would be also ok for me.
 I will have a look to the possible implementation. Breaking it into multiple lines would be nicer,
 I think, but let's see the implementaion ... I will check it.

teresanunez:
I will implement it as ViewOption, thanks.

reszelaz:
Done by #1108

PR #1108, opened 2019-05-16
Issue #1107: description length lsdef macro

teresanunez:


teresanunez:
The displayed description is divided in lines.

PR #1109, opened 2019-05-17
[WIP] Qtspock

schooft:
This PR implements a Qt widget with Spock functionality. It is based on the Juypter Qt Console example [here](https://github.com/jupyter/qtconsole/blob/master/examples/embed_qtconsole.py), but with an IPython kernel that loads a Spock profile. The advantage of using an out-of-process kernel is, that the GUI stays responsive when the kernel executes long running commands. The disadvantage is, that the frontend namespace cannot be easily accessed from the kernel and vice versa.

There already exists some, apparently dead, [code](https://github.com/sardana-org/sardana/blob/76287b416650f40da79871ee3849340d0ff31f1d/src/sardana/spock/ipython_01_00/genutils.py#L1335) for a Spock Qt widget in the `genutils` module. I could get it to work with a few changes but I decided to write a standalone version for increased flexibility. Should that code be deleted?

The first commit is similar to the above mentioned example but extends the `RichJupyterWidget` instead of just creating an instance.

The next commit changes the `genutils` module to set the banner for the `ZMQInteractiveShell` as well. There was already a section in the code for this shell, but is was commented out for some reason. Note that only the newest `genutils` version is changed, as this PR was not tested with old IPython versions.

The following commit refactors the part of the `genutils` module that checks the Spock profile on startup, so that it can be reused.

The fourth commit implements a convenience method for retrieving data from the kernel process. This was temporarily used in the subsequent commit to retrieve the door alias, but it might be useful for other applications as well.

The third last commit provides a keyword to load additional IPython extensions and uses this to overwrite the `edmac` macro with a not implemented error.

The next commit tries harder to ensure that the kernel is only started with the correct Spock profile version. This still leaves a small time window for changing the profile without detection. A more secure approach could be implemented in the Spock profile module itself or in a custom kernel app. 

Finally, I added a few tests. I would need help to integrate the tests into the sardana test suite.

#### Open problems

I could find the following problems and differences from the Spock CLI version:

- [x] `tauruscustomsettings._MAX_DEPRECATIONS_LOGGED` not set to `0` and therefore several deprecation warnings show up in the terminal. This would need to be done on the kernel side before the Spock profile is loaded, e.g., in a custom kernel app. The same applies to the pretty printing of the deprecation counts at the end of the execution.
- [x] The requirements are not checked, but this could be added. 
- [ ] The method `prepare_input_handler`, see [here](https://github.com/sardana-org/sardana/blob/ca8d68677265a2d5a117577760afbf78c00c8398/src/sardana/spock/ipython_01_00/genutils.py#L1370), is not called. It seems that only the input handler created [here](https://github.com/sardana-org/sardana/blob/76287b416650f40da79871ee3849340d0ff31f1d/src/sardana/taurus/core/tango/sardana/macroserver.py#L329) is actually used. Might that be dead code as well?
- [ ] No interface for creating or updating the Spock profile is provided, yet. Instead, the user is simply referred to the CLI version when the Spock profile check fails.
- [x] Some macros, e.g., `edmac`, do not work. More tests are needed to find malfunctioning macros.
- [x] No stty. In `sardana/taurus/core/tango/sardana/macroserver.py:694`, the function
`logReceived` calls the function `_get_console_width` which tries to execute
`os.popen('stty size', 'r').read()`. This prints `stty: 'standard input': Inappropriate ioctl for device` for each log event, i.e., each output line returned from the macro server.
- [ ] There remains a risk for the kernel to load a Spock profile created with a different Spock version.

Possibly fixes https://github.com/taurus-org/taurus/issues/900


reszelaz:
Thanks for this PR @schooft!

I couldn't go too deep into this yet. For the moment I will try to give quick answers to some of your questions.

> I could get it to work with a few changes but I decided to write a standalone version for increased flexibility. Should that code be deleted?

My first impression was also to separate these two things. On the one hand to have the ipython spock profile implementation that could be used by either the CLI ipython or the ipython kernel (in this case together with the Qt widget). And on the other hand the Qt widget (with all the necessary stuff to launch the ipython kernel...). I see in the code that the profile implementation does not need to know anything whether it is used in one or the other way, right?

> Note that only the newest genutils version is changed, as this PR was not tested with old IPython versions.

Don't worry about the previous versions. I think that we at Alba are the only ones who still need them. And we are in the process of migrating to Debian 9 where we will have IPython 5 and higher.


> The following commit refactors the part of the genutils module that checks the Spock profile on startup, so that it can be reused. 

Looks ok!

> The fourth commit implements a convenience method for retrieving data from the kernel process. This was temporarily used in the subsequent commit to retrieve the door alias, but it might be useful for other applications as well.

I would need more time to understand this code:)

> The third last commit provides a keyword to load additional IPython extensions and uses this to overwrite the edmac macro with a not implemented error.

Looks ok to me. I also think that we don't need to support all the CLI spock features from the beginning. So it is fine leave them as TODOs.

> The next commit tries harder to ensure that the kernel is only started with the correct Spock profile version.

Here you mean the spock profile version (at the end this is the Sardana version)? So this is the question of spock profile upgrade offered at the spock startup after the Sardana version update?

> This still leaves a small time window for changing the profile without detection. A more secure approach could be implemented in the Spock profile module itself or in a custom kernel app.
 
I wonder who may change the profile? If I understand correctly the QtSpock should not be used on the same Door where someone runs CLI spock. I mean not used at the same time. In theory the profile can change only on the spock startup, right? Well, one could also go and modify the file manually but this is kind of advanced use and one needs to know that this changes won't apply until you restart the spock application.

> I added a few tests. I would need help to integrate the tests into the sardana test suite.
 
The tests are already run within the sardana test suite. For the moment these fail on travis on Debian 8 and hang on Debian 9. We can check it afterwards.

> - [ ] tauruscustomsettings._MAX_DEPRECATIONS_LOGGED not set to 0 and therefore several deprecation warnings show up in the terminal. This would need to be done on the kernel side before the Spock profile is loaded, e.g., in a custom kernel app. The same applies to the pretty printing of the deprecation counts at the end of the execution.
  
Yes, I confirm that the deprecation logs appear on the terminal but this is not so dramatic as it was in the spock CLI. There the logs appeared in the CLI. In this case this means that the GUI terminal (sometimes GUIs are not even started with the terminal) will be polluted by the deprecation warnings. IMHO we could accept that. I hope that at some time next year.. we could drop support to Taurus 3. After that we will eliminate all the deprecated usage of Taurus 3.

> - [ ] The requirements are not checked, but this could be added.
 
At some point we have talked about removing the requirement check at runtime. The requirements are already checked on the packaging level (pip, debian, etc.). So this may not be mandatory to check it. Let's clarify this at more general level if we want to check at runtime the requirements...

> - [ ] The method prepare_input_handler, see here, is not called. It seems that only the input handler created here is actually used. Might that be dead code as well?
> 

I can not open the first link.

>  - [ ] No stty. In sardana/taurus/core/tango/sardana/macroserver.py:694, the function
logReceived calls the function _get_console_width which tries to execute
os.popen('stty size', 'r').read(). This prints stty: 'standard input': Inappropriate ioctl for device for each log event, i.e., each output line returned from the macro server.

Don't worry about this for the moment. I will talk with @rhomspuron who proposed the use of `stty` in #924 how we can find a sollution which fits both escenarios of CLI spock and Qt spock.

Finally, most probably I won't have time to try more before the Tango Meeting week. We could discuss there the rest of the topics and also carify these points.

Thanks again!

schooft:
Thanks for the feedback!

>   I see in the code that the profile implementation does not need to know anything whether it is used in one or the other way, right?

Mostly, yes. The profile can set configuration options for all supported frontends and a particular frontend will simply ignore the options it doesn't know about. Most magic commands and macros seem to work as well. I found only the `edmac` macro which relies on the special behavior of the CLI. There might be others, though.

> The next commit tries harder to ensure that the kernel is only started with the correct Spock profile version.

In the CLI version, the profile is checked and loaded on startup and then kept in memory until the application terminates. In the Qt version the backend can be restarted independently of the frontend. When the backend restarts, it will read the profile again, which might have been changed in the meantime. It is therefore necessary to update the frontend to reflect these changes, e.g., of the door name or the profile version.

I am not sure what could happen if the loaded profile and Spock version differ. As the version check exists in the Spock CLI code, I tried to make sure it is executed in QtSpock as well.

> I wonder who may change the profile?

Yes, manually or by a run of the CLI version. For example, if sardana was updated to a new version and spock is run on the terminal, it will prompt the user to update the profile. It seems also not to be completely uncommon to delete and recreate the spock profile in case of problems.

> If I understand correctly the QtSpock should not be used on the same Door where someone runs CLI spock.

Nothings prevents this and it should have the same consequences as starting CLI spock multiple times with the same profile. Is this a problem?

> The requirements are already checked on the packaging level (pip, debian, etc.). So this may not be mandatory to check it.

I agree.

> I can not open the first link.

Sorry, I meant [this](https://github.com/sardana-org/sardana/blob/ca8d68677265a2d5a117577760afbf78c00c8398/src/sardana/spock/ipython_01_00/genutils.py#L1370) line.

I am looking forward to discuss the remaining topics on the Tango Meeting.

schooft:
The basic idea of the `get_value` method is as follows:
1. The `execute` method of the Jupyter kernel client allows to evaluate expressions in the user namespace on the kernel side via the `user_expressions` keyword. The expression values are returned as strings formatted using `repr`, see [here](https://jupyter-client.readthedocs.io/en/stable/api/client.html#jupyter_client.KernelClient.execute).
2. Use this mechanism to evaluate `pickle.dumps(value)`, which evaluates to `bytes` or `str`, and read its `repr` from the reply.
3. Use `ast.literal_eval` to safely recover the `bytes` or `str` object which was returned by `pickle.dumps` from its `repr`.
4.  Use `repr.loads` to restore the `value` on the frontend side.

This is mostly a copy of https://github.com/moble/remote_exec/blob/master/remote_exec.py#L61. I did only a few tests and didn't try it with complex objects or across Python versions.

reszelaz:
> In the CLI version, the profile is checked and loaded on startup and then kept in memory until the application terminates. In the Qt version the backend can be restarted independently of the frontend. When the backend restarts, it will read the profile again, which might have been changed in the meantime. It is therefore necessary to update the frontend to reflect these changes, e.g., of the door name or the profile version.

This make sense to me.

Could you please explain me this (from one of the commit messages): "... by the auto-restarter or by calls to QtSpock.restart_kernel." In which occasions could this happen? Does it mean that the kernel may be silently restarted? Does it happen in case of unhandled exceptions? Or this is always done on the user request? Or maybe there are some other cases?

> I am not sure what could happen if the loaded profile and Spock version differ.

What do you mean here by loaded profile and spock version? Do you mean by *loaded profile* the spock_creation_version comment in the profile file? and by spock version the `sardana.release.version`? If yes, you must probably be wondering what could happen if we don't check for the profile upgrade on startup and let it run? 

> Yes, manually or by a run of the CLI version. For example, if sardana was updated to a new version and spock is run on the terminal, it will prompt the user to update the profile. It seems also not to be completely uncommon to delete and recreate the spock profile in case of problems.

I agree, these may happen.

> Nothings prevents this and it should have the same consequences as starting CLI spock multiple times with the same profile. Is this a problem?

Exactly, this is not restricted. However currently running two parallel sessions may lead to confussions. The macro being run, for example a long scan, may be stopped only from the session which started the macro. A user may issue a Ctrl+C in the wrong session and expect it to get stopped while it will never do. In principle, only the macro execution client which started the macro has a control over the execution. If one would like to have a widget which prints the macro output streams, there are some alternatives. Of course we may think on how to improve this situation by printing differently macro outputs on just listening sessions, etc.
My comment was mainly because of that.

>  The method prepare_input_handler, see here, is not called. It seems that only the input handler created here is actually used. Might that be dead code as well?
>  

With the new link this is more clear now, thanks!

You ment that the input handler preparation from the genutils is is not called or maybe that this input handler is not used? I reviewed this code long time ago and would need to debug it again, but the comment: "initialize input handler as soon as possible" recalls me that it has some role.

> I am looking forward to discuss the remaining topics on the Tango Meeting.

Me too!

schooft:
> Could you please explain me this (from one of the commit messages): "... by the auto-restarter or by calls to QtSpock.restart_kernel." In which occasions could this happen? Does it mean that the kernel may be silently restarted? Does it happen in case of unhandled exceptions? Or this is always done on the user request? Or maybe there are some other cases?

The kernel manager periodically checks if the kernel is alive. If that fails, the kernel is restarted automatically.  The check can be disabled by setting `kernel_manager.autorestart` to `False`. Additionally, the kernel client opens a heartbeat connection and emits a `kernel_died` signal if the heartbeat was not received for a certain amount of time. By default, the signal is connected to the `_handle_kernel_died` method, which prints a message to inform the user.

The kernel should not die on unhandled exceptions but, e.g., on segmentation faults (try `pickle.dumps(DeviceProxy("sys/tg_test/1"))`). Another way to trigger this automatic restart is to type `exit` into the console.

Apart from this, there exist several ways to actively request a restart from the frontend side: `restart_kernel("message")` will open a QMessageDialog to ask the user if the kernel should be restarted, `kernel_manager.restart_kernel()` will restart the kernel without user interaction. There might be other cases as well.

A nice way to play around with QtSpock is to execute the following expressions in a freshly started ipython session:

```
In [1]: import sip
   ...: try:
   ...:     sip.setapi('QDate', 2)
   ...:     sip.setapi('QDateTime', 2)
   ...:     sip.setapi('QString', 2)
   ...:     sip.setapi('QTextStream', 2)
   ...:     sip.setapi('QTime', 2)
   ...:     sip.setapi('QUrl', 2)
   ...:     sip.setapi('QVariant', 2)
   ...: except ValueError, e:
   ...:     raise RuntimeError('Could not set API version (%s): did you import PyQt4 directly?' % e)
   ...: 

In [2]: %gui qt4

In [3]: import sardana.taurus.qt.qtgui.extra_sardana.qtspock as qtspock
   ...: sw = qtspock.QtSpockWidget()
   ...: sw.show()
```
Then one can use, e.g, `sw.restart_kernel("message")` or `sw.get_value("value")` in ipython while interacting with the QtSpock window at the same time.
 
> If yes, you must probably be wondering what could happen if we don't check for the profile upgrade on startup and let it run?

I am. Until now, I just assumed the version check is there for a reason and bad things might happen when loading a Spock profile that would make the check fail.

> You ment that the input handler preparation from the genutils is is not called or maybe that this input handler is not used?

It seems that the input handler is not used. At least I could not trigger a breakpoint in its `input` method (and also not in the `on_message` method of the `MessageHandler`, if I recall correctly). Added print statements didn't result in visible output as well.

reszelaz:
> The kernel manager periodically checks if the kernel is alive [...]
 
Thanks for all this explanation. I will try to play around with what you suggested before the Tango Meeting so we could comment it there.

>> If yes, you must probably be wondering what could happen if we don't check for the profile upgrade on startup and let it run?

> I am. Until now, I just assumed the version check is there for a reason and bad things might happen when loading a Spock profile that would make the check fail.

The version check is there for historical reasons. In the last 6 years we have not done any action in the upgrade process apart from updating the spock creation version in the profile. Yes, it is annoying to the users to upgrade the profile everytime the sardana software gets updated, especially knowing that it does nothing... We need to improve that. IMHO we could leave the version check in the code base just in case we need it in the future, but until we plan to really upgrade the profile the check is not necessary to be done.

>> You ment that the input handler preparation from the genutils is is not called or maybe that this inp>ut handler is not used?

> It seems that the input handler is not used. At least I could not trigger a breakpoint in its input method (and also not in the on_message method of the MessageHandler, if I recall correctly). Added print statements didn't result in visible output as well. 

Now I review my last comment about the input handler. It is from 2014... I don´t remember why this input handler is instantiated, but I would not ingore it. Maybe it is somehow related to https://github.com/sardana-org/sardana/issues/143#issuecomment-262974296?

schooft:
> The version check is there for historical reasons.

I can remove the version check. If we then set the macroserver name, the door name and the door alias on the frontend side and ignore their values in the Spock profile, the custom kernel manager could be removed completely.

> Maybe it is somehow related to #143 (comment)?

Indeed the call to prepare_input_handler is for some reason needed in case of `sardanacustomsettings.SPOCK_INPUT_HANDLER == "Qt"`. I didn't know about that option and it does not work with QtSpock, yet. I think, it would make sense to implement this for QtSpock as well. Any idea how to start? Without understanding this, I only see the possibility to reimplement the `_handle_input_request` method of the Jupyter widget.

stanislaw55:
Hello,
first of all, thanks for all the effort put to make this widget work. Really appreciate that.
I have a question related to changing profile at runtime. Let assume that I start spock with some profile (i. e. `spockdoor`) but during the application life I would like to change profile to something else without needing to restart the application. I've tried playing with some kernel shutting down and starting it anew but it did not work properly (some errors related to internal Jupyter Kernel stuff in which fluent I'm not). Is there any way to do this? If there are many ways, which is the best one?

schooft:
To change the profile, one needs to construct the `extra_arguments` oneself and call, e.g., 
`qtspock.kernel_manager.restart_kernel(extra_arguments=["--profile", profile, "--ext", "sardana.taurus.qt.qtgui.extra_sardana.qtspock_ext"])`.

If this is a common use case, one could overwrite the `restart_kernel` methods to accept the same keywords as the constructor. The original settings might then be either replaced or merged with new ones. What would be more useful?

stanislaw55:
Thank you, `restart_kernel` was actually what I needed.

reszelaz:
Since the QtSpock development is in progress maybe it is a good idea to clarify the behavior of different macro execution tools/widgets in terms of stopping/pausing/resuming macros.

I start with a quick overview of the current situation. We have the following tools/widgets for executing macros:
* spock
* macroexecutor
* sequencer
* macrobutton

The current behavior is as following:
1. One can not stop a macro using "Ctrl+C" from a spock instance which has not started the macro.
2. One can not stop a macro with the stop button of the macroexecutor if this macroexecutor has not started it. The same applies for pausing.
3. One can not stop a macro sequence with the stop button of the sequencer if this sequencer has not started it. The same applies for pausing.
4. One can not stop a macro with the macrobutton if this one has not started it. 

Points 2 and 3 may be however confusing due to the "Check door state" context menu action introduced as a workaround and explained in #1148.

Now, since the QtSpock is about to come, we need to clarify how all these tools should behave and what should be possible to do from them and what not. Below my own observations, feel free to comment and propose your ideas.

I think that the point 1 explained above is the correct behavior. Two independent spock processes connected to the same door need to be isolated i.e. "Ctrl+C" issued on one should not stop macro started on the other one.

Also the points 2, 3 and 4 seems to be correct to me. For me all these three widgets configures different macro executions (the sequencer even a macro sequence). For example, if the sequence editor displays a sequence of macros and we enable the stop button on this sequencer, this could make the user think that by pressing on it he would stop this sequence (the one currently displayed). IMHO allowing the user to stop/pause an execution started from another widget may create confusion and this is what #1148 explains. It is even more obvoius in case of the macro button, which role is reserved to execute a fixed macro. 

I think what is on the table with respect to the QtSpock is to connect it with the macroexecutor widget. This is in order to allow some of the users to use the CLI behavior for some easy spock syntax but a GUI executor for complex repeat parameters, pausing macros, etc..
For example, the macro started from the QtSpock could be stopped/paused from the macroexecutor with the stop/pause button. Maybe also that the macro history in the macroexecutor gets updated everytime we run a macro in the QtSpock. This would be nice. Actually this resembles what the "spock yellow line" provides. However the "spock yellow line" is fully connected to the macroexecutor. What do I mean by that? - whenever one type in it this automatically updates the macro combobox and parameters editor of the macroexecutor. Furthermore this connection works in both directions...
Having the same full connection between the QtSpock and the macroexecutor IMO would be very hard do achieve. Furthermore, the code of the "spock yellow line" is something what we ideally would like to get rid of whenever the QtSpock gets available.

One argument against connecting the QtSpock and macroexecutor is that when both are used in the same taurusgui, the macroexecutor panel may be hidden, making it impossible to find the pause button, etc.
As an alternative to this connection could be to add a simple control bar to the Qt spock. Something similar to this widget from PyCharm:
image

In the control panel we could add buttons for pausing/resuming and progress bar. Of course this will not solve the repeat parameters edition issue, but maybe this is enough? An argument for this solution is that it make the QtSpock ehnanced but independent to other widgets. What is your opinion?

teresanunez:
@reszelaz, thanks for your explanations and comments about the different behaviors, I like the idea of adding a bar to QtSpock for stopping/pausing. I think it should be enough.

schooft:
Thank you for the detailed explanation. Except for the confusion introduced by "Check door state", the current behavior makes sense without a QtSpock widget, because the other widgets are linked only by the LED and the inability to execute macros simultaneously. 

The QtSpock widget and Spock in the terminal are different, because they show the output of all macros executed via the same door. This looks especially confusing as the prompt is still available and it is possible to execute normal python code during a macro execution started by another widget (or the python code might already be running, when a macro is started), resulting in mixed up output.

I personally would prefer that QtSpock appears as if it executed all macro itself. Such a behavior can already be achieved across several QtConsole sessions:
1. In one terminal, execute `ipython kernel --profile spockdoor`.
2. In two other terminals, execute `jupyter qtconsole --existing kernel-21047.json --JupyterWidget.include_other_output=True`, with the kernel.json file given in the output of the ipython kernel.
3. In one of the QtConsole applications, execute, e.g, `ascan eh1a_mot01 0.0 10.0 10 1.0`. The other will mirror the appearance (due to a bug, the prompt is not always printed automatically and one has to press enter).

The default behavior of the QtConsole is to not show the output of commands executed in other clients. Try the above steps without `--JupyterWidget.include_other_output=True` to test it. I would also prefer this over the current behavior of Spock and QtSpock.

A related issue is #150.

stanislaw55:
Hi,
I would like to add my view on desired behaviour/appearance of QtSpock widget.

* side bar with stop/pausing buttons is great idea
* printing output of different macros executed from other spock consoles in QtSpock is really confusing and should be avoided (`JupyterWidget.include_other_output` set to False by default)
* QtSpock should *not* be able to stop macros launched by other widgets/consoles. This behaviour would be confusing.

I like the current behaviour of widgets in sardana (being able to stop only macros that were launched by this widget) and I think this is sane solution. Also, AFAICS, there's so called `PanicButton` in MacroGui. It can stop Pool (abort on double click). So general macro stopping can be handled in this way, just with single button.

schooft:
@stanislaw55 You described a coherent view that should be supported. Unfortunately, setting `JupyterWidget.include_other_output=False` is not enough to suppress the output of macros started by other widgets. The output of every macro will always be shown on the most recently active Jupyter client. This probably needs a change on the door side.

In my application, I would like to avoid the macro output widget and only have the QtSpock widget. In that view, the QtSpock CLI would be the main interface for executing macros and the other widgets would be only optional GUI helpers (like the date fields on many web pages where you can either type the date into the field directly or optionally let it be filled by selecting the date in a calendar widget). At least for the macro executor widget this should be relatively easy to achieve by, e.g., replacing the call to `door.RunMacro(string)` by a call to `macroExecutionRequest.emit(string)` and connecting this signal to `qtspock.execute`. A `customMacroExecution` flag on the macro executor could make this configurable.

reszelaz:
Thanks for sharign your ideas! These are very interesting and certainly will help in making the macro execution tools/widgets better!

First some answers to the points you raised:

> I personally would prefer that QtSpock appears as if it executed all macro itself. Such a behavior can already be achieved across several QtConsole sessions:
> 1. [...]
> 2. [...]
> 3. [...]
> 

This works very nice! When one executes a macro from one qtconsole (let's call it *master*) the *line prompt* of the other qtconsole (let's call it *slave*) is always printed at the very bottom and the output keeps appearing above being appended to the previous *line*. IMO this is something what we miss when we execute two simultaneous spock instances connected to the same door. The only detail which I see in this demo is that the line numbers appearing in the slave qtconsole are corresponding to the line numbers comming from the master qtconsole. So, when we switch the roles master-slave,  these may lead to a history of not consecutive line numbers - jumpling up and down. Have you observe that? I wonder if this may not be confusing, for example we do not know what was executed in this qtconsole?

> The output of every macro will always be shown on the most recently active Jupyter client. This probably needs a change on the door side.

I think that this should not be difficult to change. I see at least the following two approaches. Either we add/remove Taurus listener every time we execute a macro (keeping a dummy listener just to not unsubsribe from Tango events). Or we subscribe/unsubscribe to Tango events every time we execute a macro. Either of these two would need to be done for all stream attributes and the result attribute. The first approach is a little bit more CPU intensive when we don't execute macros - we still process events. The second approach has a little performance impact - additional ~20ms per macro execution. Anyway, either of these two will require a change on the client level and is perfectly doable.

Now a little bit of brainstorming about how the spock (terminal CLI) should behave, with the focus on two simultaneous sessions connected to the same Door. Below some ideas...

Let's start from the assumption that spock does not log the macro outputs if the macro was not executed from it.
The spock line prompt should indicate if the Door is running a macro i.e. macro was started from another session, or from a macroexecutor, etc.

For example if you start spock and the macro is already in progress this could look like this (in blue color instead of green).

```
Door_demo1_1 (RUNNING) [1]: 
```

As soon as the said macro finish this could become a normal line prompt (in green).

```
Door_demo1_1 [1]: 
```

And if another macro is started afterwards the prompt should get back to:

```
Door_demo1_1 (RUNNING) [1]: 
```

At any time, it should be possible to enabled the "listening" mode e.g. with a new magic `listen on|off`. In this mode the spock will log the macro outputs as if it was the one starting the macro. The difference in this case is that the line prompt would be kept at the very bottom and the macro outputs would be constantly appended to the previous line (similar to the example of qtconsole from the very beginning of this comment). I think that it is interesting to change the prompt to blue RUNNING also in this case:

```
Door_demo1_1 (RUNNING) [1]: listen on
   [...]
   7         7         1      8.52278 
   8         8         1      9.71236 
   [...]
Door_demo1_1 (RUNNING) [2]: 
```

First of all, I don't know if it is technically possible to change the line prompt without having to press return by the user? Also I don't know if it would be possible to append to the previous line prompt? Anyone has experience with these? If this is not possible, let's forget about this...

Also, kind of related to this is the issue #150. IMHO the Ctrl+C keybard combination should be reserved for stopping macros started from the same spock session. But what we could add is another magic (executed on the client) e.g. `stop` and `abort` that would call the door's `StopMacro` and `AbortMacro`. Maybe it should require user confirmation.

But let's for a while come back to the QtSpock. I understand the requirements pointed in https://github.com/sardana-org/sardana/pull/1109#issuecomment-503476571. Let me summarize the requirements in a list:

* log any macro outputs executed on this door? Something like the listen mode explained above?
* allow graphical introducion of the parameter values

Or this also includes?:
* possibility to pause/resume 
* macro execution progress bar
* storage of favourites
* storage of history

And sorry for so long message again... (if someone reached this line:P)

schooft:
> So, when we switch the roles master-slave, these may lead to a history of not consecutive line numbers - jumpling up and down.

The kernel should maintain a monotonically increasing execution number and past executions should be displayed with the correct number. However, the prompt is only updated when you press enter, so the prompt at the bottom might show a too low number.

> I wonder if this may not be confusing, for example we do not know what was executed in this qtconsole?

It should never matter in which qtconsole something was "executed", i.e., sent to the kernel, as long as they are all connected to the same kernel.

By the way, I just found that qtconsole version 4.4 introduced the `other_output_prefix` option, which by default prints `[remote]` before each line coming from other clients. This might help to reduce potential confusion.

 > First of all, I don't know if it is technically possible to change the line prompt without having to press return by the user? 

The Jupyter console (not qtconsole) has the option `--ZMQTerminalInteractiveShell.include_other_output`. Its description states "Outputs are not displayed until enter is pressed". This indicates an automatic update is not trivial. Unfortunately, with this option enabled, the session crashes when pressing enter for my version 5.0.0.

For the qtconsole, it could be possible to connect the frontend widget to the door as well for receiving status change events directly. Then the frontend could update the prompt accordingly.

> Let me summarize the requirements

For my own application I plan the following behavior (not necessarily implementation): 
* QtSpock is the only interface to a door, i.e., everything should be possible using QtSpock only. This includes
    * Start/Pause/Resume/Stop all macros
    * Show the output of all macros
    * Scroll through the full execution history
* I like the idea of the vertical sidebar of buttons for Start/Pause/Resume/Stop being part of the QtSpock widget in addition to Enter, Crtl + C, and potential other shortcuts
* Other macro widgets are optional helpers that the user can show or hide according to his preferences. For the user, the widget's action should be indistinguishable from formatting a spock command, copying it to QtSpock, and pressing enter. These widgets then provide, among others:
    * A favorite list
    * A history list
    * A progress bar

I don't know how complicated this would be for the macro sequencer. For all other widgets, this should be relatively easy?

schooft:
In the two newest commits, the door name from the Spock profile is ignored by default. Instead, it has to be set via `setModel` after contruction of the QtSpock widget. This behavior can be changed by setting the `use_model_from_profile` flag. In both cases, the kernel is not started in the constructor any more, but either by setting a valid model or by calling `start_kernel`.

I am not sure if I use the Taurus features correctly. For example, I couldn't get the door name to be loaded from the Taurus configuration. Could somebody review the `__init__` function of the QtSpock widget and the QtSpock main window?

Open questions:
* [ ] Correct Taurus usage?
* [ ] Calling `setModel` when the kernel is already running restarts the kernel automatically. This will delete the whole state of the Spock/IPython user session. Should there be a warning in/after the configuration dialog?
* [ ] As now the Spock related profile settings are ignored, it might make sense to create a default `qtspockdoor` profile automatically if no profile was found, instead of asking the user to start Spock in the terminal.

schooft:
> Either we add/remove Taurus listener every time we execute a macro (keeping a dummy listener just to not unsubsribe from Tango events). Or we subscribe/unsubscribe to Tango events every time we execute a macro..

I found the method `SpockBaseDoor.setConsoleReady(bool)`. When setting it to `False`, the output is suppressed without any overhead (it just sets a flag which is read before writing to the output stream). Would that be a faster alternative to the above proposal?

Partially related: Spock in a Jupyter Notebook

1. Create the file `.local/share/jupyter/kernels/spock/kernel.json` with the following content:
```
{
 "argv": [
  "/usr/bin/python", 
  "-m", 
  "ipykernel", 
  "-f", 
  "{connection_file}", 
  "--profile", 
  "spockdoor"
 ], 
 "display_name": "Spock", 
 "language": "python"
}
```
2. `jupyter notebook`
3. Choose New -> Spock

Here, the output of a macro started elsewhere is printed to the most recently active cell, which might even change during the execution.

stanislaw55:
Hi,
I am no Taurus expert but I see one improvement in `setModel`. When received model is the same as the current one, nothing should happen (no restarting kernel). 

About creating `qtspockdoor` profile for QtSpock it is good idea IMO. However, I think there would be little problem with creating profile programatically. See functions `create_spock_profile` and `_create_config_file` in `sardana.spock.ipython_01_00.genutils`. I think `door_name` argument is ignored and QtSpock would still ask about Door name in command line probably.

reszelaz:
Hopefully I could join the review tomorrow afternoon. Thanks @schooft and @stanislaw55!

schooft:
@stanislaw55 Thank you, I updated `setModel` accordingly. 

The last two commits make it possible to use the macro executor as a helper widget for QtSpock. It should be fully backwards compatible.

Unfortunately, i couldn't find a simple solution to show the complete macro history in the history widget. I think, one approach would be to implement a `runningMacro` attribute in the door tango server, but for the moment I will just live with the history widget as is.

schooft:
What should I do with the failing test? The failure of the standard test seems to be unrelated (already `docker exec sardana-test supervisorctl start Pool` failed) but the failure of the taurus-support-3.x test is due to the docker container missing the xvfb-run executable. Should the QtSpock tests be excluded from the taurus-support-3.x tests?

reszelaz:
Finally I could come back to QtSpock...

I try to catch up with the discussion before the code review and tests:

> By the way, I just found that qtconsole version 4.4 introduced the other_output_prefix option, which by default prints [remote] before each line coming from other clients. This might help to reduce potential confusion.
 
Sounds good. However I could not test it cause I use the 4.2.1 version (Debian 9).

>> First of all, I don't know if it is technically possible to change the line prompt without having to press return by the user?

> The Jupyter console (not qtconsole) has the option --ZMQTerminalInteractiveShell.include_other_output. Its description states "Outputs are not displayed until enter is pressed". This indicates an automatic update is not trivial. Unfortunately, with this option enabled, the session crashes when pressing enter for my version 5.0.0.

If I understand correctly this assumes a change in the architecture for two parallel spock sessions (terminals) connected to the same door. Currently these are two "single process" ipythons and this idea assumes running a single jupyter kernel and two console frontend processes connected to it, right?

If yes, IMHO we could change the architecture if this opens the possibility to implement the behavior described in https://github.com/sardana-org/sardana/pull/1109#issuecomment-503632823

I also use 5.0.0 and I couldn't make it work.

> For the qtconsole, it could be possible to connect the frontend widget to the door as well for receiving status change events directly. Then the frontend could update the prompt accordingly.

Yes, in this case it may be possible with some "home made" solution.

I'll continue tomorrow...

schooft:
> If I understand correctly this assumes a change in the architecture for two parallel spock sessions (terminals) connected to the same door. Currently these are two "single process" ipythons and this idea assumes running a single jupyter kernel and two console frontend processes connected to it, right?

In Jupyter, a client gets notified about every line/cell executed by the kernel for other clients and every output of all executions. This makes it possible to mirror every input/output of all clients. 

For changing the prompt to `door (Running)` and/or including the output of macros, it might be easier to just listen to the corresponding Tango attributes of the door. Implementing automatic updates for the terminal might be difficult, given that the jupyter console in the terminal does not support automatic updates yet but only after pressing enter.

Currently, I am playing with pausing/resuming macros in QtSpock and I wonder why a macro returns when it is paused. As I can't execute other macros anyway when the door is in the STANDBY state, it might be better to just block until the macro finished/stopped. The cause is that `QDoor.runMacro` returns on any state change. What is the idea behind this behavior?

reszelaz:
> Currently, I am playing with pausing/resuming macros in QtSpock and I wonder why a macro returns when it is paused. As I can't execute other macros anyway when the door is in the STANDBY state, it might be better to just block until the macro finished/stopped. The cause is that QDoor.runMacro returns on any state change. What is the idea behind this behavior?

I agree with you, when a macro is executed with `synch=True` this should wait until it finished or get stopped. Since spock was the only one executing in this way (macroexecutor and macrobutton are running with `synch=False` and it was not foreseen for the users to pause macros from spock, we have not realized about this. IMO we could change this behavior. Thanks for pointing this out!

reszelaz:
> In the two newest commits, the door name from the Spock profile is ignored by default. Instead, it has to be set via setModel after contruction of the QtSpock widget. This behavior can be changed by setting the use_model_from_profile flag. In both cases, the kernel is not started in the constructor any more, but either by setting a valid model or by calling start_kernel.
> 

LGTM

> I am not sure if I use the Taurus features correctly. For example, I couldn't get the door name to be loaded from the Taurus configuration. Could somebody review the __init__ function of the QtSpock widget and the QtSpock main window?
> 

I fixed this in d851a3f1f after prior merge of develop branch. I hope you are fine that I pushed it directly to your branch.

> Open questions:
> 
> * [ ]  Correct Taurus usage?
> 

I also corrected some minor things. But in general I see it ok.

> * [ ]  Calling setModel when the kernel is already running restarts the kernel automatically. This will delete the whole state of the Spock/IPython user session. Should there be a warning in/after the configuration dialog?

We can always add it in the future if someone complain. 

> * [ ]  As now the Spock related profile settings are ignored, it might make sense to create a default qtspockdoor profile automatically if no profile was found, instead of asking the user to start Spock in the terminal.
> 

I agree with you! It certainly make sense but I would also leave it for the future.

> I found the method SpockBaseDoor.setConsoleReady(bool). When setting it to False, the output is suppressed without any overhead (it just sets a flag which is read before writing to the output stream). Would that be a faster alternative to the above proposal?
> 

In terms of the performace I would say it is worse - we are always processing the event and just printing or not based on the flag. But as the first approach IMHO it is enough. And this solution already exists:)

> Partially related: Spock in a Jupyter Notebook
> 
Great! I tried it and it worked directly. 

> About creating qtspockdoor profile for QtSpock it is good idea IMO. However, I think there would be little problem with creating profile programatically. See functions create_spock_profile and _create_config_file in sardana.spock.ipython_01_00.genutils. I think door_name argument is ignored and QtSpock would still ask about Door name in command line probably.
> 

Yes, to make it work the `door_name` would need to be propagated from `create_spock_profile` to `_create_config_file`. Then it should be pretty reusable. But now I wonder if the DOOR_NAME should go to the qtspockprofile? Now I realized that the taurus model and spock profile settings compete for the same role (who stores the persistance?). Also the selection of the Taurus model (Door from the GUI) has one down side - if we select a door which belongs to a different macroserver, the macro history (stored in the profile) may loose sense, for example if the macro or macro argument does not exist in the newly selected macroserver. This is not something new - the macroexecutor widget actually has the same problem and we lived like this for years. So I would not worry about this now...

> The last two commits make it possible to use the macro executor as a helper widget for QtSpock. It should be fully backwards compatible.
> 

How can I test that?

Also IMO the QtSpock should get instantiated within the TaurusGUI framework when one chooses to connect to the Sardana system:

https://github.com/sardana-org/sardana/blob/4b24003ebbb7ecedaaefb2b857b81330bfcd755c/src/sardana/taurus/qt/qtgui/macrolistener/macrolistener.py#L478

But as with the previous TODOs, this does not need to be done now.

> The failure of the standard test seems to be unrelated (already docker exec sardana-test supervisorctl start Pool failed)

Exactly, there are problems from time to time with the TravisCI infrastructure. I just restarted this build and now it is green.

> but the failure of the taurus-support-3.x test is due to the docker container missing the xvfb-run executable. Should the QtSpock tests be excluded from the taurus-support-3.x tests?
> 
I tried to make it work, I managed to reach the point when the only thing which is missing is the qtconsole. In Debian 8 this would be: https://packages.debian.org/jessie/ipython-qtconsole.
Should the test work with this version? If not we could install it with pip... But anyway I would not dedicate too much time to this cause soon we will drop support of the develop branch to Debian 8 (it commes only with Python 3.4).

IMHO it is very convenient to integrate this PR before we start migrating Sardana to Python 3 (in a week or so we will start executing 2to3 od a dedicated branch). I see it pretty usable now. We could integrated it with a warning of an experimental API and work on the pending TODOs when we are already in Python 3. What is your opinion?

stanislaw55:
One more comment on indicating that in prompt that Door is running other macros:
Maybe instead of changing prompt, there should be Door status led in the quick actions bar mentioned earlier by @reszelaz (actions like stop, start, pause, resume etc.). It would be easier to implement (at least I think it should) and make clear that Door is running some macro now. What do you think?

stanislaw55:
One more thing: I think that `modelInCofnig` should *not* be set to `True` for QtSpock widget in constructor. It could be set to `True` in standalone widget. This way, it is more flexible in case of configuration.

reszelaz:
> One more comment on indicating that in prompt that Door is running other macros:
> Maybe instead of changing prompt, there should be Door status led in the quick actions bar mentioned earlier by @reszelaz (actions like stop, start, pause, resume etc.). It would be easier to implement (at least I think it should) and make clear that Door is running some macro now. What do you think?

Yes, for the QtSpock this is a very good idea. For the normal spock we still need something based on the prompt I think.

> One more thing: I think that modelInCofnig should not be set to True for QtSpock widget in constructor. It could be set to True for standalone window. This way, it is more flexible in case of configuration.

Also fine with me. As I wrote above there is still the point of where to store the persistance of the model: profile vs. taurus settings. It is also somehow related to the way the new profile will be created by the qtspock. We could think of creating a profile with a name based on the Taurus application name, and even if we anticipate a possibility of having two QtSpock widgets in the same application connected to two different doors then we may append a widget name to the profile name.

Now a more organizational question, @schooft will it be fine to you to use this feature in Sardana directly in Python 3, or you will need to also for Python 2?

stanislaw55:
About storing door name: for me it is absolutely fine to store door name in Taurus settings since this is widget used standalone or in bigger application. I would avoid creating profiles per app with door name appended since it could cause in specific scenario to flood IPython directory with profiles. However, I'm not sure if one profile can be reused for different QtSpocks without side effects.

reszelaz:
Thanks for the feedback @stanislaw55!

> I would avoid creating profiles per app with door name appended since it could cause in specific scenario to flood IPython directory with profiles.

I didn't mean to use the door name in the profile name, but to use `QApplication.applicationName` and eventually append `QObject.objectName` in case multiple QtSpocks connected to different doors are used in the same application. But I would say that the object name would be rerelly needed. Also I would say that having to QtSpock in the same application connected to the same door does not make sense. 

> However, I'm not sure if one profile can be reused for different QtSpocks without side effects.

I think that the ipython history is per profile, so sharing the same profile would collide here.

In anyway, I think this discussion should not block the integration of this PR. Now the QtSpock is not able to create new profile and it is fine as the first appoach.

stanislaw55:
> I think that the ipython history is per profile, so sharing the same profile would collide here.

If that's true, then different profiles are necessary indeed. 

Sorry for misunderstanding. Actually, proposed way of naming profiles looks good.

reszelaz:
> If that's true, then different profiles are necessary indeed.

I just cross-checked that and this is true. So at least the history is one thing that is in favour of not sharing the `qtspockprofile` between different QtSpock instances connected to different doors.

schooft:
Sorry, I am on vacation and only on my phone.

> I hope you are fine that I pushed it directly to your branch.

Of course, thank you for the improvements.

> How can I test [the macro executor changes]?

The idea is to enable applications to connect the customMacroRunRequested signal of the MacroExecutorWidget to the runMacro slot of QtSpock. Further changes I started implementing included a new onDoorStatusUpdated slot for the macro executor that keeps its buttons synchronized with the door state. 

Probably the commit that changes the macro executor should be moved to a separate PR. Can you delete the commit? I can open a new PR in about 2 weeks when this one is merged and my changes are more complete.

> We could integrated it with a warning of an experimental API and work on the pending TODOs when we are already in Python 3. What is your opinion?

I am fine with that. I am happy to only use Python 3 as soon as Sardana supports that.

> We could think of creating a profile with a name based on the Taurus application name

I would suggest to create a default profile with the name given in the constructor if the profile does not already exist and only if use_model_from_profile is False. In that case it doesn't really matter if the door name is stored there as it is ignored anyway.

Currently the profile name can only be set in the constructor. Should that be a Taurus setting instead?

> So at least the history is one thing that is in favour of not sharing the qtspockprofile between different QtSpock instances connected to different doors.

All my bash sessions, even via ssh, share the same history and I never found that confusing. But I do not have a strong opinion on that.

It just came to my mind, that the history (all commands entered, not just macros) might leak sensitive information about the previous user. This might be something to be aware of.

One remaining problem that might be a bit buried in the discussion is that setting sardanacustomsettings.SPOCK_INPUT_HANDLER to "Qt" does not work, yet. 

All the other check boxes seem not to be blocking.

stanislaw55:
Hi,
does anyone had problems with `QtKernelClient`? For some reason it fails with `AttributeError: QtKernelClient has no attribute 'iopub_channel'`. I'm asking hear to find out if that might be problem with version update or something else.

schooft:
@stanislaw55 Thank you for finding this error. I added a fix.

I also removed the commit with the changes to the macroexecutor and will create a new PR after this is merged.

stanislaw55:
Fix you introduced works just fine. Thanks @schooft!

schooft:
I rebased the PR on the current develop branch, dropping Python 2 compatibility.

If only Python versions >= 3.3 are targeted now, the `test_requires` argument in the setup.py can be removed.

For some reason, I got segmentation faults when running the tests in the updated docker container and just using `xvfb-run`. Using `xvfb-run -s '-screen 0 1920x1080x24'` seems to fix this problem.

stanislaw55:
I have a question regarding `umv` macro. In general, all update-macros. Does it get updated inline in QtSpock like it does in console spock?

schooft:
It does not and unfortunately there seems to be no easy way to implement this as rewriting multiple lines is not supported by the Jupyter qtconsole, see https://github.com/jupyter/qtconsole/issues/263.

cpascual:
>If only Python versions >= 3.3 are targeted now, the test_requires argument in the setup.py can be removed.

Sardana now tagets python >=3.5 , so you may remove it.

TNoga0:
Hello,

I was wondering if there is an easy way to increase the lines buffer value in the widget. I need a few thousand (maybe even more) lines to be displayed. By default they are truncated after a few hundred and I haven't found any "easy" Jupyter options to modify.

schooft:
Unfortunately, I cannot reproduce the problem. I found two settings that sound relevant, `ConsoleWidget.buffer_size`, which is set to 500 by default, and `ZMQInteractiveShell.cache_size`, which is set to 1000 by default. However, something like
```
for i in range(10000):
    print(i)
```
or
```
s = ""
for i in range(10000):
    s += str(i) + "\n"
print(s)
```
prints all the lines for me, independent of these settings. 

Do the two settings above work for you? If not, could you give a small example that shows the behavior you described?

TNoga0:
Setting `ConsoleWidget.buffer_size` in any class that inherits from ConsoleWidget does the trick, `ZMQInteractiveShell.cache_size` didn't quite work. 
Thank you for your help!

TNoga0:
Hello again,

I have encountered a problem with the widget regarding the command prompts.
Sometimes after a finished scan or macro, the command prompt does not appear, and pressing the Enter key does nothing in terms of calling it. Some cases are solvable by left-clicking right after the last printed character and pressing the Enter key, but the majority of situations require restarting the TaurusGUI which contains QtSpock widget.

Is there a simple way to force the command prompt to appear?

Unfortunately I cannot provide you with a screenshot as the issue is quite unpredictable.

reszelaz (approved):
Hi all,

It has been a very long time since @schooft submitted this PR. Even if the subject indicates WIP and there were some pending TODOs, and planty of ideas identified during the discussion, on the last follow-up meetings we decided to integrate the QtSpock in Sardana as an experimental feature.

Before the final integration, I just:
* Added user's and API documentation
* Added note about experimental feature
* Removed dead code from genutils

Also, I checked some TODOs:
* taurus deprectation warnings should not be a problem anymore - Sardana does not support Taurus 3 anymore and a lot of effort was put on removing the deprecated usages
* it is not intended by Sardana applications to check requirements at runtime
* stty call was removed from Sardana - see #1245 and #1268
* QtSpock was successfully tested with other macros (looks like only edmac and block output macros are not working as expected)

From the initial TODOs the following are still pending:
* Only CLI input handler is working now - Qt input handler would be desirable
* No interface for creating or updating the Spock profile is provided, yet. Instead, the user is simply referred to the CLI version when the Spock profile check fails.
* There remains a risk for the kernel to load a Spock profile created with a different Spock version.

As I mentioned before this PR lead to very interesting discussions, I reviewed it again and wrote down the following ideas that we could address in the future:
* QtSpock could get instantiated within the TaurusGUI framework when one chooses to connect to the Sardana system
* We could leave the Spock version check in the code base just in case we need it in the future, but until we plan to really upgrade the profile the check is not necessary to be done and could be removed
* Discussion about the optimal behavior of different macro execution tools/widgets in terms of stopping/pausing/resuming macros, printing outuput, etc.
    * QtSpock as if it executed all macros itself i.e. allow to stop/pause macros which were not started in QtSpock
    * macroexecutor could be helper to QtSpock for complex parameters edition i.e. edit parameters in macroexecutor and execute it in QtSpock and print the output in QtSpock (DoorOutput widgets would not be necessary)
    * QtSpock similar to other widgets e.g. macroexecutor, macrobutton, sequencer. So these have control only on macros which were executed on these widget, furthermore output of macros not executed from these clients could be surpressed.
    * Enhance the QtSpock and Spock prompt with more information (RUNNING), blue color, etc. and surpress output when macro was started in another client. Output from macros started in other clients could be enabled/disabled with a magic command. This output should be always pre-pended maintaining the prompt at the very bottom.
    * Add control panel to QtSpock (pause, resume, stop buttons, status led, etc)

Here I would like to thank to all people participating in this ehancement: @schooft, @stanislaw55, @teresanunez, @TNoga0 and @cpascual! Hopefull it will be usefull to you and we will continue improving this widget!

PR #1110, opened 2019-05-20
Fix nb_starts calculation for software synchronization.

rhomspuron:


reszelaz:
Great! Thanks for spotting this! Merging..

PR #775, opened 2018-06-01
SEP2 - Improve integration of 1D and 2D experimental channels

reszelaz:
Even if there are no changes in this PR yet I already start it so we have a place for discussions.

The current [SEP2](https://github.com/reszelaz/sardana/blob/sep2/doc/source/sep/SEP2.md) document requires modifications.
This branch will also serve for future implementation.

PR #1111, opened 2019-05-22
Compose default DataSource value pointing to Value Tango attribute

reszelaz:
This was discovered when testing SEP2 in: https://github.com/sardana-org/sardana/pull/775#issuecomment-494749686.

Default behavior of 2D controllers, if not overridden, is to return
None as data source. However it is not possible to push Tango events with
DevString type with None as value. Compose the default data source as it
is done in read_DataSource.


teresanunez:
I have just tested it and it solves the problem. Many thanks.

PR #1113, opened 2019-05-24
Add auto-deploy to PyPI with Travis

cpascual:
Make Travis deploy to PyPI when a version tag (major.minor.patch) is pushed. It uses an ad-hoc PyPI account (sardana_bot) which is given Maintainer permisions in PyPI. The sardana_bot password is encrypted (using [travis-encrypt](https://pypi.org/project/travis-encrypt/)) with the sardana-org/sardana Travis public key so that Travis can decrypt it for deployment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,3 +73,14 @@ doctr:
   require-master: false
   sync: true
   built-docs: build/sphinx/html/
+
+deploy:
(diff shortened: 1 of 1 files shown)


cpascual:
Note: the sardana_bot PyPI account is not yet added as pypi maintainer because its associated email needs to be verified. @reszelaz , can you please take care of verifying?

PR #1120, opened 2019-05-28
Conditional output of send2ctrl macro

dschick:
the send2ctrl macro generates always an `self.output()` even if there is no response from the controller.
This results in unwanted blank lines in spock, e.g. if multiple calls of the `send2ctrl` macro are done in a user hook like the preScanHook.

This PR simply adds a condition to the `send2ctrl` macro.

Generally one could avoid printing blank lines in the `self.output() ` method itself, but I guess, here it is intentionally.


dschick:
yeah, somehow I missed the gitignore. Thanks for merging

PR #1121, opened 2019-05-29
fix for FIO recorder

jkotan:
* removes circular dependencies of macro object related to FIO recorder 
* applies the new naming scheme (with tango://) for MCAs in FIO recordder


jkotan:
I fixes one of the sources of memory leak.

teresanunez:
HI @reszelaz, thanks, actually the FIO format is from Thorsten. I have not done anything there but it
also looks fine to me, so i IMO it can be merged.

PR #1136, opened 2019-06-06
Add spock format documentation for sequencer

daneos:


dschick:
Do you think one should add also the usage of ipython macros/files for sequences? Here you can use also for loops and lists, etc. In fact it is much more powerful then pure sequences. I find it pretty handy and much closer to the SPEC is doing it. So maybe there is a place where to document it.

PR #1146, opened 2019-06-13
WIP: Don't use Tango for snapshots

daneos:
This is one of the fixmes in the code.
The pre scan snapshot is now done using Taurus Attribute instead of Tango AttributeProxy.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -803,9 +803,13 @@ class GScan(Logger):
                                         label=label,
                                         source=src)
 
-                # @Fixme: Tango-centric. It should work for any Taurus
-                # Attribute
-                v = PyTango.AttributeProxy(column.source).read().value
+                try:
+                    # Taurus 4
+                    v = taurus.Attribute(column.source).read().rvalue.magnitude
+                except AttributeError:
+                    # Taurus 3
+                    v = taurus.Attribute(column.source).read().rvalue
+
                 column.pre_scan_value = v
                 column.shape = np.shape(v)
                 column.dtype = getattr(v, 'dtype', np.dtype(type(v))).name


reszelaz:
Looks good!
Do you also have a feeling of the performance of using Taurus instead of PyTango? If I'm not mistaken Taurus will subscribe to the configuration events.
It would be good to have numbers to compare, for example with a snapshot of 50-100 attributes (I think that in some cases the scientists for example enter all the motors of the beamline to the snapshot group).

daneos:
This is exactly the case on one of our beamlines, all motors are in the snapshot group.
I'll try to do some timings for Taurus vs. PyTango.

pjb671:
We have order 100 motors in the snapshot at Nanomax. I tried this solution yesterday, but it seemed to cause additional strange behaviour.  I have taurus 4.4 and sardana 2.5.  The time taken to read the snapshot is visibly longer, several s of visible waiting time that I did not see before:

Door_NanomaxExpert [51]: ascan dummy_motor_1_expert 1 2 1 0.1
Error taking pre-scan snapshot of sy (tango://b-v-nanomax-csdb-0.maxiv.lu.se:10000/motor/lc400_ctrl/2)   <<< expected since motor missing
Error taking pre-scan snapshot of sz (tango://b-v-nanomax-csdb-0.maxiv.lu.se:10000/motor/lc400_ctrl/3)  <<< expected since motor missing
Error taking pre-scan snapshot of sx (tango://b-v-nanomax-csdb-0.maxiv.lu.se:10000/motor/lc400_ctrl/1)  <<< expected since motor missing
<<<<  about 4 s here before next line printed  >>>> 
kafka_recorder: send data to scicat  <<<< hooked macro
Operation will be saved in /data/staff/nanomax/commissioning_2019-1/20190619/raw/test_prescan_snapshot/test_prescan_snapshot.h5 (HDF5::NXscan)
Scan #83 started at Wed Jun 19 10:55:25 2019. It will take at least 0:00:00.765685
#Pt No dummy_motor_1_expert uxtimer dt

But after running about 20 scans, I started to see the following lines printed in spock exactly after the above (ie after the prescan had finished and the scan started):

TaurusTP.W001 ERROR 2019-06-19 10:55:39,424 b-v-nanomax-csdb-0.maxiv.lu.se:10000.MS_NanomaxExpert: Exception occurred processing elements
TaurusTP.W001 ERROR 2019-06-19 10:55:39,426 b-v-nanomax-csdb-0.maxiv.lu.se:10000.MS_NanomaxExpert: Details:
File "/usr/lib/python2.7/site-packages/sardana/taurus/core/tango/sardana/macroserver.py", line 893, in on_elements_changed
return self._on_elements_changed(evt_src, evt_type, evt_value)
File "/usr/lib/python2.7/site-packages/sardana/taurus/core/tango/sardana/macroserver.py", line 912, in _on_elements_changed
element = self._addElement(element_data)
File "/usr/lib/python2.7/site-packages/sardana/spock/spockms.py", line 573, in _addElement
device_proxy = element.getObj().getHWObj()
File "/usr/lib/python2.7/site-packages/taurus/core/util/log.py", line 1055, in new_func
return func(*args, **kwargs)
File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangodevice.py", line 210, in getHWObj
return self.getDeviceProxy()
File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangodevice.py", line 213, in getDeviceProxy
if self._deviceObj is None:
File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangodevice.py", line 74, in __getattr__
if self._deviceObj is not None:
File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangodevice.py", line 74, in __getattr__

.... lines repeated many times.

I could not see how this behaviour could be related to the change from AttributeProxy to taurus Attribute, but after restarting Pool and MS it would always reoccur after a few scans.   Can it be related to the delay in taking the snapshot?  When I reverted to the original AttributeProxy (or DeviceProxy) this issue went away. I had not seen this behaviour before.

daneos:
Hm, the `if self._deviceObj is None:` error is usually related to loss of communication between Pool and MacroServer. I however don't see how this change would trigger that behaviour, and I'm not observing this problem.
I noticed the delay in snapshots when changing Sardana version from 2.3.2 to 2.7.0, could it be related to the version update, and not this PR?

reszelaz:
I think this PR is no more needed. Since #1190 we don't need this workaround.

Issue #1027, opened 2019-01-29
Bug in showing default values in sequencer parameter editor


cmft:
The sequencer parameter editor is not updated with the default value of a macro.
It works in Sardana (2.5.0).

mrosanes:
Commit 26dc568 introduces this bug (default values on sequencer).

--

A possible workaround could be to set as before the defValue in init method of SingleParamNode (sardana/taurus/core/tango/sardana/macro.py):
```
self.setValue(self.defValue())
```

But side effects could appear. 
If possible, it should be the macroserver which passes the default values to the client (sequencer).

mrosanes:
This bug also exists in macroexecutor (tested in sardana devel branch version 2.7.1-alpha)

reszelaz:
The part of the bug affecting the macroexecutor was fixed in #1153. The sequencer is still buggy.

PR #1153, opened 2019-06-21
Fix 1027

cmft:
Fix issue #1027 
Fix bug with command editor in macroexecutor (yellow line)
Improve the behaviour with default values (Now the param tree shows the default value instead none).

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/macro.py
+++ b/src/sardana/taurus/core/tango/sardana/macro.py
@@ -570,7 +570,7 @@ class SingleParamNode(ParamNode):
         Empty list indicates default value."""
         if isinstance(v, list):
             if len(v) == 0:
-                v = self.defValue()
+                v = str(self.defValue())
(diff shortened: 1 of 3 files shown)


cmft:
I run locally this test and works

FAIL: test_stop_meas_cont_acquisition_3 (sardana.tango.pool.test.test_measurementgroup.TangoAcquisitionTestCase)
Stop of the synchronized acquisition with two channels from the same controller using hardware trigger

reszelaz:
@cmft won't fix this problem now. I will proceed to merge this PR as it is and I will reopen the #1027 .

Also I will add a TODO in the [`SingleParamNode`](https://github.com/sardana-org/sardana/blob/1f05860d241a42cc4bd549140c91fe316fe678e9/src/sardana/taurus/core/tango/sardana/macro.py#L485) class. I think it should not store values as `str`.

Many thanks!

PR #1160, opened 2019-06-26
Fix spock's output block on windows

reszelaz:
On Windows `stty` is not available what fallbacks to infinite number of characters per line. In this case the number of lines is fixed and always 1.

This was discovered during Jul19 release tests on Windows.
This bug was introduced by #924. I think it is serious enough to be fixed during the release process.

I confirm it fixes the issue on Windows.
@sardana-org/integrators before integrating please crosscheck that on Linux it sill works.


teresanunez:
I confirm that it does not introduce any problem in Linux (at least that I could detect)

PR #1185, opened 2019-08-21
Remove the sardana.requirements module and usage

cpascual:
The sardana.requirements module provides check_requirements() which is
(only) used in the `__init__.py` of sardana.

The current implementation and usage of check_requirements() presents
some disadvantages:

- checking requirements at import time is too strict and causes trouble
in some scenarios where one may want to legitimately relax or ignore the
requirements (e.g. when preparing a deb package while some deps are not
yet available). The check is better done on installation.

- the implementation of check_requirements() overlaps the more standard
functionality provided by the pkg_resources module.

- some requirement specifications are duplicated (in setup.py and in
sardana.requirements) and must be kept in sync manually

Fix all this by completely removing the module and its usage, and
relying on the requirement checks done at installation time according to
the specifications from the setup.py


cpascual:
Note that I do not propose a deprecation before removing because this PR targets the just-created v3.0.0-alpha

cpascual:
If any @sardana-org/integrators can review this soon, it would help for creating the sardana deb package...

cpascual:
Thanks for the review @teresanunez,

I merge it.

PR #1189, opened 2019-09-03
[WIP] Fix auto upload to pypi

### LINKED ISSUE #1166: Automatic release to PyPI does not work
The automatic release to PyPI introduced in #1113 does not work correclty.
It only uploads the egg file and not the source distribution.

According to the [documentation](https://docs.travis-ci.com/user/deployment/pypi/#uploading-different-distributions) it should by default also upload the source distribution.

Not sure if the source distribution does not need to be created in the script part, like for example in [PAScual](https://github.com/cpascual/PAScual/blob/master/.travis.yml) project.

In the investigation we could run a test with a PR which modifies the `.travis.yml` (remove the condition) and upload to PyPI to see if it works.

reszelaz:
Based on the https://github.com/sardana-org/sardana/issues/1166#issuecomment-522724917 I have tried to do the same in Sardana (make a sdist build before the deploy). However I did not manage to make it work.

Furthermore, I'm not even able to upload the egg file which used to work - see [2.8.1 upload](https://pypi.org/project/sardana/2.8.1/#files) done by sardana_bot.

The error which we get seems to have to do with the authentication:

```
Deploying application
running sdist
running egg_info
/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/setuptools/dist.py:474: UserWarning: Normalizing '3.0.0-alpha' to '3.0.0a0'
  normalized_version,
error: [Errno 13] Permission denied
Uploading distributions to https://upload.pypi.org/legacy/
Uploading sardana-3.0.0a0-py3.5.egg
100%|██████████| 2.13M/2.13M [00:00<00:00, 4.87MB/s]
NOTE: Try --verbose to see response content.
HTTPError: 403 Client Error: Invalid or non-existent authentication information. for url: https://upload.pypi.org/legacy/

build/lib/sardana/__init__.py already exists, no checkout
build/lib/sardana/macroserver/__init__.py already exists, no checkout
[...]
src/sardana.egg-info/top_level.txt already exists, no checkout
src/sardana/__pycache__/release.cpython-35.pyc already exists, no checkout
Could not restore untracked files from stash entry
PyPI upload failed.
```

I have it leave it for now - lack of time. If someone is willing to try more, go ahead. Otherwise we will deal with that during the next release process.

IMPORTANT: Remember to revert the a34bfe0 before integration.

Fixes #1166.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ services:
   - docker
 
 python:
-  - "3.5"
+  - "3.6"
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,10 +73,15 @@ doctr:
   sync: true
   built-docs: build/sphinx/html/
 
+before_deploy:
+  # build source dist, otherwise deploy to PyPI does not work
+  - if [ $TEST == "testsuite" ]; then docker exec -t sardana-test /bin/bash -c "cd /sardana ; python3 setup.py sdist"; fi
+
 deploy:
   # deploy to pypi when a version tag is pushed to the official repo
   - provider: pypi
     user: sardana_bot
+    skip_existing: true
     password:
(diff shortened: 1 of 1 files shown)


reszelaz:
Not needed anymore. We are uploading usign GitHub workflows.

Issue #1080, opened 2019-03-19
prescan snapshot sometimes hangs


pjb671:
Hello,
Apologies for opening more issues when you still need feedback on the first. At Nanomax at Max IV we have noticed scans sometimes failing in the prescan snapshot of gscan. The beamline recently added many more attributes to the snapshot (over 100 in total) and the failures started then. I can see very clearly that the scan simply hangs at line 765 here https://github.com/sardana-org/sardana/blob/release-Jul18/src/sardana/macroserver/scan/gscan.py (v = PyTango.AttributeProxy(column.source).read().value)
The strange thing is that it is not always the same attribute that causes the stop, and I do not believe there is any problem with the attribute when it happens. Any anyway, this is within try/catch so if the attribute could not be read there should be an error. I have a debug statement before and after the read line, so I see which attribute it is trying to read, but the read simply never returns.  And as I say, it can be any random attribute. I would say it happens maybe 1 in 20 times, perhaps once a week.
cheers,
Paul

reszelaz:
To be honest this is a first time I hear about this kind of behavior.

For sure you must have checked this already but just in case... could you please somehow stress test the tango reads (from the same host where the MacroServer runs), something like the following snippet run in a loop:

```python
import tango
import shelve

path = "/home/zreszela/workspace/MacroServer/zreszela/macroserver.properties"
env = shelve.open(path)
snapshot = env["PreScanSnapshot"]
for item, label in snapshot:
    dev, attr = None, None
    try:
        dev = tango.DeviceProxy(item)
    except:
        attr = tango.AttributeProxy(item)
    if dev:
        if "Position" in dev.get_attribute_list():
            value = dev.read_attribute("Position").value
        elif "Value" in dev.get_attribute_list():
            value = dev.read_attribute("Value").value
    elif attr:
        value = attr.read().value
    print("{0}: {1}".format(item, value))
```
(please remember to change path to your environment location, if you don't change the `EnvironmentDb` property of the MacroServer this should be: `/tmp/tango/MacroServer/<inst_name>/macroserver.properties`).

dschick:
I encountered the same problem. After around 200-400 scans Sardana got stuck during the snapshot.

I am currently running the snippet of @reszelaz 
Before that I tried a` %wum` motor for every motor in the snapshot group 1000 times without an issue.
I post the result here as soon as possible.

mirlin:
[...]

PR #1190, opened 2019-09-05
Do not detect events on MeasurementGroup's LatencyTime and NbStarts

reszelaz:
Due to a bug in PyTango (or Tango) we need to be very carefull with not properly configured event system. Subscription to an attribute in stateless mode which constantly fails may collide with destruction of Attribute/DeviceProxy.

More details in [tango-controls/pytango#302](https://github.com/tango-controls/pytango/issues/302).

To be on the safe side don't check event criteria for LatencyTime neither NbStarts. While in the case of the
first one we would like to always receive the events, even if the change is extreamly low, the second one has sense only to send events if the difference is 1. But for the moment we don't even check the NbStarts criteria.

Fixes #1118 . At least I run it for several hours now and it works, I will continue overnight...
@dschick could you try it on your system (the changes should be easily portable to your code)
It could be even the reason for #1080: @amilan, @pjb671 could you verify that?

I suspect (not sure) that this bug was present since 2.7 (Jan19) for the NbStarts attribute which was affecting step scans and continuous scans. And since 2.6 (Jul18) affecting only continuous scans.

Also, I suspect that until the bug in PyTango (or Tango) gets fixed removing of the event system criteria for the following attributes:

- {CTExpChannel}/Value
- {IORegister}/Value
- {Motor}/Position
- {Motor}/DialPosition
- {PseudoCounter}/Value
- {PseudoMotor}/Position
- {ZeroDExpChannel}/Value

 may lead to similar hangs.

### DIFF
--- a/src/sardana/tango/pool/MeasurementGroup.py
+++ b/src/sardana/tango/pool/MeasurementGroup.py
@@ -73,10 +73,14 @@ class MeasurementGroup(PoolGroupDevice):
     def init_device(self):
         PoolGroupDevice.init_device(self)
         # state and status are already set by the super class
-        detect_evts = "latencytime", "moveable", "synchronization", \
-                      "softwaresynchronizerinitialdomain", "nbstarts"
+        detect_evts = "moveable", "synchronization", \
+                      "softwaresynchronizerinitialdomain"
+        # TODO: nbstarts could be moved to detect events with
+        # abs_change criteria of 1, but be careful with
+        # tango-controls/pytango#302
         non_detect_evts = "configuration", "integrationtime", "monitorcount", \
-                          "acquisitionmode", "elementlist"
+                          "acquisitionmode", "elementlist", "latencytime", \
+                          "nbstarts"
         self.set_change_events(detect_evts, non_detect_evts)
 
         self.Elements = list(self.Elements)


reszelaz:
With this fix, the long scan that previously never finished correctly, now finished without problems.

```
Running print_sth...
 99999     99999     0.001     0.002     0.003     0.004     0.005     0.006     0.007     0.008    8344.51
Running print_sth...
 100000    100000    0.001     0.002     0.003     0.004     0.005     0.006     0.007     0.008    8344.59 
Scan #16 ended at Thu Sep  5 21:15:28 2019, taking 2:19:04.640308. Dead time 98.8% (motion dead time 42.2%)
```

@sardana-org/integrators can you take a look on this PR and eventually integrate it?

dschick:
hej @reszelaz ,

do I need to do anything else than adapting the` MeasurementGroup.py` file?

reszelaz:
> do I need to do anything else than adapting the MeasurementGroup.py file?

No, that's it! Just move the "latencytime", "nbstarts" to the to the non_detect_evts.

dschick:
I have the patch installed and I am testing it now.

dschick:
I ran sardana overnight doing nearly 1000 scans each having 100 points without any errors.
It seems that the bug fix is working as expected.

Thanks @reszelaz

reszelaz:
Thanks for the tests @dschick !

IMHO this bug is serious enough to make the hotfix release 2.8.3.
@dschick since you were also experiencing #1080, could you check if this patch fixes it as well? I have a strong feeling that this was the same issue - there we also use disposable `AttributeProxy` objects.

reszelaz:
I've manually merged the hotfix-2.8.2 branch, so this PR is no more necessary.
I have also bumped the version to 3.0.1-alpha in the develop brach to mark that that this crical bug was integrated. I will also close the #1080 as @dschick confirms that this problem did not appear in his tests.

PR #1193, opened 2019-09-12
Hotfix 2.8.2

teresanunez:
Apply fix #1190


reszelaz:
Thanks @teresanunez for preparing the hotfix!

The travis-ci fails with `sardana-test:latest` docker image. This is because it was already updated to use python 3. But all the changes are ok, so I'm merging it. I just updated the CHANGELOG.md to reflect this release.

I wil also merge the same branch into develop.

izadworny:
Hi @reszelaz 
I've noticed that latencytime stays in both detect_evts and non_detect_evts.
Is it correct?

reszelaz:
> I've noticed that latencytime stays in both detect_evts and non_detect_evts.
> Is it correct?

Of course not, thanks for spotting this out! We will prepare the 2.8.3.

PR #1199, opened 2019-09-19
Add snapshot configuration macros

reszelaz:
For the moment two macros were developed: `lssnap` and `addsnap` macros. Of course the macro names, or anything else, can be changed. These are just suggestions...

Also the lssnap macro could be improved to print output in the table format. 

I would also add a macro to remove an alement from the snapshot.

I conservatively marked it as WIP. If someone wants to continue this development - feel free (I won't have time in the next month, or so). If you consider it can already go to develop branch, I think this is also ok since it adds value and is usable.

Thanks in advance for your feedback!

### DIFF
--- a/doc/source/users/standard_macro_catalog.rst
+++ b/doc/source/users/standard_macro_catalog.rst
@@ -117,6 +117,7 @@ list related macros
     * :class:`~sardana.macroserver.macros.lists.lsmac`
     * :class:`~sardana.macroserver.macros.lists.lsmaclib`
     * :class:`~sardana.macroserver.macros.env.lsgh`
+    * :class:`~sardana.macroserver.macros.env.lssnap`
 
 measurement configuration macros
 --------------------------------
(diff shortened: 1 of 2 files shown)


teresanunez:
Hi Zibi, I have tested it and it works fine, thanks for the implementation. I also think that it would
 be good to have a macro for deleting elements from the snapshot, but IMO this macros could
 be already go to develop.

reszelaz:
Thanks @teresanunez for the feedback!

I've found some time to advance with it. As we said on the last follow-up I've renamed the macros to follow other macro names e.g. `defgh`, `udefgh`, etc.

I've also added the `udefsnap` macro. With this I think the feature is now complete.

As a demo I post here some example of usages:

```
Door_demo1_2 [27]: lssnap
No pre-scan snapshot

Door_demo1_2 [28]: defsnap mot39

Door_demo1_2 [29]: lssnap
   Snap item                              Snap item full name
 ----------- ------------------------------------------------
       mot39   tango://pc255.cells.es:10000/motor/motctrl10/3

Door_demo1_2 [30]: defsnap sys/tg_test/1/ampli

Door_demo1_2 [31]: lssnap
   Snap item                                Snap item full name
 ----------- --------------------------------------------------
       mot39     tango://pc255.cells.es:10000/motor/motctrl10/3
       ampli   tango://pc255.cells.es:10000/sys/tg_test/1/ampli

Door_demo1_2 [32]: udefsnap mot39

Door_demo1_2 [33]: lssnap
   Snap item                                Snap item full name
 ----------- --------------------------------------------------
       ampli   tango://pc255.cells.es:10000/sys/tg_test/1/ampli

Door_demo1_2 [34]: udefsnap

Door_demo1_2 [35]: lssnap
No pre-scan snapshot
```
I proceed with the merge. Thanks!

PR #1201, opened 2019-09-20
Fix WrongParam message showed in spock

reszelaz:
WrongParam message, for example due to object not found, are showed in spock as unknown sardana exception instead of showing a verbose message. This is due to the fact that spock shows just the high level exception message. Since Py3 we raise exception from lower level exceptions (see 77a7aa91) then
we must also pass the exception message to the high level exception. Do it.

To reproduce one can try to run a `mv` macro on a motor which does not exist.

Before this fix:
```
Door_py3_1 [1]: mv mot09 10
Error: Unknown sardana exception
```
After this fix:
```
Door_py3_1 [1]: mv mot09 10
Error: Moveable with name mot09 does not exist
```

@sardana-org/integrators can you look at this please? Many thanks!

### DIFF
--- a/src/sardana/macroserver/msparameter.py
+++ b/src/sardana/macroserver/msparameter.py
@@ -436,9 +436,9 @@ class ParamDecoder:
                     param = param_type.getObj(str(value))
 
             except ValueError as e:
-                raise WrongParamType from e
+                raise WrongParamType(str(e)) from e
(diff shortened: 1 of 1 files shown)


teresanunez:
IMO it can be merged. Thanks Zibi

reszelaz:
Merging then. Thanks for the review!

Issue #674, opened 2018-01-04
Sardana server (standalone) has problems with Pool Elements subscription


reszelaz:
Seems like there is a limitation in Tango that it is not possible to subscribe to the attribute events before the complete DS was started. The following [query](https://github.com/tango-controls/TangoTickets/issues/7) was raised to the Tango experts.

The following WARNING message is logged at the server startup:

```
MainThread     DEBUG    2018-01-04 11:15:18,909 TaurusPollingTimer[3000].Timer on _pollAttributes: Timer::stop()
MainThread     WARNING  2018-01-04 11:15:18,909 pc255.cells.es:10000.Pool_zreszela_2: Received elements error event Pool was shutdown or is inaccessible
MainThread     DEBUG    2018-01-04 11:15:18,909 pc255.cells.es:10000.Pool_zreszela_2: DevFailed[
DevError[
    desc = Device dserver/sardana/zreszela is not exported (hint: try starting the device server)
  origin = DeviceProxy::get_corba_name()
  reason = API_DeviceNotExported
severity = ERR]

DevError[
    desc = Failed to execute command_inout on device dserver/sardana/zreszela, command ZmqEventSubscriptionChange
  origin = Connection::command_inout()
  reason = API_CommandFailed
severity = ERR]

DevError[
    desc = Device server send exception while trying to register event
  origin = EventConsumer::connect_event()
  reason = API_DSFailedRegisteringEvent
severity = ERR]
]
```

As a consequence of the failed event subscription the Taurus polling gets enabled on this attribute which will be disabled at some point when the reconnection succeeds. 

The consequences of this behavior were not studied yet. But IMHO, since the Elements attribute is very critical to the system integrity we should look for a more reliable way of the Pool and MacroServer connection on the startup. So most probably we will need to change the way the Sardana server starts up.

The following segfaults https://github.com/sardana-org/sardana/issues/466 are also relaed to this subscription failure. Also on the old open SUSE 11.1 and ALBA's Tango stack (still Tango 7) the segfault is reproducible. I don't know to which extend they depend on the underneeth Tango stack  - on my Debian PC the segfault is not reproducible.

reszelaz:
[...]

PR #1203, opened 2019-09-23
Improve channel TV widget 

reszelaz:
Pool channel TV widget does not offer too many possibilities. Currently it shows the channel's value attribute and a button to access the rest of the channel's interface

This PR allows to execute single acquisition over the experimental channel. Since single acquisition for the moment works only for the timerable channels, the 0D and pseudo counters still use the old widget.

Acquisition is controlled with the extra widget which now contains line edit for the integration time and a button for starting and stopping the acquisition.

The new functionalities are demonstrated on the below screenshot:

image

After integrating this PR we will need to open a PR to taurus to modify the `tauruscustomsettings.T_FORM_CUSTOM_WIDGET_MAP` to the following:

```
T_FORM_CUSTOM_WIDGET_MAP = \
    {'SimuMotor': ('sardana.taurus.qt.qtgui.extra_pool.PoolMotorTV', (), {}),
     'Motor': ('sardana.taurus.qt.qtgui.extra_pool.PoolMotorTV', (), {}),
     'PseudoMotor': ('sardana.taurus.qt.qtgui.extra_pool.PoolMotorTV', (), {}),
     'PseudoCounter': ('sardana.taurus.qt.qtgui.extra_pool._PoolChannelTV', (), {}),
     'CTExpChannel': ('sardana.taurus.qt.qtgui.extra_pool.PoolChannelTV', (), {}),
     'ZeroDExpChannel': ('sardana.taurus.qt.qtgui.extra_pool._PoolChannelTV', (), {}),
     'OneDExpChannel': ('sardana.taurus.qt.qtgui.extra_pool.PoolChannelTV', (), {}),
     'TwoDExpChannel': ('sardana.taurus.qt.qtgui.extra_pool.PoolChannelTV', (), {}),
     'IORegister': ('sardana.taurus.qt.qtgui.extra_pool.PoolIORegisterTV', (), {})
     }
```
Thanks @cpascual and @guifrecuni for the suggestions bout UX.

@sardana-org/integrators please, could you take a look on it. Thanks!


teresanunez:
Hi Zibi, I have looked at this, but I am afraid, my current Taurus version is not compatible
with this, I get errors related to taurus.

reszelaz:
Hi Teresa, 
Thanks for looking into this. I just checked this with the latest taurus from develop and it works. At least for `taurus form ct01`. But since this branch is very outdated I just tried it after updating it with the latest develop and then it does not work anymore. There are no errors reported but the acquisition does not start. I push the updated branch and change the subject to WIP - this will need to be fixed. Then if we don't arrive for the release I propose to exlude it from the milestone.

reszelaz:
Hi agian,
I think I wrongly tested it the last time.
After the server startup the integration time of a timerable channel is 0. Then the acquisition is immediate and it made me think it the widget was not starting the acquisition at all. So to notice that the acquisition takes place one needs to set a different integration time. This has to be changed in the edit field and then enter has to be pressed. Afterwards one can see that the widget is working correctly.
Remember that for 1D and 2D dummy channels you need to set timer attribute. Their controllers do not have default_timer set. 
I'm removing the WIP. It is ready for the review.

teresanunez:
Hi Zibi, thanks for checking. The problem in my case is not in the use or performance, the problem is that the widget ist not open at all, I am afraid it is not compatible with the Taurus version I have installed.

reszelaz:
Hi Teresa,
You can use virtualenv or conda to prepare an environment for tests. For example, to use virtualenv:
```console
sudo apt-get install python3-venv
python3 -m venv pr1203  # create virtualenv
source pr1203/bin/activate  # activate virtualenv
pip3 install pytango
pip3 install PyQt5==5.14  # install PyQt5 - it must be this version cause the newer versions has some problems
pip3 install git+https://github.com/taurus-org/taurus.git  # install taurus from develop
pip3 install git+https://github.com/reszelaz/sardana.git@channel_widget  # install Sardana from this PR branch
pr1203/bin/Sardana pr1203  # start Sardana server
```

Then in another console activate virtualenv and start a spock session to populate sardana with the `sar_demo` macro.

And finally open taurus form e.g. `taurus form ct01 zerod01 oned01 twod01`. 

Please let us know if this worked. Thanks a lot!

teresanunez:
Hi Zibi,
              many thanks for your help. But it doesn't work for me. I have just tried it and
 I do not get Sardana running.

                         Regards,
                                      Teresa

teresanunez:
I have managed now to get Sardana and Spock running (Sardana was actually running, but
 some errors were appearing, Spock had a problem with the profile). I can open the new gui,
 but for the ZeroD it shows not number, for the counter it works and for the oned and twod it
throws an error when starting the acquisition: it does not find the Timer, it says always that it
 is not found. So the problem is not the widget, but the single acquisition itself.

reszelaz:
Great Teresa! I see that you advanced a lot alrady. I will try to clarify some issues:

> Sardana was actually running, but some errors were appearing

You most probably are hitting https://github.com/sardana-org/sardana/issues/674. It is a know issue. It is not very harmful until you don't use the server in the very first 10 s after the startup.

> Spock had a problem with the profile

Most probably this is due the lack of matplotlib in your virtual environment. If you install it: `pip3 install matplotlib` then there should be no problems at the Spock startup, at least they disappear for me.

> for the ZeroD it shows not number

This is normal. The ZeroD value attribute readout raises an exception until you run one acquisition. It was designed like this. If you create a measuerement group with this zerod channel and run a ct then the widget will show the result of the last acqusition.

But anyway it is not intended to use this widget for the zerod neither for the pseudo counters - these still do not support experimental channel acquisition. As I said in https://github.com/sardana-org/sardana/pull/1203#issue-320420714 we need to maintain the backwards compatibility and and display the old widget for them. Since this PR review is advancing I think it is the right momento to create a PR to taurus with this change.

> for the oned and twod it
throws an error when starting the acquisition: it does not find the Timer, it says always that it
is not found. So the problem is not the widget, but the single acquisition itself

I think there is no problem with the experimental channel acquisition (single acquisition). You just need to configure the timer attribute for oned and twod. You can do it from the "Tango attribute" form. Just right click on the channel's label and from the context menu select "Tango attribute". There, set the timer attribute to "__self" - then each of the channel will act as a timer at the same time.
I just realized that you will need taurus_pyqtgraph to fully test it for oned and twod. So in the virtualenv install it: `pip3 install taurus_pyqtgraph`.

Just let me know if you have more problems. Thanks a lot!

teresanunez:
Thanks, Zibi, it works now.

cpascual:
please have a look at my comment in the related PR in Taurus: https://github.com/taurus-org/taurus/pull/1101#pullrequestreview-408768248

reszelaz:
I have just updated it with the latest develop. @cpascual what do you think, is it ready to merge? Thanks!

reszelaz:
Thanks Teresa and Carlos for the review. I'm merging..

PR #1206, opened 2019-09-27
Avoid taurus deprecation warnings

cmft:
Adapt sardana code to avoid taurus deprecation warnings
starting with the GUIs


reszelaz:
@cmft thanks for this PR! It reduces significantly the taurus 3/4 deprecation warnings.
I've tested this branch on the icalepcs2019 training materials and there are no side effects.
Merging...

PR #1208, opened 2019-10-02
Don't use theme icon for "Add to favourites" in macroexecutor

reszelaz:
software-update-available is not accessible when using taurus-org/taurus#1012
via theme (on Debian Buster with KDE). Use directly the icon.
This works correctly on Debian Stretch and Debian Buster (both with KDE).


reszelaz:
I'm auto-merging since this is a trivial change and taurus-org/taurus#1012 is already integrated.

PR #1213, opened 2019-10-07
WIP: first draft of a suitcase based recorder

tacaswell:
This needs a bunch more work, but as a proof-of-concept it works.

Verified that the output files from this can be read back in via databroker.

### DIFF
--- a/src/sardana/macroserver/recorders/suitcase.py
+++ b/src/sardana/macroserver/recorders/suitcase.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+
+#
+#
+# This file is part of Sardana
+#
+# http://www.sardana-controls.org/
+#
+# Copyright 2019 Brookhaven National Laboratory
+#
+# Sardana is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Sardana is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


tacaswell:
Is there a way to inject additional meta-data at run time (sample information, semantic information about _why_ the user did this ascan, etc)?  If so, where does in show up?

reszelaz:
> Is there a way to inject additional meta-data at run time (sample information, semantic information about _why_ the user did this ascan, etc)? If so, where does in show up?

There is a concept of _custom data_ in Sardana recorders. Unfortunately it is not yet easily configurable from the UI. You would need to do it programmatically. We plan to add this option from the UI at some point.Take a look on this example:
https://github.com/sardana-org/sardana/blob/be8d64d34234b35a37b4cd8233e86fcda13f64b8/src/sardana/macroserver/macros/examples/scans.py#L594

From the recorder plugin side you just need to implement the `_addCustomData` method. You can take a look on this example:
https://github.com/sardana-org/sardana/blob/be8d64d34234b35a37b4cd8233e86fcda13f64b8/src/sardana/macroserver/recorders/h5storage.py#L561

Related to the recorders, there is also a concept of _pre-scan snapshot_. So at the beginning of the scan a set of configurable elements (either proceeding from Sardana or external) are read once and the readout values are then passed to the recorders. This snapshot elements are easily configurable from the GUI or CLI.

tacaswell:
It looks like the custom data can come in at _any_ point between the `_startRecrodList` and `_endRecordList`?

We collect all of the extra data (user name, plan name, requested number of points, sample name, blah blah) into the start document (which in this example I am generating in `_startRecordList`).


----
to write the data, should be able to use the versions on pypi:

https://github.com/bluesky/event-model/
https://github.com/bluesky/suitcase-jsonl


To read it back (not by just looking at the jsonl):

https://github.com/bluesky/databroker <- install the master branch

```
import databroker
cat = databroker._drivers.jsonl.BlueskyJSONLCatalog('/tmp/*.jsonl')
h = cat[-1]
h.primary().read()  # an xarray of the data we just took
```

PR #1214, opened 2019-10-10
Fix Tango inheritance

tiagocoutinho:
Make sure sardana tango servers always use the latest tango API

### DIFF
--- a/src/sardana/tango/core/SardanaDevice.py
+++ b/src/sardana/tango/core/SardanaDevice.py
@@ -35,7 +35,7 @@ import time
 import threading
 
 import PyTango.constants
-from PyTango import Device_4Impl, DeviceClass, Util, DevState, \
+from PyTango import LatestDeviceImpl, DeviceClass, Util, DevState, \
     AttrQuality, TimeVal, ArgType, ApiUtil, DevFailed, WAttribute
 
 from taurus.core.util.threadpool import ThreadPool
@@ -66,7 +66,7 @@ def get_thread_pool():
         return __thread_pool
 
 
-class SardanaDevice(Device_4Impl, Logger):
+class SardanaDevice(LatestDeviceImpl, Logger):
     """SardanaDevice represents the base class for all Sardana
     :class:`PyTango.DeviceImpl` classes"""
 
@@ -74,7 +74,7 @@ class SardanaDevice(Device_4Impl, Logger):
         """Constructor"""
         self.in_constructor = True
         try:
-            Device_4Impl.__init__(self, dclass, name)
+            LatestDeviceImpl.__init__(self, dclass, name)
             self.init(name)
             Logger.__init__(self, name)
 
@@ -512,7 +512,7 @@ class SardanaDeviceClass(DeviceClass):
         return dict(ProjectTitle="Sardana", Description="Generic description",
                     doc_url="http://sardana-controls.org/",
                     __icon=self.get_name().lower() + ".png",
-                    InheritedFrom=["Device_4Impl"])
+                    InheritedFrom=["Device_5Impl"])
 
     def write_class_property(self):
         """Write class properties ``ProjectTitle``, ``Description``,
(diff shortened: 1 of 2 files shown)


tiagocoutinho:
@reszelaz, I think this will fix the blocking issues we had when we were trying to test sardana with tango 9.3.x. Do you remember if there is an issue reported in sardana? I couldn't find it.

reszelaz:
Great! I completely agree. I had even started migrating Sardana to use HLAPI since during ICALEPCS I tried the PyTango `TestDeviceContext` and it requires the HLAPI but this work is not yet ready for making a PR. I will continue when I'm back. You can merge it yourself whenever the documentation build start to work again on Travis. I tried it twice now and constantly gives error on git-stash. Please check it tomorrow morning and if it is green merge.

src/sardana/tango/core/SardanaDevice.py
```diff
@@ -512,7 +512,7 @@ def _get_class_properties(self):
         return dict(ProjectTitle="Sardana", Description="Generic description",
                     doc_url="http://sardana-controls.org/",
                     __icon=self.get_name().lower() + ".png",
-                    InheritedFrom=["Device_4Impl"])
+                    InheritedFrom=["Device_5Impl"])
```
reszelaz: Maybe this to not hard-code it could be `LatestDeviceImpl.__name__`? or sth like this? Anyway if its difficult to do don't worry. I think that in the HLAPI we could avoid using `DeviceClass`, right?

tiagocoutinho: It would be easy if `LatestDeviceImpl` was an alias for `Device_5Impl`. The PyTango implementation is actually a subclass of `Device_5Impl` (I cannot remember why) so to find a name we are looking for, we would have to fetch the super class and get its name.

Taking into account that:
* this information is not actually used by anyone 
* we will probably have a HLAPI sardana implementation much sooner than a Device_6Impl. (and anyway a PyTango 10 will probably not expose the low level API anyway ;-)

I don't think it is worth the effort

reszelaz: I agree, I commented just in case it was easy to do.

reszelaz:
> You can merge it yourself whenever the documentation build start to work again on Travis. I tried it twice now and constantly gives error on git-stash. Please check it tomorrow morning and if it is green merge.

Seems like there are some problems with the travis-ci documentation job. I rebuild it several times (also today) and it still fails with the same [error](https://travis-ci.org/sardana-org/sardana/jobs/596033859).
However if I rebuild the latest develop branch, it [works](https://travis-ci.org/sardana-org/sardana/jobs/593672056).

rhomspuron:
If I use this branch without to merge to develop the expconf can not create a MG:
error
```
MainThread     WARNING  2019-10-29 15:50:21,213 TaurusRootLogger: Problem with taurus.qt.editor (hint: is spyder >=3 installed?)
  File "/homelocal/rhoms/develop/pycharm/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py", line 1069, in addChannel
    qmodel.addChannel(chinfo=ch_info)
  File "/homelocal/rhoms/develop/pycharm/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py", line 632, in addChannel
    ctrlsdict = self.dataSource()['controllers']
KeyError: 'controllers'

```

tiagocoutinho:
Hi, #1268 seems to fail Travis with the same error.
It looks like it has nothing to do with the PRs.
We should have a close look at this. Seems to be blocking a few PRs

reszelaz:
Thanks for the reminder! 
Hmm, strange. The develop branch does not have travis-ci problems. Actually I was more worried about https://github.com/sardana-org/sardana/pull/1214#issuecomment-547470549 than on travis-ci.
But anyway I see that there are a lot of commits in the PR, also the changes are unrelated to the subject and there are conflicting files.

tiagocoutinho:
Thanks for bringing this to my attention.
The branch should be synchronized now. Probably I did an unintended push.
At least the checks pass now

reszelaz:
Tiago, I just reverted this merge. Could you please update the branch with the latest develop and then propose it again (I don't know if you can re-open it or will need to propose a new one).

Issue #1172, opened 2019-07-18
MacroButton does not work with repeat parameters


stanislaw55:
Hi,
I've been trying to used sardana's `MacroButton` widget but with no success. 
This is the snippet which I used
```python
import sys

from taurus.qt.qtgui.application import TaurusApplication

app = TaurusApplication(sys.argv)
w = MacroButton()
w.setModel(sys.argv[1])
w.setMacroName("mv")
w.updateMacroArgument(0, "mot01")
w.updateMacroArgument(1, 100)
w.show()
sys.exit(app.exec_())
```

However, Door fails with error "Error: Moveable with name m does not exist"
Taurus version: 4.5.0
Sardana version: 2.7.1

It looks like this is error with Door itself. For some reason it splits string arguments and that causes error. However, I think this can come from wrong argument passing to `MacroButton`. Hence my question: how to properly use `MacroButton`

reszelaz:
It may be a bug related to repeat parameters. `mv` macro uses repeat parameter.

Have you tried the main section of the macrobutton.py? There should be a widget with macro button and some text field for imputing parameters? Does it work with `mv`?

stanislaw55:
Hi,
Another thing is that it works for macros without arguments, like `lsmac` of something similar.

> Have you tried the main section of the macrobutton.py?

Yes, I've tried that but there's only field for macro name and no place to provide macro arguments. See attached screenshot
![image](https://user-images.githubusercontent.com/32959446/61517933-b387ea00-aa08-11e9-880c-d27f134e4234.png)

When I tired to pass full macro with arguments (`mv mot01 100`) to macro name edit, it behaved just like I described earlier, resulting with error in Door.

So @reszelaz this may be the issue you pointed since running macro with limited number of arguments(`relmac ascan`) works fine. However, this seems like error in Door processing arguments, not MacroButton itself. Am I right?

reszelaz:
`lsmac` has one repeat parameter called filter, and this is the text field that you see:
https://github.com/sardana-org/sardana/blob/9ea3b041491e776cd674e1b62700ede76037b536/src/sardana/macroserver/macros/lists.py#L49

I'm not sure but I think the limitation of the main section of the macrobutton.py module is that it only allows to pass one repetition of the repeat parameter. But this should not be the limitation of the `MacroButton` itself.

stanislaw55:
So setting macro arguments by `updateMacroArgument` is correct approach?

cpascual:
[...]

PR #1216, opened 2019-10-21
Use uuid for MacroNode.id() in macroexecutor and sequencer

### LINKED ISSUE #1215: Some issues with macroIDs
Hello,

I have recently encountered an issue with MacroButton objects - executing macros from sequences in plain text results in macroID being an UUID, not an int, which pops up a warning:
```
  File "/usr/lib/python2.7/site-packages/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py", line 747, in onMacroStatusUpdated
    id = int(id)
ValueError: invalid literal for int() with base 10: 'ff476ebe-f0d7-11e9-b72c-0242ac130006'
```
I would like to parse these sequences to XML, which produce int-type IDs, but I can't seem to find a method which would basically take a plain text sequence and convert it into XML in any of the base classes.

Is there a convenient method which would help me?
Let me know if the explanation of the problem is too short or chaotic, I will elaborate.

daneos:
This quick fix removes the cast of macro id to int, as now the id can also be an UUID.
Closes #1215 and [this part](https://github.com/sardana-org/sardana/issues/1172#issuecomment-514997000) of #1172.


reszelaz:
Thanks for the PR @daneos! What I see now is that the progress bar in macroexecutor stopped updating when we execute a macro from there.

What about using uuid in `MacroNode.assignId`? I think then we won't need casting to `int` neither..

reszelaz:
Hi @daneos, @cpt-majkel,
I have implemented the idea mentioned in https://github.com/sardana-org/sardana/pull/1216#issuecomment-545535940 in 1f733ce471267a52c706a35228631f9cd1450c4f and it works correctly as far as I tested (spock, macrobutton, macroexecutor, sequencer). I will proceed with the merge. It would be great if you could verify if it works for your cases as well before the Jan21 release.
Cheers!

reszelaz (approved):

PR #1220, opened 2019-10-24
Fix issue 1218

amilan:
Fixing the FScan macro which was trying to use the sscan as a
deterministic macro without being ready/supported.


reszelaz:
Thanks @amilan for the fix!

It was broken in SEP18.
I also added tests for `fscan` to avoid eventual future regressions.
I took a profit of extending the documentation about the _deterministic scans_ which were added in SEP18 and when accepting it we sacrified the documentation part :(

The documentation job on travis fails similarly as in https://github.com/sardana-org/sardana/pull/1214#issuecomment-542241861. I will merge anyway, cause it may be some problem just with PR. The last develop build worked well....

PR #1222, opened 2019-11-05
Use travis deploy instead of doctr to deploy documentation

reszelaz:
Travis documentation jobs started to fail recently. This was first observed in https://github.com/sardana-org/sardana/pull/1214#issuecomment-540885289 and from then on affects all the other builds.
Apparently nothing had changed: Docker image, doctr had not changed around this dates. The investigation on this issue is quite difficult. The job builds either report: `git stash --all failed: 1` or simply hangs.

As an alternative solution I have tried to directly use travis deploy to GitHub Pages - therewith this PR.

There are two important issues to be considered now:
1. Other, different versions of the documentation won't be automatically build. But IMHO this is not really a problem, actually I consider it as an improvement. We have never used them in the last 1 year and a half and now we need to clean them up - see this [page](https://sardana-controls.org/other_versions.html). So, basically instead of manually removing documentation versions now will need to manually add them. To add a new version, a new project will need to be added in sardana-controls organization and travis.yml would need to be updated with the new deploy.
2. Now doctr uses SSH key to push to sardana-doc project. The travis deploy needs a personal token. If we add a personal token, for examply my personal token, it will have rights to push to any of the repository that I have access. To overcome this issue a bot user could be created with right to push to sardana-doc repository only (and eventually other documentation versions).

Still pending to be done:
- [x] solve point 2 mentioned above
- [x] uncomment other jobs (now commented out to speedup tests)

Meanwhile I have used my personal token and push it to my fork of sardana-doc. One can check the results on https://reszelaz.github.io/sardana-doc/


.travis.yml
```diff
@@ -51,27 +51,8 @@ script:
...
-    fi
-
-doctr:
-  key-path : ci/github_deploy_key.enc
-  deploy-repo : sardana-org/sardana-doc
-  require-master: false
-  sync: true
-  built-docs: build/sphinx/html/
+  - if [ $TEST == "doc" ]; then docker exec -t sardana-test /bin/bash -c "cd /sardana ; sphinx-build -W doc/source/ build/sphinx/html" ; fi
+  - if [ $TEST == "doc" ]; then docker exec -t sardana-test /bin/bash -c "touch /sardana/build/sphinx/html/.nojekyll" ; fi
```
cpascual: This is a minor thing, but I think that forcing the .nojekyll file is no longer needed because it can be set as a project option (not 100% sure). Checking it is something that I had in my low-priority TO-DO list... maybe we can try it now.

doc/source/docs.rst
```diff
@@ -18,7 +18,6 @@ scientific installations.
    devel/index
    sep/index.rst
    Glossary <glossary>
-   other_versions
```
cpascual: Even if the other versions do not get automatically built, I think that it is interesting keeping a pointer to some selected older versions of the docs (ideally to the officially released versions, but if not, at least to the v2.x) so that someone using an older version could see the docs for their version. 

I do not see this as blocking or even worth investing too much time, but if it can be kept without much efort, I would keep it. If not, maybe a note somewhere indicating how an user can build the older docs from her own copy of the source could be added.

reszelaz: For the moment I just added a [guide](https://reszelaz.github.io/sardana-doc/other_versions.html) on how to build the docs for other versions.

Note that until now we did not really have other versions of the docs. The [page](https://sardana-controls.org/other_versions.html) with other versions just contains other branches uploaded to the upstream repo. No one never complained that the other versions were missing. Furthermore, before holidays the develop branch was already migrated to python 3. Since then also no one complained about the last python 2 version of the docs. That's why, IMHO, for the moment I would continue as we are. As an extra, with this PR we will have instructions on how to build other versions of the docs. If there is a strong interest from the Community for online other versions we can reconsider adding them.

.travis.yml
reszelaz: I looked in the GitHub repo (in the GitHub Pages zone) but could not find this parameter. Now I would not block this PR for that.

reszelaz:
Thanks for the review @cpascual. I think that all your comments are resolved.

I proceeded with the solution for point 2. This is short summary of what was done:
* I created @sardana-bot user and added as a collaborator to sardana-org/sardana-doc project (write access)
* I generated its personal token with _repo_ permissions and configured it as GITHUB_TOKEN in sardana-org/sardana project

When still doing the tests the build report some problmes with the deploy:

```
gh-token is invalid. Details: GET https://api.github.com/user: 401 - Bad credentials // See: https://developer.github.com/v3
dpl.2
Preparing deploy
failed to deploy
```

but actually the documentation is correctly deployed to the GitHub Pages. So for the moment I would not worry about it (maybe it is something related to my fork and it will work better for the sardana-org/sardana-doc project?).

I also changed the `.travis.yml` file to already point to the final repo and also reverted all the changes done for the testing purposes.

doc/source/docs.rst
cpascual: ok. I just added a small change to be more explicit about the command that needs to be executed for building the docs.

Also, as aside note: accepting this PR has implications on the debian packaging: it makes [this patch](https://salsa.debian.org/science-team/sardana/blob/master/debian/patches/0001-Remove-embedded-script-avoid-privacy-breach-lintian-.patch) unnecessary.

cpascual:
LGTM

PR #1225, opened 2019-11-09
Add basic Readme file

johnbickmore:
So you can see a description of the project on Github.


cpascual:
I like this.

I would include a link to the official docs page, though.
Something like [what we have for taurus](https://github.com/taurus-org/taurus)

reszelaz:
Thanks for pointing this out @johnbickmore. I can take care for adding the badges.

PR #1231, opened 2019-11-12
Write missing default attribute values when pool starts

HEnquist:
If a controller is deployed with sardanadsconfig the default values of attributes are not written to the database. Normally this is dine by the defctrl or defelem macros. Depending on the implementation of an affected controller this may or may not be a problem, but it can lead to strange and annoying errors. This PR adds a check when the Pool starts to write any missing default value after each device is started.


HEnquist:
Sorry this went to the wrong repo..

PR #1223, opened 2019-11-06
Hotfix 2.8.4

amilan:
Hotfix to add the fixes to the broken fscan macro (#1218, #1220 )

PR #1235, opened 2019-11-13
Make static manpages

reszelaz:
Manpages are dynamically recreated on every release. This is an extra effort
which does not add value. The help of Sardana commands does not change so often.
Make the manpages static with a very short description of commands.

This is related to this https://github.com/sardana-org/sardana/pull/1223#issuecomment-551022442.


cpascual:
Good. 
I'll do a debian packaging test to see if lintian is fine with these new files.

cpascual:
All is good. Merging

PR #1236, opened 2019-11-19
Add missing default attribute values on Pool start

HEnquist:
If a controller is deployed with for example [json2tango](https://github.com/MaxIV-KitsControls/lib-maxiv-dsconfig) the default values of attributes are not written to the database. Normally this is done by the defctrl or defelem macros. Depending on the implementation of an affected controller this may or may not be a problem, but it can lead to strange and annoying errors. This PR adds a check when the Pool starts to write any missing default value after each device is started.

### DIFF
--- a/src/sardana/pool/poolmetacontroller.py
+++ b/src/sardana/pool/poolmetacontroller.py
@@ -227,9 +227,9 @@ class DataInfo(object):
         return kwargs
 
     def __repr__(self):
-        return "{0}(name={1}, type={2}, format={3}, access={4})".format(
+        return "{0}(name={1}, type={2}, format={3}, access={4}, default_value={5})".format(
             self.__class__.__name__, self.name, DataType[self.dtype],
-            DataFormat[self.dformat], DataAccess[self.access])
+            DataFormat[self.dformat], DataAccess[self.access], self.default_value)
 
 # class PropertyInfo(DataInfo):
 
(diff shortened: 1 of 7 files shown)


reszelaz:
Just to keep updated the rest of the community, today we had a short meeting with @HEnquist and @julianofjm.
As the first approach we thought of changing a little bit the current strategy (the one currently implemented in this PR) of using threads to write the values on AttributeProxies to writing the `__value` properties in the `init_device` methods and delegate the job to the Tango memorized attribute restoration mechanism. Actually @HEnquist already tried it, but setting these properties using the Tango database object obtained from the ApiUtil in the device does not affect restoration of the memorized attributes. It looks like we are doing something wrong or as a bug in Tango/PyTango - to be investigated.

We also think that if it is possible and relativelly easy, the default values should be decoupled from the memorized values and set on the Sardana core level. This will have several benefits:
* it will work for non-memorized attributes
* default value changes (controller code changes) will take effect
* it may also facilititate similar solution for the controller properties (not sure)

So now we will investigate this last option. We will keep you updated..

reszelaz:
Hi @HEnquist,

Comming back to this PR, I think that the initial approach taken in this PR becomes outdated after our last discussion (more than a year ago). Then I will proceed to close this PR and opened #1458 to discuss the new approach. 

Hope you agree and again thanks for dedicating your time to this issue!

PR #1238, opened 2019-11-22
Fix #1237: Macro functions with result don't return any result

### LINKED ISSUE #1237: Macro functions with result don't return any result
When the following macro is executed I don't see any result in spock:
```python
from sardana.macroserver.macro import macro, Type

@macro(result_def=[['important', Type.Integer, None, 'answer to life, the universe and everything']]
def bla(self):
    return 42
```
Also calling this macro from another macro and getting the result, you get `None`.

tiagocoutinho:
Macro functions are just instances of the same MacroFunction class.
For every macro function call, the macroserver creates an instance
of MacroFunction.
The constructor of MacroFunction is already overwriting the values
of param_def and result_def so promoting hasResult() from
classmethod to method, should be sufficient to allow each macro
function to have distinct results.

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -2235,15 +2235,14 @@ class Macro(Logger):
         """**Unofficial Macro API**."""
         return self._pause_event.isPaused()
 
-    @classmethod
-    def hasResult(cls):
+    def hasResult(self):
         """**Unofficial Macro API**. Returns True if the macro should return
         a result or False otherwise
 
         :return: True if the macro should return a result or False otherwise
         :rtype: bool
         """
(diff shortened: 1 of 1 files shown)


reszelaz:
Many thanks Tiago! This works as expected. I think in the description you wanted to say [`MacroFunc`](https://github.com/sardana-org/sardana/blob/4ee47f0bd31cba2a54f25bac15159c2cba952459/src/sardana/macroserver/macro.py#L2482) class instead of [`MacroFunction`](https://github.com/sardana-org/sardana/blob/4ee47f0bd31cba2a54f25bac15159c2cba952459/src/sardana/macroserver/msmetamacro.py#L271).
Merging..

PR #1239, opened 2019-11-22
Fix macro function environment and hints

tiagocoutinho:
Seems that macro function environment and hints are swapped.
This PR fixes that


reszelaz:
It looks ok to me. Merging..

PR #1240, opened 2019-11-22
Fix RunMacro execution with string parameters without extra quotes

reszelaz:
In spock syntax when you use string parameters with white spaces you need
to use quotes. However when you use RunMacro Door's command, you already
pass string parameters as DevVarStringArray separate items. Then it
should not be mandatory to use quotes for string parameters with white spaces.
Fix it by quoting this parameter values when necessary.

One can test it with this kind of macro:

```python
class str_par(Macro):

    param_def = [["value", Type.String, None, "string parameter"]]

    def run(self, value):
        self.print(value)
```

Before, executing RunMacro command with ["str_par", "Hola, que tal?"] was giving error. With this PR it works. 

@sardana-org/integrators could you take a look on it? Many thanks!

### DIFF
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -1137,8 +1137,8 @@ class MacroExecutor(Logger):
             macro_params = par_str_list[1:]
 
             def quote_string(string):
-                # if string contains double quotes, use single quotes, otherwise
-                # use double quotes
+                # if string contains double quotes, use single quotes,
+                # otherwise use double quotes
                 if re.search('"', string):
                     return "'{}'".format(string)
                 else:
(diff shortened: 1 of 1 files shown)


bourtemb:
@reszelaz , I tested this patch with our macros which were no longer working as before and now it works fine again!
Thanks you very much!

reszelaz:
@sardana-org/integrators please don't review it yet. There are some tests failing. I will let you know whenever these are fixed.

reszelaz:
All the failing tests are fixed now. Finally I had to made the quoting simpler cause previously it was giving problems with the repeat parameters. Now it simply quotes any passed token with whitespaces.

@sardana-org/integrators, now I think it is ready for review.

reszelaz:
Thanks for the review @teresanunez. I'm merging then.

PR #1245, opened 2019-12-06
Console size py3

tiagocoutinho:
Use `os.get_terminal_size()` to determine console size instead of doing an expensive popen call.

Bonus: it works on windows ;-)

I did not use `shutil.get_terminal_size()` on purpose: it is about 500x slower and doesn't add anything useful


rhomspuron:
I tested it and it work as expected. I changes the console size and spock overwrite the lines without problems.

reszelaz:
Thanks to both! Merging.

reszelaz:
I just realized that I merged it too early. There were test failing on develop branch. @rhomspuron , @tiagocoutinho could you take a look on it, please? Now the tests are already failing on develop, so please open a new PR to fix them. Thanks!

PR #1246, opened 2019-12-06
Determine console size with ioctl instead of popen stty size

tiagocoutinho:
Use `os.get_terminal_size()` to determine console size instead of doing an expensive popen call.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/macroserver.py
+++ b/src/sardana/taurus/core/tango/sardana/macroserver.py
@@ -30,14 +30,18 @@ __all__ = ['BaseInputHandler', 'BaseDoor', 'BaseMacroServer',
 
 __docformat__ = 'restructuredtext'
 
+import os
 import sys
 import time
 import uuid
 import math
+import fcntl
+import struct
+import termios
 import weakref
 import threading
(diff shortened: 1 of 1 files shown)


reszelaz:
Thanks for the PR Tiago. I think that the correct one is #1245, right? If yes, please close this one.

tiagocoutinho:
This one is for the python 2 branch in case we want a hot fix. Should I close it?

The #1245 is for the develop branch (python 3)

reszelaz:
Ahh ok. If we want to integrate it will need to proceed with a normal [hotfix release process](https://github.com/sardana-org/sardana/wiki/How-to-Hotfix-Sardana). As on today this would be 2.8.5. So, if someone is interested just let us know. This issue affects macros producing fast outputs e.g. continuous scans when the [OutputBlock ViewOption](https://sardana-controls.org/users/spock.html#changing-appearance-with-view-options) is enabled. If no one is interested I would close the PR, but let's wait some time...

reszelaz:
Hi @tiagocoutinho,
There was no interest in hotfixing this for Sardana 2.8. I will proceed with closing this PR.
Thanks anyway!

PR #1252, opened 2019-12-11
Use magnitude of attributes values in PMTV

reszelaz:
PMTV is not ready to work with quantities and #1206 while removing
deprecation warnings introduced the use of quantities. Extract quantity
magnitude so the widget works as previously.


cpascual:
I am aware of time/resource constraints and such, but... wouldn't this be a good time for making PMTV quantity-aware?
:angel: :sweat_smile:

reszelaz:
Yes, this is a good point. However currently the PMTV (relative move) is broken when you use units. I would treat it in a separate feature (and the priority would need to compete with other issues in the backlog).

teresanunez:
Hi Zibi, thanks, IMO it could be merged.

reszelaz:
Thanks for the review Teresa! Merging then..

reszelaz:
There were some bugs in this PR. I just corrected them and pushed directly to the develop branch: f0eb792e

PR #1256, opened 2019-12-16
Spock abort macro (2nd Ctrl+C) and release macro (3rd Ctrl+C)

### LINKED ISSUE #34: spock returns prompt after ctrl+c even if the macro is still running (SF#47)
This issue was reported by Francois Fauth.

If in the macro code, a long time.sleep() gets called and the macro is aborted during this sleep, spock returns the prompt even if the macro is still running.

Furthermore it informs the user that stopping was correctly done:

~~~
^C
Ctrl-C received: Stopping... Done!
~~~
However if you ask the spock about the door state, it reports that macro is still running:

~~~
Door_XXX [58]: _DOOR.state()
Result [58]: PyTango._PyTango.DevState.RUNNING
~~~

This way user could get confused, and execute another macro. The subsequent macro gets queued, and does not start until the previous macro execution reaches a checkPoint() call.

Please see attached python module with two macro examples: `abort_fail` and `abort_pass`

Original Ticket: [sardana/tickets/47](https://sourceforge.net/p/sardana/tickets/47)

### LINKED ISSUE #978: Spock does not allow to abort macros
In spock documentation there is information on [how to stop macros](https://sardana-controls.org/users/spock.html#stopping-macros) but nothing on how to abort them...

I had always in mind that the second `Ctrl+C` should abort the macro.

On the last follow-up meeting @teresanunez reported that macro which can not stop its elements, what results in a macro which does not stop itself, stays running in spock. This is of course correct (requested in #34 and implemented in #596).

In this case the user should have a possibility to abort the elements. Now this possiblity does not exist. The second `Ctrl+C` hangs the spock completly making it necessary to restart the MacroServer.

This can be easily reproduced by:
1. Modifying the `DummyMotorController` and adding `StopOne` method which does nothing:
```python
    def StopOne(self, axis):
        self._log.info("StopOne(%d)", axis)
```
2. Executing a long move of a dummy motor e.g. `mv mot01 10000`
3. Stopping this move macro: `Ctrl+C`
4. Issuing a second `Ctrl+C`

I propose to implement a way of aborting a macro by issuing a second `Ctrl+C`.

This operation however, similartly to stopping a macro may not succeed. In this case we should place a configurable timeout in order to release the Door to the user (of course with an option to be set to infinite).

In this initial proposal, just to make it simpler, I do not consider calls to the macro API methods: `on_stop` and `on_abort`. But should also be taken into account. Now these are called after stopping/aborting the reserved elements. When implementing these we should also take into account https://github.com/sardana-org/sardana/issues/34#issuecomment-328788041.

reszelaz:
This PR adds to the Spock handling of the second and the third Ctrl+C.

The first Ctrl+C (already present in Spock) is stopping the Macro.
The macro stopping process in the first order calls the Stop command on all the reserved elements.
Then, whenever these react (change state to something different than Moving) it calls the `on_stop` macro method.

Currently the next Ctrl+C, which could be pressed by an "impatient"  user, or simply when the reserved elements does not react on Stop command hangs the Door.

This PR fixes #978 by capturing the second Ctrl+C . What happens then is that the macro gets abruptly aborted (we raise an exception in the macro execution thread asynchronously from another thread) and then we stop the process of stopping the elements started on the first Ctrl+C, and another process starts which calls the Abort command on all the reserved elements (including the ones which were previously stopped - this could be improved in the future). When these react (change state to something different than Moving) we call the on_abort macro method.

Then again, the next (third) Ctrl+C, which could be pressed by an "impatient"  user, or simply when the reserved elements does not react on the Abort command we release the macro by stopping the abort process started on the second Ctrl+C. This however may leave elements in the Moving state, but it allows to execute another macros afterwards.

I imagine that the on_stop method may be aborted by the second Ctrl+C but the on_abort method won't be aborted by the third Ctrl+C. Could someone test this, please? Also, do you have a preferences here?

I have tried it on a mv macro with a dummy motor which does not stop on the AbortOne method call. It would be great if you could test it in another "rare" scenarios, also very demanding, inpatient use. I limited myself to a gentle Ctrl+C sequences:)

Threading synchronization and notification could be solved on  many other ways, but I found this one (raising exception between threads) the easiest one taking into account that all the 
"waiting events" code was already existing (a big part of this code belongs to Taurus).

Also fixes #34 by giving more info to the user of what is going on in the stop/abort process.


teresanunez:
Hi @reszelaz, many thanks for this work. i have test it, what I see:
- #978 is solved, as you say. The second ctrl+c interrupt the first one and execute the abort function.
- if the AbortOne function hangt for some reason ... then I get again problems, one can execute
some macros but not all of them and the third ctrl+c brings the system in a strange state. I could
 manage for getting everything hanging and having to start the pool again.
- if AbortOne do not have problems, everything is fine ... 
I think it is nice because it solves the problem with the second ctrl+c and everything goes fine if
there is not problems with abort but one can not say that the system is free of hangings.

teresanunez:
Also playing with the state of the counters in the Scans I manage for bringing sardana in an state
in which nothing runs and I have to restart it again.

teresanunez:
I get this kind of error messages when killing the Pool that is hanging:

Thread 0x00007fe522f18700 (most recent call first):
  File "/usr/lib/python3.5/threading.py", line 293 in wait
  File "/usr/lib/python3.5/queue.py", line 164 in get

  File "/home/tnunez/.local/lib/python3.5/site-packages/taurus-4.6.1-py3.5.egg/taurus/core/util/threadpool.py", line 134 in run
  File "/usr/lib/python3.5/threading.py", line 914 in _bootstrap_inner
  File "/usr/lib/python3.5/threading.py", line 882 in _bootstrap

Thread 0x00007fe541fbb700 (most recent call first):
  File "/usr/lib/python3.5/threading.py", line 293 in wait
  File "/usr/lib/python3.5/threading.py", line 549 in wait
  File "/home/tnunez/sardana_ctrlc/build/lib/sardana/pool/poolmonitor.py", line 158 in run
  File "/usr/lib/python3.5/threading.py", line 914 in _bootstrap_inner
  File "/usr/lib/python3.5/threading.py", line 882 in _bootstrap

dschick:
Just a comment,

at MBI we often use the ipython syntax to write simple macros like
```
for i in range(10):
    %ascan dummy 0 10 100 0.1

```
In order to stop such macro-loops, one need to first stop the current macro with one Ctrl-C and then quickly, before the next one in the loop starts, one needs to kill the loop with another Ctrl-C.

I guess this might interfere with your current PR?

Best

Daniel

reszelaz:
Thanks Teresa and Daniel for your comments. 
I have added the following improvements:
* more info to the user about what is going on - requested in https://github.com/sardana-org/sardana/issues/34#issuecomment-328788041
* also whenever an object is successfully stopped it won't be aborted if one aborts the macro afterwards (can someone test it?)

Teresa, regarding your comments, could you please provide steps to repeat these tests to me?
I know that when one releases the macro (3rd Ctrl+C) and it was hang on aborting the reserved objects, then the on_abort method will be called. This one may take long time or even hang itself. What should we do then? allow the 4rd Ctrl+C? Capture any further Ctrl+C's and ignore them - meaning even wait forever in some cases? I'm slightly inclined to the first option. But would like to hear your opinion.

What is still missing to be implemented is that when we successfully stop the reserved objects (after 1st Ctrl+C) but the on_stop method takes long or even gets hung then the abort macro (2nd Ctrl+C) should break it and call the on_abort. Do you agree?

Daniel your https://github.com/sardana-org/sardana/pull/1256#issuecomment-566564910 won't be valid anymore whenever we integrate #767 - hopefully after finishing with this one.

teresanunez:
Hi @reszelaz, i see that in this new implementation AbortOne is not called if the motor was stopped
 (this is what you wanted to be checked?). About how to make the system hanging, it is enough
 with making the status of a counter after the acquisition staying in MOVING.

reszelaz:
> i see that in this new implementation AbortOne is not called if the motor was stopped
> (this is what you wanted to be checked?).

Yes, this was the improvemend done in fe26468. Thanks for testing it!

> About how to make the system hanging, it is enough
> with making the status of a counter after the acquisition staying in MOVING.

OK, I was always trying it with a motor and a mv macro. Now I will try with ct. I will let you know soon on it!

teresanunez:
Thanks @reszelaz. With only motors and mv I got the system hanging if in the Abort function I put
 a time.sleep(1000), for example. Sometimes it was recovered after this time sleep and another 
times not, depending on how oft I had done ctrl-c or what I have tried to run in the mean time

reszelaz:
I have made some further improvements:

Now it is possible to execute 4th Ctrl+C - it will try to release the macro from the on_abort method. When it is not possible e.g. one programmed a long sleep in the on_abort, then the 5th Ctrl+C will simply give up.
Here the same rule apply as for any other macro programming - you should never program a macro with a long sleep. I think we will not be able to break this sleeps - macros run in a worker thread.

I have also protected Spock from allowing to queue macro executions when the Door is in RUNNING state. I think it is much safer this way. So basically when someone pressed the 5th Ctrl+C and gave up in waiting for the macro to finish it won't be able to run any other macros. Then the user will need to either wait of restart the server (which of course will stop the on_abort execution). Of course I have added some information to the user on what is goind on.

Please could you try this version? For the moment it does not work with `ct` macros. I have realized the `ct` macro has some problems. It is not reserving the countable object so Ctrl+C is not doing what it is supposed to do. This may be a regression of #940 or #1049 or it was never working as expected. Do you remember if a long ct was ever possible to be aborted?

reszelaz:
The problem with stopping `ct` macro should be solved with #1264.
I have quickly tested this branch with the #1264 applied and it looks really good with stopping/aborting/releaseing ct macro.

reszelaz:
One more thing before testing. Whenever you are unable to on stop and abort (no, or buggy implementations of StopOne or AbortOne methods) an element and you recover the prompt in Spock you should not run a macro which will try to involve the same element in an operation. This will currently not work. This will need to be protected on the Pool side.

teresanunez:
I have test this, and I still manage for getting the system hanging ... but I think, I am really
 bringing the system to extreme conditions, using loops and time.sleep in the stop and abort
functions.

teresanunez:
Even if one can still crash the system, this implementation helps, so I merge it. Further improvements can be always done if needed.

reszelaz:
I just realized that this PR introduced a small bug. I just fixed it and pushed directly with https://github.com/sardana-org/sardana/commit/696382506b437f831480f08c99517f46667a01ad.

reszelaz:
I just found my design notes from the time I was working on this PR:

macro_stop_abort_release

This is a _pseudo_ activity diagram, where different colors represent different flows of actions explained on on the right.

dschick:
Hi all,

I fear that the current implementation of catching `Ctrl-C` does not allow to stop `ipython`.
First, it catches all `Ctrl-C` up to the 5th repetition and secondly, the proposal by @rhomspuron #767 also switches from stop to abort without interacting with `ipython`.

I see that this is not a primary Sardana issue and that there also other ways of creating sequences than with `ipython`.
However, it would be great if we could find a way to transfer an abort to the `ipython` layer (maybe with another command?)

Best

Daniel

PR #1258, opened 2019-12-18
advertise newfile macro in case no ScanDir or ScanFile is set

### LINKED ISSUE #1254: advertise newfile macro when no datafile is set
when there is no datafile configured, spocks prints:

`ScanDir is not defined. This operation will not be stored persistently. Use Use "expconf" (or "senv ScanDir <abs directory>") to enable it`

I guess, this should also state the `newfile` macro as an alternative to `senv ScanDir`

dschick:
Fixes #1254


reszelaz:
Thank you! Tested and works as expected.

PR #1260, opened 2019-12-19
Fix a showscan bug that loaded regular "showscan" and "showscan onlin…

ByRellex:
Fix a showscan bug that loaded regular "showscan" and "showscan online" while only calling "showscan online"


reszelaz:
Thanks for the PR @ByRellex! Your solution was still running the second subprocess. I changed it slighlty - as soon as we start the ShowScanOnline window I return from the magic.

I took a profit of this PR to remove the code for showscan online_raw. This does not work with "taurus trend " (pyqtgraph) and I think it does not have any sense to implement it. The current showscan online is much better. 

Will merge as soon as CI gets green.

PR #1262, opened 2019-12-20
Improve gscan to detect if a file or dir path is set but empty in the…

### LINKED ISSUE #1259: remove storage configuration in expconf does not undefine env variables
When one remove strogate configuration in expconf (leave empty text fields for File name(s) or Path in the Storage tab) then the environment variables stays configured:
```
Door_zreszela_1 [8]: lsenv
                  Name                                                          Value   Type
 --------------------- -------------------------------------------------------------- ------
               ScanDir                                                                   str
              ScanFile                                                           ['']   list
```

This prevents to print the warning/hint message as described in #1254.
There are two options:
* unset environment variables from expconf
* improve gscan to consider these values as unset

The second one seems to be more robust (also for cases when someone configured it like this from other macros.

ByRellex:
Improve gscan to detect if a file or dir path is set but empty in the environment. Sends a warning message indicating the fields are empty and rises an exception telling the field is not set and the results of the gscan will not be saved.
Fixes #1259

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -484,6 +484,9 @@ class GScan(Logger):
         macro = self.macro
         try:
             scan_dir = macro.getEnv('ScanDir')
+            if scan_dir == '' or None:
+                macro.warning('ScanDir value is empty')
+                raise Exception('ScanDir value is empty')
         except InterruptException:
             raise
         except Exception:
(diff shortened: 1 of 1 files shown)


teresanunez:
Thanks @ByRellex, i have tested it and it works as expected.

reszelaz:
Merging! Thanks @ByRellex and @teresanunez !

Issue #928, opened 2018-09-06
macro reserved objects in prepare are not stopped on StopMacro


reszelaz:
Just got this report from a colleague. Still not sure if this is a bug or a design limitation - it will require further investigation. Meanwhile I report it here:

This macro does not work - motor is not stopped on ctrl+c:

```python
class mv_prepare(Macro):
    param_def = [["motor", Type.Motor, None, "motor to move"],
                 ["positions", Type.Float, None, "position to move"]
                 ]

    def prepare(self, motor, position):
        self.output("In prepare")
        motion = self.getMotion([motor])
        motion.move([position])

    def run(self, motor, position):
        self.output("In run")
```

This works:

```python
class mv_run(Macro):

    param_def = [["motor", Type.Motor, None, "motor to move"],
                 ["positions", Type.Float, None, "position to move"]
                 ]

    def prepare(self, motor, position):
        self.output("In prepare")

    def run(self, motor, position):
        self.output("In run")
        motion = self.getMotion([motor])
        motion.move([position])
```

PR #1264, opened 2019-12-20
Allow to stop ct and uct

reszelaz:
Reserve obj in run method to allow its automatic stop.
Workaround for #928.

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -689,6 +689,8 @@ class ct(Macro, Hookable):
                                                     self.countable_elem_name))
             self.error(msg)
             return
+        # due to sardana-org/sardana#928 we need to explicitly reserve obj
+        self.addObj(self.countable_elem)
         # integration time has to be accessible from with in the hooks
         # so declare it also instance attribute
         self.integ_time = integ_time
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi @reszelaz, i can stop ct and uct with this change ... but I could also stop it without it ... I can not
reproduce a case where I could not stop those macros. In any case, with the change one can stop
them, it is not making anything wrong.

reszelaz:
That's strange. How long ct are you stopping? Please use a long
integration time so it would be visible if it does not stop. Also maybe
put some prints in the DummyCounterTimerController.StopOne method to see
if this one was called. This hints are just to reproduce the issue without
this PR. With this PR it should work as expected.
Many thanks for testing it!

teresanunez:
Hi @reszelaz, I put 100 seconds in ct, and I could stop the macro wihtout problems. I test it again.

teresanunez:
I know now the difference, none of my countertimers controllers (also not the Dummy one) has a 
explicit StopOne function, so only the ct.stop_one was called and ct was stopped. If I define
a StopOne function in the controller, so the ct macro is not stopped with the first ctrl+c but with
the second one, but this is the case with and without this PR. This PR does not make any
difference in my test.

reszelaz:
You are right. Now I can not reproduce this issue. I must had something messed in my environment.
Sorry that I made you loose your time. I'm closing this PR.

PR #1265, opened 2020-01-09
Allow MacroButton widget to be smaller

cpascual:
The minimum size the `MacroButton` (calculated by default by Qt) is too large in some cases.
Set the minimum size of the MacroButton widget to a smaller value (25x15 pixels for the button, plus whatever it is for the frame and the bar).

It can be seen by running the following snippet and reducing the widget size with the mouse:

```python
from taurus.qt.qtgui.application import TaurusApplication
from sardana.taurus.qt.qtgui.extra_macroexecutor.macrobutton import MacroButton

if __name__ == '__main__':
    import sys
    app = TaurusApplication(cmd_line_parser=None)
    w = MacroButton()
    model = 'door/bl97/1'   # <--your door name here
    w.setModel(model)
    w.setMacroName('lsmac')
    w.ui.button.setText('a')    
    w.show()   
    sys.exit(app.exec_())
```

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/ui/MacroButton.ui
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/ui/MacroButton.ui
@@ -1,7 +1,8 @@
-<ui version="4.0" >
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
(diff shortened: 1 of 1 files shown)


reszelaz:
Thanks for the review Aureo!

PR #1266, opened 2020-01-11
Fix meas grp configuration URIs back compat

reszelaz:
MeasurementGroup's configurations created with Taurus3 were supposed to be backwards compatible when migrating to Taurus 4. Now this is broken (most probably after accepting SEP18). Fix it by converting timer and monitor configurations to use full URIs (scheme + FQDN in authority).

We discovered this issue while migrating BL04 at ALBA to Python 3. @tiagocoutinho could you confirm that this reintroduces the backwards compatibility? Many thanks!


tiagocoutinho:
I have tested this PR in ALBA-BL04 and I confirm it fixes the problem.
Many thanks for fixing.

reszelaz:
Thanks for testing it! I'm merging.

PR #1269, opened 2020-01-15
Revert "Console size py3"

reszelaz:
Reverts sardana-org/sardana#1245


reszelaz:
Just trying to revert the merge to see if it fixes the testsuite.

reszelaz:
Finally I'm reverting the said PR cause there are comming PRs in parallel and the failing testsuite complicating their evaluation.

PR #1273, opened 2020-01-21
Call controller's DeleteDevice when Tango calls delete_device

reszelaz:
AddDevice is called on init_device. Call DeleteDevice on delete_device.
This is done for all elements proceeding from a controller e.g. motor, counter/timer, pseudomotor, pseudocounter, etc.

We have discovered this when working with @tiagocoutinho on [sardana-ni660x](https://github.com/ALBA-Synchrotron/sardana-ni660x/blob/master/sardana_ni660x/ctrl/Ni660XCTCtrl.py) at BL04 at ALBA. There we have realized that the DeleteDevice is not called on the server shutdown. It is only called when one call DeleteElement command on the pool. @sardana-org/integrators you are welcome to review this. @tiagocoutinho could you also test it at the beamline, please. Many thanks!

### DIFF
--- a/src/sardana/pool/pool.py
+++ b/src/sardana/pool/pool.py
@@ -576,6 +576,8 @@ class Pool(PoolContainer, PoolObject, SardanaElementManager, SardanaIDManager):
         self.remove_element(elem)
 
         self.fire_event(EventType("ElementDeleted"), elem)
+        if hasattr(elem, "get_controller"):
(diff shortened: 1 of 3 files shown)


tiagocoutinho:
LGTM!

reszelaz:
I'm merging it myself after review from Tiago.

Issue #1274, opened 2020-01-24
Remote recoder


stanislaw55:
Hello Sardana community
First of all, I am not Sardana expert and my question may be obvious but I want to be sure.
I am aware of Sardana recorders feature which allow to save data from scans in various forms. IS there any already functioning recorder which allows sending this data to remote server? 

Use case: company has strong policy about not storing any data on application servers. Any data must be stored on special storage servers. Sardana will run on application server but its data must somehow appear on storage server.

Thanks

cpascual:
Have you considered a remote FS?
Any of the file recorders combined with e.g. nfs, may be enough (that's how we use it at ALBA)

stanislaw55:
Hi @cpascual 
Yes, I've considered that but wanted to now if there's some recorder which already sends data somewhere, without the need for NFS

cpascual:
I see... 
AFAIK, the closest thing would be the  `JsonRecorder` for working with streams. Maybe you can implement something for whatever storage backend you have in mind inspired on that.

Disclaimer: I am also not an expert on existing recorders, so there may be something already done of which I am not aware.

stanislaw55:
Thank you @cpascual, I will look into `JsonRecorder`

reszelaz:
Hi, I think that MaxIV made some recorder to sent data to some remote service for storing the metadata, was it Apache Kafka? 
Maybe @julianofjm could comment here?

aureocarneiro:
@reszelaz was Apache Kafka, we just made a prototype and it is not in production. Here, we save the data directly to the storage cluster. We mount the storage on the sardana server computer.

@stanislaw55 here is a very simple example:

```python

"""This is the macroserver Kafka Recorder"""

__all__ = ["KafkaRecorder"]

__docformat__ = 'restructuredtext'

from uuid import uuid4
import requests
from sardana.macroserver.scan.recorder import DataRecorder


class KafkaRecorder(DataRecorder):

    def _startRecordList(self, recordlist):
        """
        Method to start the record.

        :param recordlist: Set of Records
        :type recordlist: RecordList
        """
        pass

    def _writeRecord(self, record):
        """
        Method executed while recording.
        
        :param record: Set of values measured at the same time.
        :type: Record
        """
        pass

    def _endRecordList(self, recordlist):
        """
        Method to execute when finishing the record.

        :param recordlist: Set of Records
        :type recordlist: RecordList
        """
        ###env = recordlist.getEnviron()
[...]

PR #1275, opened 2020-01-27
Add basic information to "how to write custom recorder" docs

reszelaz:
We had it on the todo list, but since #1274 raises a question about that I prefered to add already something to the docs. Yes, it is very minimal infomration, but may be already usefull. @sardana-org/integrators could you check that? Also @stanislaw55 it would be great if you could review it and comment (or maybe even complete:)) based on your experience in wrting recorders...

Thanks to all!

### DIFF
--- a/doc/source/devel/howto_recorders.rst
+++ b/doc/source/devel/howto_recorders.rst
@@ -48,8 +48,9 @@ the environment variables.
 Writing a custom recorder
 -------------------------
 
-This chapter (still under development) provides the necessary information to
-write recorders in Sardana.
+.. todo:: finish chapter based on user's input
+
+This chapter provides the necessary information to write recorders in Sardana.
 
 Before writing a new recorder you should check the `Sardana plugins
 catalogue <https://github.com/sardana-org/sardana-plugins>`_.
@@ -82,8 +83,8 @@ Configuration
 
 Custom recorders may be added to the Sardana system by placing the recorder
 library module in a directory which is specified by the MacroServer
-*RecorderPath* property. RecorderPath property may contain an ordered, 
-colon-separated list of directories.
+:ref:`RecorderPath <sardana-configuration-macroserver>` property.
+
 In case of overriding recorders by name or by file extension (in case of the
 file recorders), recorders located in the first paths are of higher priority
 than the ones from the last paths.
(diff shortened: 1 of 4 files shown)


teresanunez:
It looks fine to me. Only a typo: recerder -> recorder.

stanislaw55:
Thank you @reszelaz for this PR! This will definitely make it easier to write new custom recorders!

Unfortunately, I do not have any experience in writing recorders, only few macros and controller. However, I have few ideas how to improve this document:

Explanation of interface of methods - what is passed to them, what is expected to be returned.

Signalling of errors - does recorder have possibility to somehow inform user that something went wrong? How should it be done? Raising exception? Returning some special value? Maybe just logging of errors? In case of logging - what methods to use?

Those are my thoughts on writing recorders from perspective of maybe-soon-to-be Sardana recorder developer

doc/source/devel/howto_recorders.rst
```diff
@@ -48,7 +48,34 @@ the environment variables.
...
+write recorders in Sardana.
+
+Before writing a new recorder you should check the `Sardana plugins
+catalogue <https://github.com/sardana-org/sardana-plugins>`_.
+There's a chance that somebody already wrote one that fits to your needs.
+
+If finally you decide to write a new one below you can find some useful
+information:
+
+* Your recorder class would need to inherit from ``BaseFileRecorder`` and
```
cpascual: Shouldn't it inherit from `DataRecorder` instead of `BaseFileRecorder`?

For example, the `ShmRecorder` and the `OutputRecorder` inherit from `DataRecorder`

reszelaz: You are right! I put two options in the docs: if your destination is a file use BaseFileRecorder if not use DataRecorder. Thanks!

reszelaz:
Thanks for all the comments! I applied corrections from @teresanunez, @cpascual and @guifrecuni (passed verbally here at ALBA). 
Thanks to @stanislaw55 for your suggestions as well. I 100 % agree on them. However these require some more time and for the moment I can not dedicate it to this. I left explicit todo in the docs.

stanislaw55:
Nice to see this merged so quickly!
I left my comments more like a proposal and I totally understand your situation.
Anyway, thanks for reaction!

PR #1277, opened 2020-01-27
Protect against exceptions in MeasurementGroup.count_continuous

cmft:
The count_continuous method is not protected against possible
exceptions in the count_raw call (e.g. TimeOuts). Due to this the
subscription of all value buffers are kept.
This cause that the online recorder still reports the output even
the door is in alarm.


reszelaz:
Thanks for this PR! In the future we could also protect the `MeasurementGroup.subscribeValueBuffer` and store the state of which channels were actually subscribed and in the case of an exception also unsubscribe these. However for the moment the improvement that you propose is alreday a step forward! I'm merging... Thanks!

Issue #599, opened 2017-09-15
macroexecutor does not load macros if macroserver is started later on


mrosanes:
If MacroServer is not running when we execute the macroexecutor; and later we execute the MacroServer, the macros are not loaded in the GUI. We have to restart the macroexecutor to find the macros correctly loaded.

reszelaz:
Confirmed with taurus3 and taurus4.

reszelaz:
Fixed in #1278

PR #1278, opened 2020-01-28
Macro combo box load after start

ByRellex:
After looking at the problem for a while, finally removed the deprecated `setParentModel()` and directly used the `setModel()` of the widgets inside the parent `setModel()`.
This solves issue #599


cpascual:
Yes, @reszelaz : you are 100% right:
setting `modelInConfig=True` makes no sense when there is some other mechanism such as explicit user action, e.g. via a model selector, or a parent widget that sets it for subwidgets, or when it is hardcoded or set by configuration files, etc.

In other words: it essentially only makes sense when the model is given by a user action on the specific widget (e.g. via a drag&drop) **and** we want to record the result of that action to avoid repeating it on a later run.

reszelaz:
Good, then we need to make changes in this PR. I will push them directly without a PR.
In summary only the `TaurusMacroExecutorWidget` will have `modelInConfig=True`.

Carlos, I think you accidentally put model selector in the first group of use cases (no sense to use `modelInConfig=True`). I think that `TaurusForm` with `modifiableByUser=True` which models were configured with the model chooser at runtime should also use `modelInConfig=True`. This of course when our application wants to keep this configuration between then startups.

PR #1280, opened 2020-01-31
Make sure pyplot examples correctly refresh plot on the client

tiagocoutinho:
Newer versions of ipython / matplotlib seem to miss plot call refresh when calls to plot are executed in different threads.

This PR tries to make the pyplot macro examples work correctly and gives advice in the documentation


reszelaz:
Thanks for this investigation and the PR! Merging.

PR #1281, opened 2020-01-31
Avoid unnecessary creations of DeviceProxies in ascanct

reszelaz:
To determine if a channel is proceeding from Sardana or is an arbitrary Tango attribute it is enough to use TangoDeviceNameValidator.

@sardana-org/integrators could you please take a look on this one? @tiagocoutinho this may also improve the issues that we observe at BL04 so it would be nice if you could test it as well. Thanks to all!


tiagocoutinho:
LGTM! Nice enhancement :-)

reszelaz:
I'm merging it myself after review from Tiago.

PR #1286, opened 2020-02-10
New feature to save motor expert view and read widged

### LINKED ISSUE #277: Keep selection of the absolute or relative movement in PoolMotorTV (SF#432)
This issue is reported by Fulvio Becheri.

Motor Widget: in Napp experiment work with “Relative” movements but the widget by default starts with “Absolute”. It is not stored in any place (in a config file or somewhere else) and it can cause big problem if the user restarts the widget and then moves quickly…

Original Ticket: [sardana/tickets/432](https://sourceforge.net/p/sardana/tickets/432)

ByRellex:
New feature that saves motor expert view and write widged on perspective and LAST settings.
Fixes #277


reszelaz:
Thanks for this PR!
To test it I had to use https://github.com/taurus-org/taurus/pull/1061 together with changes from the latest taurus develop branch. Otherwise `taurus form` does not even start.
This works as expected. I just change the names of the configuration properties to be CamelCase.

PR #1293, opened 2020-02-19
Reset pending operations on relative movement

### LINKED ISSUE #1145: Bug in PMTV apply button
If you relative move a motor using PMTV (e.g. from 0 (current pos) to 1) and then press the apply button the motor are moved to the previous relative position (following the example, the motor will come back to 0)

ByRellex:
PMTV was storing pending operations from absolute widget when using relative widget, this causes the motor to move to the absolute position when pressing apply button on the relative mode when it shouldn't. Now the pending operations are reset on changing the widget and after moving on relative.
Fixes #1145


ByRellex:
Example:
- Actual position 80
- Write 100 on absolute, but don't press anything
- Change to moving relative and don't writte anything
- Press apply button

Behaviour without reseting on `emitEditingFinished()`:
Motor will move to 100 because it has saved the previous absolute movement even though it is on relative on your screen.

Behaviour reseting on `emitEditingFinished()`:
Motor will reset the previous absolute movement and will only move the relative movement.

reszelaz:
I was not able to reproduce this and with @ByRellex we agreed on removing this extra reset of pending operations. With resetting it on swtching between the editors (aboslute and relative) it is enough. Merging..

PR #1268, opened 2020-01-14
Protect get_terminal_size() call introduced by #1245

tiagocoutinho:
Fix sardana clients which don't have an interactive terminal attached to them

PR #1295, opened 2020-02-19
DEBUG #1268: Console size py3

reszelaz:
This PR is just to debug travis-ci problems that suffer #1268. It is not intended to be integrated.


reszelaz:
This investigation is done. Closing

PR #1298, opened 2020-02-26
Protect worker threads with tango.EnsureOmnitThread

### LINKED ISSUE #1291: MS crash during the execution of two parallel scans
At BL01 we observed that the MS hangs (you can not read the DS status and you have to kill -9 MS) when we you launched from any Sardana client the next sequence:
- macro that run a time_scan in a second door (from first door)
- ascanc macro (from first door)
You can workaround the problem if you add a sleep time after starting the time_scan.

reszelaz:
Tango is not thread safe when using threading.Thread. One must use
omni threads instead. This was confirmed for parallel event subscriptions
in https://github.com/tango-controls/pytango/issues/307.

Customize taurus ThreadPool worker threads (https://github.com/taurus-org/taurus/pull/1081) to use
EnsureOmniThread (https://github.com/tango-controls/pytango/pull/327). Do it for:
* `sardanathreadpool.get_thread_pool` (global thread pool used in actions
  and dummy TG)
* `CAcquisition._thread_pool` (used to execute value_buffer callbacks -
  these employ recorders which could potentially subscribe to
  Tango events)

For the moment don't protect:
* `ThreadPool` used by `FuncGeneratorTestCase.thread_pool` (tests - so it is controlled usage)
* `threading.Thread` used by `PoolMonitor` (not really used)
* `threading.Thread` used by server execution task (not really used - so it is controlled usage)
* `threading.Thread` used by `PositionGenerator` (tests - so it is controlled usage)
* `threading.Thread` used by `DummyEventSource` (tests - so it is controlled usage)

It's very probable that this PR fixes #1291.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -52,6 +52,7 @@ from taurus.core.util.enumeration import Enumeration
 from taurus.core.util.threadpool import ThreadPool
 from taurus.core.util.event import CallableRef
 
+from sardana.sardanathreadpool import OmniWorker
 from sardana.util.tree import BranchNode, LeafNode, Tree
 from sardana.util.motion import Motor as VMotor
 from sardana.util.motion import MotionPath
@@ -1962,7 +1963,7 @@ class CAcquisition(object):
 
     def __init__(self):
         self._thread_pool = ThreadPool(name="ValueBufferTH", Psize=1,
-                                       Qsize=100000)
+                                       Qsize=100000, worker_cls=OmniWorker)
         self._countdown_latch = CountLatch()
         self._index_offset = 0
 
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1962,8 +1962,17 @@ class CSScan(CScan):
 class CAcquisition(object):
 
     def __init__(self):
-        self._thread_pool = ThreadPool(name="ValueBufferTH", Psize=1,
-                                       Qsize=100000, worker_cls=OmniWorker)
+        # protect older versions of Taurus (without the worker_cls argument)
+        # remove it whenever we bump Taurus dependency
+        try:
+            self._thread_pool = ThreadPool(name="ValueBufferTH",
+                                           Psize=1,
+                                           Qsize=100000,
+                                           worker_cls=OmniWorker)
+        except TypeError:
+            self._thread_pool = ThreadPool(name="ValueBufferTH",
+                                           Psize=1,
+                                           Qsize=100000)
         self._countdown_latch = CountLatch()
         self._index_offset = 0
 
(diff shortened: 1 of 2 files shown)


reszelaz:
Note that the travis-ci will fail until:
* we https://github.com/taurus-org/taurus/pull/1081 gets merged
* we update sardana-test docker image with the latest taurus

reszelaz:
Still failing. Is it something related to this PR or a known issue?

reszelaz:
Maybe before testing this PR let's wait for travis-ci results. I have just triggered recreation of the sardana-test docker image so it includes the latest taurus. I will let you know whenever it is green.

How I tested it? First I confirmed that we were affected by this bug. Running the following two macros multiple times e.g. `repeat 1000 macro01` and `repeat 1000 macro02` in prallel on two different doors:

```python
@macro()
def macro01(self):
    # this can be any other Sardana elements
    names = ["ct0{}".format(i) for i in range(1, 9)]
    for name in names:
        dev = tango.DeviceProxy(name)
        id_ = dev.subscribe_event("state",
                                  tango.EventType.CHANGE_EVENT,
                                  tango.utils.EventCallback())
        dev.unsubscribe_event(id_)


@macro()
def macro02(self):
    # this can be any other Sardana elements
    names = ["ct0{}".format(i) for i in range(1, 9)]
    for name in names:
        dev = tango.DeviceProxy(name)
        id_ = dev.subscribe_event("state",
                                  tango.EventType.CHANGE_EVENT,
                                  tango.utils.EventCallback())
        dev.unsubscribe_event(id_)
```

And I could observe periodically `API_EventTimeout` on the MacroServer output.

Then with this PR it is not reproducible.

Note1: I had to use Pool and MacroServer as separate servers.
Note2: It is necessary to use the latest taurus and PyTango from develop together with this PR. To compile PyTango I did it with: `python3 setup.py develop --user`. Then it takes precedence over the PyTango installed in the OS.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -1961,8 +1962,17 @@ def scan_loop(self):
...
+        # remove it whenever we bump Taurus dependency
+        try:
+            self._thread_pool = ThreadPool(name="ValueBufferTH",
+                                           Psize=1,
+                                           Qsize=100000,
+                                           worker_cls=OmniWorker)
+        except TypeError:
+            self._thread_pool = ThreadPool(name="ValueBufferTH",
+                                           Psize=1,
+                                           Qsize=100000)
```
tiagocoutinho: I would prefer that sardana fails if we are not using a minimum taurus with the worker_cls feature.
If we hide it the users may have this bug and don't know about it

reszelaz: Yes, I agree with you. To do that we still need to ask taurus and PyTango to bump the version numbers and then bump the requirement. I propose that we defer it until the release.

tiagocoutinho: Ok, I understand that it is complicated to have to wait for new taurus and PyTango versions.
Maybe we can put a warning saying that the user might be affected by the bug?

reszelaz: I added warning logs in 6d2583d. Thanks!

teresanunez:
I have not run it, but IMO it can be merged.

reszelaz:
Many thanks @teresanunez and @tiagocoutinho for the reviews. I'm merging.

PR #1300, opened 2020-03-01
Fix clear operation (motion & acq) and reset of acq ctrls dicts

reszelaz:
This PR fixes two problems which affects very fast operations.

**Problem affecting acquisitions**

Since 6a80fdc ctrls were reset as _finish_hook_ to allow free readout of Value and ValueRef attributes after acquisition has finished (after SEP17 this won't be needed cause Sardana will always keep in memory the last acquisition results).
This is buggy in the case of very fast acquisitions - the elements were already released from the acquisition context and new acquisition could overlap with the finish_hook.

Avoid this problem by resetting the controllers in the custom `AcquisitionBaseContext`.

**Problem affecting motions and acquisitions**

`OperationConext` is wrongly cleared twice - within the `action_loop` and when the `OperationContext` exits. In the case of fast motions and acquisitions this may cause that when leaving the previous operation context we clear the operation context of the next one. Call `set_state_info` as _finish_hook_ so it will be called when the operation context is cleared already.
In the case of motion also protect it with lock (for unknown reasons it was not protected).

@tiagocoutinho it would be great if you could review and test it at the BL04. Many thanks!

### DIFF
--- a/src/sardana/pool/poolacquisition.py
+++ b/src/sardana/pool/poolacquisition.py
@@ -37,6 +37,7 @@ import time
 import weakref
 import datetime
 import traceback
+import functools
 
 from taurus.core.util.log import DebugIt
 from taurus.core.util.enumeration import Enumeration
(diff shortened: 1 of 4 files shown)


tiagocoutinho:
Great!
I just tested on ALBA BL04-MSPD beamline and works like a charm :-)
LGTM

reszelaz:
@tiagocoutinho thanks for the review and tests. I'm merging it.

PR #1302, opened 2020-03-03
Avoid using AttributeProxy when checking limits

reszelaz:
As it was already identified in https://github.com/sardana-org/sardana/pull/663#issuecomment-357734482 `AttributeProxy` creation could be avoided and device `WAttribute` object could be used directly. Do it for pseudo motors and motor groups to avoid problems due to https://github.com/tango-controls/pytango/issues/315.

@tiagocoutinho it would be great if you could review it - it affects BL04 at ALBA. Thanks a lot!


tiagocoutinho:
Avoiding to create an AttributeProxy to local objects is a very nice change.
LGTM

reszelaz:
Thanks for the review @tiagocoutinho . Merging...

Issue #1188, opened 2019-09-02
test_stop_meas_cont_acquisition_3 test randomly fails on travis-ci


reszelaz:
This can be observed in this [build](https://travis-ci.com/reszelaz/sardana/jobs/230121128).

```console
======================================================================
FAIL: test_stop_meas_cont_acquisition_3 (sardana.tango.pool.test.test_measurementgroup.TangoAcquisitionTestCase)
Stop of the synchronized acquisition with two channels from the same controller using hardware trigger
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/taurus/test/base.py", line 113, in newTest
    return helper(**helper_kwargs)
  File "/usr/local/lib/python3.5/dist-packages/sardana-3.0.0a0-py3.5.egg/sardana/tango/pool/test/test_measurementgroup.py", line 253, in stop_meas_cont_acquisition
    self.assertEqual(state, desired_state, msg)
AssertionError: tango._tango.DevState.FAULT != tango._tango.DevState.ON : channel tango://sardana-test:10000/expchan/_test_ct_ctrl_1/1 state after stop is FAULT (should be ON)
----------------------------------------------------------------------
Ran 307 tests in 833.453s
```

reszelaz:
I'm forced to reopen it. It still fails..
Anyone know an easy way to either to upload a log file as artifact or eventually increase the job output limit in order to cat the log file there?

PR #1306, opened 2020-03-10
Adapt to new taurus behavior of cmd_line_parser

reszelaz:
https://github.com/taurus-org/taurus/pull/1089 changes default behavior of cmd_line_parser kwarg of TaurusApplication. If not passed Taurus won't provide any parser. Use get_taurus_pareser (OptParse) and pass it to the TaurusApplication.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/showscanonline.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/showscanonline.py
@@ -102,7 +102,7 @@ def main(group, taurus_log_level, door):
     from taurus.qt.qtgui.application import TaurusApplication
 
     app = TaurusApplication(app_name='Showscan Online', org_domain="Sardana",
-                            org_name="Tango communinity")
+                            org_name="Tango communinity", cmd_line_parser=None)
 
     assertPlotAvailability()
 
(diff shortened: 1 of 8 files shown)


reszelaz:
Thanks for the review! I reply for each of them:
* showscanonline.py - already uses click but I removed `cmd_line_parser=None` in this case
* macroexecutor.py, hklscan.py, ubmatrix.py, diffractometeralignment.py were correcly passing parser (the grep that you used gave false positive result).
* I adapted the rest of them.

> Also, the testsuite is failing (but it does not look as a related issue to me)

This test seems to be buggy see: https://github.com/sardana-org/sardana/issues/1188.

src/sardana/taurus/qt/qtgui/extra_sardana/showscanonline.py
```diff
@@ -102,7 +102,7 @@ def main(group, taurus_log_level, door):
     from taurus.qt.qtgui.application import TaurusApplication
 
     app = TaurusApplication(app_name='Showscan Online', org_domain="Sardana",
-                            org_name="Tango communinity", cmd_line_parser=None)
+                            org_name="Tango communinity")
```
cpascual: Fine for me as it is, but maybe it is better to leave it explicit in order not to make it incompatible with taurus 4.5

reszelaz: Good point. I will revert this commit and then merge it myself. Thanks!

reszelaz:
Mergin after positive review from @cpascual

Issue #792, opened 2018-06-26
Problems in starting measurement group elements should stop the measurement action


reszelaz:
This issue was observed at MaxIV. But it is a more general problem, for example see #764 and #587 which are related.

Some example of cases which are not well treated:
* If `PoolSynchronization.start_action` fails, the experimental channels which were started previously would continue waiting for triggers/gates.
* If one of the experimental channels fails in `PoolAcquisitionBase.start_action`, the experimental channels started previously in the same action would continue running.
* ...

One of the approaches could be to treat this cases as early as possible - in the `PoolAcquisition` or `PoolMeasurementGroup` (still on the Pool side). This would require that all of the acquisition sub-actions share a common *token*. If one of the sub-actions encounters a problem, it mark the token so  the rest of them sub-actions could realize about it and implement a procedure to get into a safe state e.g. stop the involved elements. Most probably it would require to properly implement the action's states (this concept is already partially implemented).

reszelaz:
I think we could reuse the emergency stop feature present in the actions already.

PR #1308, opened 2020-03-13
Dump info on channels if MG acq fails in step scan, ct and uct 

reszelaz:
MeasurementGroup may get into the FAULT state mainly due to these three reasons:
* software synchronized acquisition startup (load or start sequence) fails  
* one of its channels reports FAULT state
* the last readout of State or Value/ValueRef fails.

This is only reported to the user with the exception message: `Exception: Measurement group ended acquisition with Fault state`.

Similarly to dumping information about moveables when motion fails also dump it for channels in the case the measurement group acquisition fails. 

This is something what we discussed on the last follow-up meeting based on a poor experience when debugging a problem at DESY. I think this PR adds more verbosity about the measurement group problems. For example, now we would get:

```
Door_zreszela_1 [17]: ascan mot01 0 1 1 0.001
Operation will be saved in /tmp/test.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #175 started at Fri Mar 13 23:00:05 2020. It will take at least 0:00:02.830427
 #Pt No    mot01      ct01     twod01      dt
Report: Stopped at step #0 with:
ct01:
       State: Fault
      Status: ct01 is Fault
              Exception: ERROR STATE
twod01:
       State: Moving
      Status: twod01 is On
              Stopped
Operation saved in /tmp/test.h5 (HDF5::NXscan)
Scan #175 ended at Fri Mar 13 23:00:06 2020, taking 0:00:00.262264. Dead time 100.0% (motion dead time 10.0%)
An error occurred while running ascan:
Exception: Measurement group ended acquisition with Fault state

```
Note that there are still things to improve. For example, the channel's Status could correspond to the last readout done by the acquisition action. Now it is not the case. The Status information is read via Tango when composing the information, so an extra `StateOne` call is performed.

@teresanunez it would be great if you could review and test it. Thanks!

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -659,7 +659,16 @@ def _value_to_repr(data):
         return data
 
 
-class ct(Macro, Hookable):
+class _ct:
+
+    def dump_information(self, elements):
+        msg = ["Elements ended acquisition with:"]
+        for element in elements:
+            msg.append(element.information())
+        self.info("\n".join(msg))
+
+
+class ct(Macro, Hookable, _ct):
     """Count for the specified time on the measurement group
        or experimental channel given as second argument
        (if not given the active measurement group is used)"""
@@ -701,7 +710,20 @@ class ct(Macro, Hookable):
         for preAcqHook in self.getHooks('pre-acq'):
             preAcqHook()
 
-        state, data = self.countable_elem.count(integ_time)
+        try:
+            state, data = self.countable_elem.count(integ_time)
+        except Exception:
+            if self.countable_elem.type == Type.MeasurementGroup:
+                names = self.countable_elem.ElementList
+                elements = [self.getObj(name) for name in names]
+                self.dump_information(elements)
+            raise
+        if state != DevState.ON:
+            if self.countable_elem.type == Type.MeasurementGroup:
+                names = self.countable_elem.ElementList
+                elements = [self.getObj(name) for name in names]
+                self.dump_information(elements)
+                raise ValueError("Acquisition ended with {}".format(state))
 
         for postAcqHook in self.getHooks('post-acq'):
             postAcqHook()
@@ -726,7 +748,7 @@ class ct(Macro, Hookable):
             self.output(line)
 
 
-class uct(Macro):
+class uct(Macro, _ct):
     """Count on the active measurement group and update"""
 
     param_def = [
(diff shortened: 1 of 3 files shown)


jkotan:
Thanks @reszelaz  for PR. Is it true when you dump information about an element  its state is read second time in `PoolElement._information()`? 
If so when the first read of State is FAULT and second one is ON then you do not get the proper information.

teresanunez:
HI Zibi, thanks. I am afraid I can not reproduce the problem we had at the beamline in order
 to test it. What I think is that both states should be shown: the one read by the MG and being
 the cause of setting the MG to fault state, and the one that is composed afterwards by infomation(). I am afraid they are often different in the difficult cases where one can not find
the reason of the FAULT state.

reszelaz:
Yes, I agree with both of you.

Then we would need to listen to the Status events (as we now listen to the State events) on the MacroServer and then extract the Status information from the cache. When implementing this feature I already tried it and it worked. If you agree I will submit another commit with this change.

Then I think it is enough to dump just the Status from the last event and if one wants to get the new one (read again from the hardware) would need to execute: `<channel>.State()` or `<channel>.Status()`. What do you think?

teresanunez:
Hi Zibi, thanks, for me it is fine what you propose.

reszelaz:
> Is it true when you dump information about an element its state is read second time in `PoolElement._information()`?
> If so when the first read of State is FAULT and second one is ON then you do not get the proper information.

Sorry, in my earlier reply I had not exactly answered your question. The state won't be read unless the event is older than 3 s. I think usually it won't be, but it is safer to use the event value regardless of its timestamp. I proposed https://github.com/taurus-org/taurus/pull/1105 to make it possible.

Then as we discussed with Teresa on the previous follow-up meeting you are actually interested in the State and not in the Status so much. In this case, for the moment, I would not add listener to the Status attribute and make the extra status readout even if the status won't correspond to the one from the acquisition action. Of course printing the corresponding note to the user.

There is also one more detail. When dumping the information not all the element state events may already arrive. This is because the first element in the measurement group reporting the Fault state is already changing the measurement group state to Fault and this lets the scan proceed with the information dump. The events from the remaining elements may arrive later. This may cause situation like the one from the PR description - ct01 - Fault but twod01 - Moving. I will check tomorrow if I can make it work better without too much effort.

reszelaz:
> When dumping the information not all the element state events may already arrive. This is because the first element in the measurement group reporting the Fault state is already changing the measurement group state to Fault and this lets the scan proceed with the information dump. The events from the remaining elements may arrive later. This may cause situation like the one from the PR description - ct01 - Fault but twod01 - Moving.

I have investigated this a little and I don't have it very clear if we want to change the current behavior - whenever there is one element in Fault the whole measurement group goes into the Fault. The same is with the Alarm. For example the _emergency stop_ feature in the case of a multiple elments move is based on that. And I think the emergency stop feature also makes sense for the measurement group. This I think would be a solution to #792. So, I propose to docouple this two issues: dumping info and emergency stop.

Then I have added the timestamp information to the dumped state and status and the information looks like this now:

```
Door_zreszela_1 [1]: ct
Fri May 08 16:09:38 2020

Elements ended acquisition with:
ct01:
       State: Fault (16:09:39.558505)
      Status: ct01 is Fault
              Exception: Communication error (16:09:39.573659)
zerod01:
       State: On (16:09:39.571555)
      Status: zerod01 is On
              OK (16:09:39.573874)
An error occurred while running ct:
Exception: Measurement group ended acquisition with Fault state
```
Then I'm merging it as it is and whenever https://github.com/taurus-org/taurus/pull/1105 gets accepted into Taurus we will use this feature to ensure that the state information comes from cache even if it is older than 3 s.

PR #1309, opened 2020-03-13
Fix dummy C/T and 2D ctrls in the case of unfinished start

### LINKED ISSUE #1188: test_stop_meas_cont_acquisition_3 test randomly fails on travis-ci

reszelaz:
When StartAll method is not called StateOne and ReadOne raises
exception - start_time is not set. Avoid this problem by detecting and
dealing especially with this situation.

Note that this most probably fixes #1188 - in the said test the measurement group is stopped after 200 ms. It may be that if the stop interrupts the start then the subsequent StateOne will fail. I will rebuild it several times. If I don't see this test to fail I will merge it myself.

### DIFF
--- a/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py
+++ b/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py
@@ -139,7 +139,7 @@ class DummyCounterTimerController(CounterTimerController):
         if self._armed:
             sta = State.Moving
             status = "Armed"
-        elif axis in self.counting_channels:
+        elif axis in self.counting_channels and self.start_time is not None:
             channel = self.channels[idx]
             now = time.time()
             elapsed_time = now - self.start_time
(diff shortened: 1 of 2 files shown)


reszelaz:
I have re-run the tests on travis-ci 10 times, without any failure. I'm not 100 % sure that #1188 will be fixed but the dummy ctrl changes are safe and necessary. I'm merging.

PR #1310, opened 2020-03-17
Fix dummy motor velocity

reszelaz:
In Sardana Motor velocity assumes user unit / s. Dummy controller interprets it in steps / s. Fix it and consider step_per_unit.

@sardana-org/integrators can you please take a look. This is very trivial change. Many thanks!


reszelaz:
Thanks for the review Teresa! I'm merging.

PR #1312, opened 2020-03-20
WIP: Code to test taurus#1093

reszelaz:
This PR is to test https://github.com/taurus-org/taurus/pull/1093. It basically:
* reverts Sardana workaround for PyTango#292 - note that not all of this revert is to be pushed into develop 
* implements the _zombie_ concept in Sardana-Taurus extensions
* use the above concept in tests

Note: the tests will certainly fail cause we will need to update Taurus on the docker image first.


reszelaz:
The corresponding Taurus PR were closed so there is no more need to mantain this one open.

Issue #1319, opened 2020-04-03
wrong data parameter in SpockBaseDoor.processRecordData


jkotan:
Hi,

from time to time we has observed  the during starting session spock the following error (I 've menaged to reproduce is only one time):
```
Traceback (most recent call last):
  File "/home/jkotan/testinst/lib/python2.7/site-packages/sardana-2.5.1.8-py2.7.egg/sardana/spock/spockms.py", line 471, in processRecordData
    if data['type'] == 'function':
KeyError: 'type'

If you suspect this is an IPython bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@python.org

You can print a more detailed traceback right now with "%tb", or use "%debug"
to interactively debug it.

Extra-detailed tracebacks for bug-reporting purposes can be enabled via:
    %config Application.verbose_crash=True
```
This error is caused because `data` parameter of the processRecordData in this case does not have `type` keyword. It only contains a stack of RecordData, e.g.
```
[...]

PR #1320, opened 2020-04-03
add a check for type keyword in spockms

jkotan:
It contains workaround for #1319


reszelaz:
Hi Jan,
Thanks for the PR. I also added check of type (dict) as you pointed in https://github.com/sardana-org/sardana/issues/1319#issuecomment-615144107.
I took a profit of this PR to add a default codec in Tango server and updated the `Macro.sendRecordData` docstring. I'm merging.

Issue #949, opened 2018-09-25
Error reloading macros once they were overwritten


teresanunez:
I have observed this bug trying to reload a macro after being overwritten:

if i overwrite a macro, let's say lsenv, and then I remove my version from my private macrolib, reload this macrolib and reload the macrolib containing the original lsenv, env.py, this original lsenv can not be loaded any more, one get the error message:

Door_issue930_1 [7]: relmaclib env
An error occurred while running Macro 'relmaclib(env) -> 2261d55e-c098-11e8-8f3f-901b0e40a03c':
'lsenv'

lsenv can only be loaded again from my private macrolib but not from env.py

reszelaz:
Thanks for the report! Could you also paste the exception traceback from the MacroServer process if there was any?

teresanunez:
Hi @reszelaz, the MacroServer does not print any more information than spock:

SardanaTP.W005 ERROR    2018-09-26 09:49:17,741 Door/issue930/1: An error occurred while running Macro 'relmaclib(env) -> abc0fb7c-c160-11e8-a97b-901b0e40a03c':
'lsenv'

teresanunez:
Another observation, if I make the reload of env.py before deleting lsenv from my private library, the reload is made correctly and I have the original lsenv. Then I remove lsenv from my lib and relaod my
lib and if I make a reload of env I get also the error ... even if the last loaded lsenv was the original one.
Here the steps:
Door_issue930_1 [4]: relmaclib mytests
mytests successfully (re)loaded (found 65 macros)

Door_issue930_1 [5]: lsenv
fake

Door_issue930_1 [6]: relmaclib env
env successfully (re)loaded (found 11 macros)

Door_issue930_1 [7]: lsenv
   Name   Value   Type

I remove lsenv from mytests.py

Door_issue930_1 [9]: relmaclib mytests
mytests successfully (re)loaded (found 64 macros)

Door_issue930_1 [10]: relmaclib env
An error occurred while running Macro 'relmaclib(env) -> 92522e44-c161-11e8-a693-901b0e40a03c':
'lsenv'

Issue #930, opened 2018-09-06
Warn when macros gets overriden


reszelaz:
Reported by @guifrecuni

cmft:
Hi @guifrecuni and comunnity,

I agree with the proposal but I have a cosmetic doubts about how `lsmac` macro should show the overwritten macros:

I propose some options:  

```
Door_cfalcon_1 [1]: lsmac umv mv
   Name   isOverwritten                                                                             Location
 ------ --------------- ------------------------------------------------------------------------------------
     mv            True                                            /home/cfalcon/tools/mis_macros/myascan.py
    umv                   /home/cfalcon/tools/sardana-sardana.git/src/sardana/macroserver/macros/standard.py
```
or
```
Door_cfalcon_1 [1]: lsmac umv mv

Note: if [*] is present together with the macro name means that a macro with the same name 
is defined in other module with less precedence and it has been overwritten. 
     Name                                                                             Location
 -------- ------------------------------------------------------------------------------------
   mv [*]                                            /home/cfalcon/tools/mis_macros/myascan.py
      umv   /home/cfalcon/tools/sardana-sardana.git/src/sardana/macroserver/macros/standard.py
```
or
```
     Name                                                                             Location
 -------- ------------------------------------------------------------------------------------
   [*] mv                                            /home/cfalcon/tools/mis_macros/myascan.py
      umv   /home/cfalcon/tools/sardana-sardana.git/src/sardana/macroserver/macros/standard.py
```
Comments and ideas are welcome.

guifrecuni:
I prefer the [*] idea over a specific column.
I'd show the NOTE at the end of the list (last line, warning color) and only if there is at least one macro with such a situation

rhomspuron:
In addition, we can create another macro like `lsoverridden` (with a shorter name) to show only the overridden macros, because if there are many macros (like in claess ~1000) the lsmac is not useful.

reszelaz:
@rhomspuron would be adding a new parameter to the `lsmac` e.g. `only_overridden` work for you?

Issue #566, opened 2017-08-28
edctrl and edctrllib to use Git as a VCS


daneos:
One of the ideas from Sardana Workshop at Solaris.

If you make amendments to a controller (or a library of those), it would be great if it would automatically save the file and make a Git commit there (provided the controller is in a Git repository). That way you'd be able to keep track of every change that someone might make to your production controllers.

cpascual:
Maybe this could be implemented via a hook mechanism (in order not to impose a given version control system)

daneos:
Of course, git is only an example here.

Issue #563, opened 2017-08-28
Plugin class inheritance


daneos:
One of the ideas from Sardana Workshop at Solaris.

The plugin classes needs to be defined in two different paths of the PoolPath.
Lets say that class `Foo` is defined in the lower order path than class `Bar`. Then class `Bar` can inherit from class `Foo`.
This should be verified and checked if we could documment it as a proper way of making the plugin class inheritance or we call it as a side effect.

PR #1328, opened 2020-04-30
SEP19 - Refactor plugin system

reszelaz:
I would like to start discussion on the refactoring of the Sardana plugin system. I have prepared some proposals in the form of [SEP19](https://github.com/reszelaz/sardana/blob/sep19/doc/source/sep/SEP19.md).
When reading it you will realize that there are several TODOs. I list them here so we can check them when agreed:

* [ ] How to deal with widgets? Shouldn't these be entry points to Taurus?
* [ ] Evaluate use of [importlib-metadata](https://importlib-metadata.readthedocs.io/en/latest/) project - standard from Python 3.8.
* [ ] Check if it will be possible to use dots in magic command completion e.g. relmaclib sardana_bl04.macro.hotblower
* [ ] Prepare implementaion plan

I think when workign on this we should also clarify solutions for the following tickets related to the plugins:

Macro specific:
* [ ] Unable to run macro: PyDs_PythonError UnknownMacro: Unknown macro macro1 (SF#401) #246
* [ ] Neither a created macro nor a macro from the macro manager is instance of an imported macro class (SF#402) #247
* [ ] Add verbosity to edmac macro (SF#479) #324
* [ ] wrong argout of the ReloadMacro and ReloadMacroLib tango cmds (SF#119) sourceforge #76
* [ ] ReloadMacro and ReloadMacroLib tango cmds are not verbose in case of errors (SF#118) #75
* [ ] relmaclib does not work with macros importing modules from PythonPath (SF#465) #310
* [ ] 'NoneType' object has no attribute 'getType' on editing macro in a new location (SF#455) #300
* [ ] MacroServer ReloadMacro failure (SF#444) #289
* [ ] Problems with edmac on Windows (SF#505) #349
* [ ] edmac raises an exception #520
* [ ] Error reloading macros once they were overwritten #949
* [ ] Warn when macros gets overriden #930

Controller specific:
* [ ] Insert new controller Library in Pool when it is Running. (SF#80) #53
* [ ] edctrl macro does not work (SF#472) #317
* [ ] Pool should not log an ERROR if hkl is not installed #394
* [ ] edctrl and edctrllib to use Git as a VCS #566
* [ ] Order of loading controllers when one imports another
  
General:
* [ ] Plugin class inheritance #563
* [ ] Module based plugin discovery #562

It is totally open for changes and comments. Thanks a lot in advance for looking into it!

### DIFF
--- a/doc/source/sep/SEP19.md
+++ b/doc/source/sep/SEP19.md
@@ -1,0 +1,277 @@
+	Title: Refactor plugin system
+	SEP: 19
+	State: DRAFT
+	Date: 2020-04-30
+	Drivers: Zbigniew Reszela <zreszela@cells.es>
+	URL: http://www.sardana-controls.org/sep/?SEP19.md
+	License: http://www.jclark.com/xml/copying.txt
+	Abstract:
+	Sardana comes with custom plugin system for controllers,
+    macros and recorders. While it worked pretty well for years
+    it has some limitations when used in advanced projects, complicates
+    configuration management and is a non-standard solution.
+    This SEP refactors the plugin system to use setuptools "entry points".
+    
+# Rationale and goals
+
+The current Sardana plugin system is a very powerful custom solution which
+strongest point is the possibility to add, remove and reload plugins at
+runtime. On the other hand it was demonstrated e.g.
+[sardana-org/sardana#246](https://github.com/sardana-org/sardana/issues/246)
+or
+[sardana-org/sardana#247](https://github.com/sardana-org/sardana/issues/247)
+that its assumption of plugin classes defined in simple python modules
+becomes a limitation. In addition installation of plugins requires a manual,
+or hard to script, step of configuring discovery paths.
+
+The goal of this SEP is to refactor the Sardana plugin system in such a way so
+it fully supports python programming principles e.g. plugin classes
+inheritance, code organization in packages and modules, etc. at the same time
+reducing to maximum the manual configuration steps.
+
+# Current Sardana plugin system
+
+The Sardana plugin system works similarly for controllers, macros and recorders.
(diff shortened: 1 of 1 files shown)


reszelaz:
After the short introduction of the SEP on the last follow-up meeting I have talk with @teresanunez and from the DESY POV:
> ..we distinguish between system Macros, DESY Macros, BL Macros and user Macros. Our BL scientists are well aware of this structure and they are used to it and they are writing Macros quite frequently. 

Thye are used to the current way of using discovery via MacroPath and eventual changes would become an extra effort for them. Their point is to not make things more complicated than they are.

So, I think this is more alligned with [Other option 2](https://github.com/reszelaz/sardana/blob/sep19/doc/source/sep/SEP19.md#other-options)

stanislaw55:
Hi @reszelaz 
I've been thinking about following idea:
* Completely refactor plugin system to using entry points on MacroServer side
* Reintroduce current path-based plugin system from Spock side.

Regarding client side plugins, I imagine it working something like this:
Path to plugins would be possible to configure in multiple places (command line, profile, etc.). Spock would load code for macros in specified path, just like it is done now. Then, if macro is already defined, skip. If it is not, send code of macro to MacroServer. This would also allow explicit reloading of macros. This could be beneficial in terms of not using NFS.

cpascual:
IMHO this should be coordinated with the planned homogeneization of plugins in Taurus ([TEP13](https://github.com/taurus-org/taurus/blob/develop/doc/source/tep/TEP13.md))

cpascual:
Hi, I am working on updating the [TEP13]. 
Since Sardana has its own SEP for this, I am copying in this comment the sardana-specific notes that existed in TEP13 (and which are now removed from TEP13) in case they can be of use here:

PR #1330, opened 2020-05-06
Build docs with Sphinx 3

reszelaz:
The documentation builds started to fail due to PyTango intersphinx mapping problems.
These appeared to be solved with the latests Sphinx. I made some effort to fix warnings after the Sphinx upgrade.

I take a profit of this PR to set default role in docs to obj. This way one could create cross-references (even intersphinx) with just back-quotes e.g.\`taurus.Device\` instead of :obj:\`taurus.Device\`

This is a house-keeping PR, so I won't bother you with the reviews and whenever it gets green in Travis I will auto-merge.

Initially it will fail - still waiting for a new version of docker image with the Sphinx v3. But I will retrigger later..

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -299,3 +299,13 @@ intersphinx_mapping = {
     'https://matplotlib.org/': None,
     'https://pythonhosted.org/guiqwt/': None,
 }
+
+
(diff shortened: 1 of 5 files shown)


reszelaz:
Now green, merging...

PR #1333, opened 2020-05-12
Register a TaurusValue factory for pool widgets

cpascual:
A recent PR in taurus (https://github.com/taurus-org/taurus/pull/1108) provides a new entrypoint for registering "TaurusValue item factories" (the entry point group name is "taurus.qt.taurusform.item_factories" )

Register one such factory to achieve the same result as with the old T_FORM_CUSTOM_WIDGET_MAP (i.e., using PoolMotorTV, etc. for sardana elements in a
TaurusForm)

If this PR is integrated, it can then easily be adapted to provide the same that was attempted with https://github.com/taurus-org/taurus/pull/1101, which is related to #1203

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,13 +84,13 @@ gui_scripts = [
 ]
 
 form_factories = [
-    "sdn_pool = sardana.taurus.qt.qtgui.extra_pool.formitemfactory:pool_item_factory"  # noqa
+    "sardana.pool = sardana.taurus.qt.qtgui.extra_pool.formitemfactory:pool_item_factory"  # noqa
 ]
 
 entry_points = {
     'console_scripts': console_scripts,
     'gui_scripts': gui_scripts,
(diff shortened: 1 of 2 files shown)


cpascual:
> Hi I think that the fix in I just fixed 16761d1. Then it works as expected.

I think that the fix should be as in b3d55cd because otherwise we would try to instantiate a None later on. In other words, we want this line raise an exception if key is not found

reszelaz:
> I think that the fix should be as in b3d55cd because otherwise we would try to instantiate a None later on. In other words, we want this line raise an exception if key is not found

Of course!

cpascual:
The name of the entry point group has been changed in taurus-org/taurus#1108, so I adapted this PR accordingly

reszelaz:
Merging since https://github.com/taurus-org/taurus/pull/1108 is alredy merged into taurus.

PR #1339, opened 2020-05-31
Rename Synchronization to SynchDescription

reszelaz:
As agreed in https://github.com/sardana-org/sardana/pull/867#issuecomment-561742834 this PR renames the MeasurementGroup's Synchronization Tango attribute, core property and Taurus extension methods.

In the above discussion there were two votes in favor of `SynchronizationDescription` (@teresanunez and @droldan) since this is more descriptive and one in favor of `SynchronizationPlan` (@rhomspuron) since this is shorter.
Finally I had chosen the name `SynchDescription` which I think is a good compromise between these two arguments. Hope you like it. 

This PR also:
* adds the `getSynchronization` and `setSynchtronization` MeasurementGroup configuration methods and their tests
* adds the synchronization parameter to `get_meas_conf` and `set_meas_conf` macro.

I will automerge if travis is green since this RP is just executing a rename which was previously agreed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,13 @@ This file follows the formats and conventions from [keepachangelog.com]
 
 ### Changed
 
+* Renamed _synchronization_ to _synch description_
+  * Tango MeasurementGroup `Synchronization` attribute to `SynchDescription`
+  * Core MeasurementGroup `synchronization` property to `synch_description`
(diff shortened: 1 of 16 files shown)


reszelaz:
I just added a TODO on raising exception when one sets Synchronization for channels which do not use it. I think this way the configuration will be better self-described - b50621c. Merging..

PR #1341, opened 2020-06-02
Add docs on how to access Tango in macros and controllers

reszelaz:
This documentation was requested by one of our scientists at ALBA. When migrating user macros to Python 3 we discovered that user macros were using several different ways of accessing to Tango devices. 

I tried to summarize what possibilities we have and give a simple rule of thumb on which one to use.

There is one more way which I have not documented yet. The `Macro.getDevice()` which currently does not give any benefit over `taurus.Device()`. But with the recent investigation of @tiagocoutinho on how Tango (and CORBA) optimizes socket connection from the client to the server I think we could change this method and make it maintain the reference to the `taurus.Device`. It is similarly done for the internal Sardana elements getters e.g. `getMotor`. @tiagocoutinho, if I understood well the connection is closed after some inactivity period. This way the only penalty of such a change would be the memory footprint of a `tango.DeviceProxy` and `taurus.Device` objects. What about the ZMQ socket for the configuration events?

Also I have not added to the _Accessing Tango from your controllers_ the possibility to use `sardana.pool.poolutil.PoolUtil` singleton. I think it has not real benefit over maintaining a reference to the `tango.DeviceProxy` in the controller itself. Unless we change the `PoolUtil` and maintain a gloabl proxy instead of one per controller instance.

### DIFF
--- a/src/sardana/pool/poolutil.py
+++ b/src/sardana/pool/poolutil.py
@@ -43,6 +43,16 @@ class _PoolUtil(object):
         return self
 
     def get_device(self, *args, **kwargs):
+        """Factory method to create a single instance the `tango.DeviceProxy`
+        per controller instance.
+
+        :param ctrl_name: Controller name to which assign the proxy object
+        :type ctrl_name: `str`
+        :param device_name: Tango device name
+        :type device_name: `str`
+        :return: single device proxy object
+        :rtype: `tango.DeviceProxy`
+        """
         ctrl_name = args[0]
         device_name = args[1]
         with self._lock:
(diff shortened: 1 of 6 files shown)


src/sardana/pool/poolutil.py
```diff
@@ -56,8 +66,15 @@ def get_device(self, *args, **kwargs):
         return dev
 
     get_motor = get_phy_motor = get_pseudo_motor = get_motor_group = \
-        get_exp_channel = get_ct_channel = get_zerod_channel = get_oned_channel = \
-        get_twod_channel = get_pseudo_counter_channel = get_measurement_group = \
-        get_com_channel = get_ioregister = get_device
+        get_exp_channel = get_ct_channel = get_zerod_channel = \
+        get_oned_channel = get_twod_channel = get_pseudo_counter_channel = \
+        get_measurement_group = get_com_channel = get_ioregister = get_device
```
cpascual, line 71: The use of backslash for breaking lines is discouraged by PEP8 . But using parentheses in this case is quite ugly (see below what, e.g. `black` would do), so I am not changing it:

```python
    get_motor = (
        get_phy_motor
    ) = (
        get_pseudo_motor
    ) = (
        get_motor_group
    ) = (
        get_exp_channel
    ) = (
        get_ct_channel
    ) = (
        get_zerod_channel
    ) = (
        get_oned_channel
    ) = (
        get_twod_channel
    ) = (
        get_pseudo_counter_channel
    ) = get_measurement_group = get_com_channel = get_ioregister = get_device
```

cpascual:
The CI is now failing because of outdated intersphinx links. I'll update them

PR #1342, opened 2020-06-04
Allow to return only state in StateOne method

### LINKED ISSUE #621: StateOne that returns only state fails
According to the [documentation](http://www.sardana-controls.org/en/latest/devel/howto_controllers/howto_controller.html#get-axis-state), the `StateOne` method should be able to return only state. Currently, this fails. Tried on counter/timer controller.

reszelaz:
On the controller level always return the state information in form of a tuple (even with just one element - state). The action then embeds this information in another tuple with the eventual exc_info or None. Then on the element level when interpreting the information consider all the cases:
- just a state
- state info with exc_info where state info contains
  - both state and status
  - only state

Fixes #621

This can be easily tested if changing
https://github.com/sardana-org/sardana/blob/f5c5d6f3625b2778fc20d152895e55d2babd0fee/src/sardana/pool/poolcontrollers/DummyCounterTimerController.py#L152
for:
```python
return sta
```


teresanunez:
Hi, I have just tried, and I get the motors in FAULT state if I return only state.

reszelaz:
Hi Teresa,
I have fixed the issue. Before it I have added some automatic tests (I should have tried from this).
Now there are 5 new tests for all the cases for a counter/timer and a motor. Actually I did not know that the motors were affected by the same bug.
The only thing is that the new tests won't be run by Travis until #1347 is merged. These new tests are developed with pytest. One can try them locally with:
```
pytest src/sardana/pool/test/test_poolmotor.py::test_state src/sardana/pool/test/test_poolcountertimer.py::test_state
```
Note that I have made a forced push so you will need to checkout a new branch if you want to test it.

reszelaz:
Hi @teresanunez , could you please check it again after my recent changes?

teresanunez:
Hi @reszelaz, i get now motors and countertimers in FAULT state.

reszelaz:
Hi @teresanunez,
Thanks for your thorough tests. I have corrected the issue with the motion and I think now everything should be ok. Could you please validate it again. Many thanks!

reszelaz:
Thanks for the review!

PR #1347, opened 2020-06-09
Run sardana tests with pytest

reszelaz:
Current sardanatestsuite duplicates some tests. Approx 50 tests are executed twice making the whole test sute longer of more than 4 min.
See for example: [this job](https://api.travis-ci.org/v3/job/694553069/log.txt) where:
```
test_meas_cont_acquisition_4 (sardana.pool.test.test_measurementgroup.AcquisitionTestCase)
```
is executed twice.
It seems like a bug in Python unittest - it is already `TestLoader.discover()` which returns duplicated tests.

Instead of investigating it change the test runner to pytest which could open doors for other interesting features comming with pytest.

Note that previously I have installed the latest pytest in sardana-test docker image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
   # run flake8 check on all python files in the project
   - if [ $TEST == "flake8" ]; then ci/flake8_diff.sh; fi
   # run the full testsuite
-  - if [ $TEST == "testsuite" ]; then docker exec sardana-test sardanatestsuite; fi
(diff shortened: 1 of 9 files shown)


reszelaz:
I'm auto-merging this one cause it is benefitial to other PR to have it fixed.

Regarding removing the old sardanatestsuite I talked about it with @cpascual and we agreed on doing the following:
* do not remove the `sardanatestsuite` console_script
* make it to not execute any tests and simply print that is is deprecated

Any corrections on it can be done afterwards with another PR.

PR #1354, opened 2020-06-14
Avoid extra state readout at the end of acquisition

reszelaz:
Acquisition's `action_loop()` exits the state poll loop when none of the
elements is in Moving or Running state. This indicates that the acquisition
has finished. This last state readout should be propagated to the
channels. Avoid the extra state readout which actually could change the
end acquisition state. This also introduces a small performance improvement - #161


teresanunez:
Hi Zibi, I have checked it with simulated devices and I do not see any improvement, but
 I do not see any disadvantage, so IMO it can be merged. Thanks.

reszelaz:
Thanks for the review @teresanunez . I'm merging.

Issue #735, opened 2018-04-16
wrong format of spec files from sardana 


jkotan:
It looks like spec files from sardana do not adhere to the specifications of the SPEC standard data file format from [Standard Data File Format](https://certif.com/spec_manual/user_1_4_1.html). This causes problems in reading them via `spec2nexus` https://github.com/prjemian/spec2nexus/issues/122 or `nexpy` https://github.com/nexpy/nexpy/issues/152

cpascual:
Before fixing this, I would like to confirm that the official ultimate reference for the spec file syntax is the one linked above. I say so because some years ago, while fixing some other issue in Sardana's spec recorder, we used some reference document for the spec file syntax which I believe it was much more extense and formal than the one in the link... but I cannot find it now among my files or in the web which makes me doubt  about my memory...  :-\

rhomspuron:
Hi,  you can find the information on the official specpython library. This block is not mandatory:
https://github.com/specpython/filespec/blob/master/specpython/spec_format.rst
For that reason the specpython and pymca can open the file without problems. 

On the other hand, the file generated by sardana use the motor label '#O' and position '#P' for the snapshots and it saves it on each scan.

prjemian:
@vrey01 : Can you comment on this?  Which is the definitive reference for the SPEC data file format?

vrey01:
The spec data format is described in detail in https://certif.com/spec_help/scans.html or, in spec, by typing "help scans". (look for paragraph "File Conventions")

The spec data format gives a list of annotation conventions but it does not specify a "minimal" set of annotations that should be present in a file.  For example, in the page referenced above, after the list of standard control lines, there is an example of a "simple data file" containing just "#S, #N and #L" lines.   Once this said,  as it could be expected, spec itself produces data with much richer annotation that the one in that "simple example", including a header block indicating "filename", "date of creation", "application name", "user name that created the file", "epoch (to precisely time-tag each data point)"...

The tools reading spec data files decide what is the minimal information needed to display data. For example, in the case of the "splot" (standard plot program distributed with spec) it is sufficient to have a "#S" and a "#L" control lines to recognise a file as a valid spec data file and display the scans included in it.  In the case of the "scans" unix tool (also distributed with spec) only the "#S" is necessary to properly manage spec files. Eventually each tool will decide how tolerant will be to missing control lines and how it will handle them.
[...]

PR #1355, opened 2020-06-17
Add custom user to scan

tiagocoutinho:
Allow scans to have specific user by defining the environment variable "ScanUser".


tiagocoutinho:
May help with #735

reszelaz:
Many thanks for the PR!
Is the idea behind to fill it with the user responsible for the entry? As it is in the [NXUser](https://manual.nexusformat.org/classes/base_classes/NXuser.html).
Could you please add this variable to the [docs](https://sardana-controls.org/users/environment_variable_catalog.html)?

reszelaz (approved):
I've tested it and works as expected. I've alos added the missing doc. We alrady reviewed them with @cpascual and @tiagocoutinho and we agreed to merge it. Thanks to all!

Issue #1008, opened 2019-01-09
epoch as default counter in addition to dt 


dschick:
as in spec, it is quite helpful to have a default column `epoch` in the data file which allows to plot all your scans over absolute time.

reszelaz:
We just discussed it on the last follow-up when @daneos raised a similar request from the Solaris side. From the [MINUTES](https://github.com/sardana-org/sardana-followup/blob/master/20181213/MINUTES.md):

> question: could an experimental channel be used simultaneouly by two doors?. Zibi answers that it should not work, if it works is a bug. It would need to change a lot of things for making this working. For the tango zerod should not be problem, tango would serialize it and read a tango attribute. But for others than zerod it will depend. In general each element in Sardana should be only used at once.
Gregorz wanted to read a Tango Attribute with the epoch time from different Doors. Zibi suggests to create the epoch time from the dt of the scan. The important thing is to save it in the file, and dt and a start time as saved (but dt is not relative to the start time), Zibi suggests to make it coherent and use then these information and not add any extra ones.

Would the coherent dt column relative to the scan start time (in case of Spec files present in the header) work for you?

We did not have in mind the additional epoch column but I would not discard it.

dschick:
thanks for the feedback,
of course the dt column and the start time of the scan is generally enough, but when I am doing a long scan macro (let's say over night) I would like to quickly plot a signal as function of absolute time e.g. in PyMCA selecting all scans of the macro. Then directly sees the trend of the whole measurement. This does not work with dt in PyMCA.

Having I guess having one epoch counter that can be used simultaneously by two doors is more about synchronisation issues, which I, for now, do not have in mind.

reszelaz:
> of course the dt column and the start time of the scan is generally enough, but when I am doing a long scan macro (let's say over night) I would like to quickly plot a signal as function of absolute time e.g. in PyMCA selecting all scans of the macro. Then directly sees the trend of the whole measurement. This does not work with dt in PyMCA.

I see the point. If the scan analysis programs are prepared to work with the epoch column it makes sense to add it.

> Having I guess having one epoch counter that can be used simultaneously by two doors is more about synchronisation issues, which I, for now, do not have in mind.

However I would not implement it as a Pool element e.g. counter, but similarly to the `dt` simply read the timestamps and add them to the record.

Continuing with the requirements, should we allow having both `dt` and `epoch` at the same time in the scan record. Or make it optional to choose between one or the other, for example with an environment variable?

[...]

PR #1356, opened 2020-06-17
Add missing #E (epoch) to spec recorder

tiagocoutinho:
The spec scan recorder is missing the #E (epoch) tag.

This PR adds it.

### DIFF
--- a/src/sardana/macroserver/recorders/storage.py
+++ b/src/sardana/macroserver/recorders/storage.py
@@ -316,7 +316,7 @@ class SPEC_FileRecorder(BaseFileRecorder):
 
         # datetime object
         start_time = env['starttime']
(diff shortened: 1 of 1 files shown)


tiagocoutinho:
May help with #735

tiagocoutinho:
Misunderstood the meaning of #E.
The current implementation is not correct

dschick:
hi @tiagocoutinho thanks for working on the EPOCH issue

I think the `#E` tag is just written once to the SPEC file at its beginning (as you might already learned yourself)
Anyhow, SPEC has a default counter named `EPOCH` that I find very usefull.

There is already an open discussion #1008 and there are some conterTimerControllers around:
https://github.com/MBI-Div-B/sardana-EpochCounterTimerController

I would vote for including the EPOCH as default in Sardana...

PR #1390, opened 2020-08-21
Fix env validation on macro execution

reszelaz:
When integrating a set of _scan stats_ macros I changed the implementation proposed by @dschick (manual check of the environment variables) to use the environment validation that comes with the MacroServer (`Macro.env` class member).
This revealed a bug in this last feature - the validation does not consider the environment variables defined on the door's or macro's levels. This PR fixes it. @sardana-org/integrators it would be nice if you could take a look on it. If not, I will auto-merge it - we need it for the release.


reszelaz:
Thanks for the review!

PR #1398, opened 2020-08-27
Use proper python3 executable regardeless of installation

reszelaz:
On some OS Python 3 is executed with python3 but in conda environments it is executed with python. Use proper Python 3 executable regardless of the Python installation.

@sardana-org/integrators could you please take a look. IMO it is release critical.
Many thanks!


teresanunez (approved):
I can not test it in different OS but it looks like reasonable to me.

reszelaz:
Thanks for the review @teresanunez!

PR #1400, opened 2020-08-27
Fix showscan offline broken in #1260

reszelaz:
Showscan without arguments should open showscan offline. When scan ID
is given as argument it should open the corresponding scan.

### DIFF
--- a/src/sardana/spock/magic.py
+++ b/src/sardana/spock/magic.py
@@ -83,7 +83,7 @@ def showscan(self, parameter_s=''):
     """
     params = parameter_s.split()
     door = get_door()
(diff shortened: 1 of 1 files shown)


reszelaz:
@sardana-org/integrators could you please take a look. IMO it is release critical.
Many thanks!

teresanunez (commented):
Hi @reszelaz, showscan doesn't work in my computer but the changes seems fine to me.
Sorry, I can not test it.

reszelaz:
Thanks @teresanunez for your comment.
Since there are no more reviews, I will proceed with the merge. Eventual problems would be discovered in the release manual tests.

PR #1401, opened 2020-08-28
Fix timescan with referable channels

### LINKED ISSUE #1399: SEP#2: timescan does not support 2D experimental channel refs
ref to file is not recognised by timescan and it parses it as nan.

reszelaz:
`timescan` and `TScan` does not handle the referable channels. Subscribe to them to make `tiemscan` compatible with referable channels.

Fix #1399.

@13bscsaamjad could you please review and test this one? Many thanks!


13bscsaamjad:
works good!

reszelaz:
Thanks for the tests! Merging.

PR #1394, opened 2020-08-22
Release jul20

aureocarneiro:
This is a PR to master for the official [Jul20 release](https://github.com/sardana-org/sardana/milestone/7). Final tests (automatic and manual) should be done based on it before accepting the merge.

So, this is a call for help from @sardana-org/integrators and any other person willing to help to test this branch ASAP in their respective systems.

For the moment the following systems are going to be tested thoroughly (feel free to add more to the list if you are willing to do the tests):

- [x] Debian 9 (#1392)
- [x] CentOS 7 (#1393)
- [x] Windows 10 (#1391)

PR #1412, opened 2020-09-14
parse dial position to nan in wa macro if motor position is None

### LINKED ISSUE #1411: wa macro does not work if any of the motors report None position
If any motor in the Pool reports None position, `wa` macro does not work, it throws the following exception.

```python
An error occurred while running wa:
TypeError: must be real number, not NoneType
```

13bscsaamjad:
fixes #1411.

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -101,6 +101,8 @@ class _wm(Macro):
                             value = float('NaN')
                             if attr.name == 'dialposition':
                                 value = motor.getDialPosition()
+                                if value is None:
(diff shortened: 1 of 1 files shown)


reszelaz:
@sardana-org/integrators could you please take a look on this one?
@aureofreitas do you think it would be better to open thin PR to the release branch directly?
Thanks to all!

aureocarneiro:
@reszelaz yes I think so, otherwise at MAX IV we will must to apply a patch in our sardana version. @13bscsaamjad can you change the PR to #1394

aureocarneiro (approved):

PR #1413, opened 2020-09-14
fix FWHM calculation

dschick:
small fix for the `scanStats` macro to correctly calc the FWHM for badly sampled data


reszelaz:
I have tested it with my gaussian dummy counter that I used when reviewing the original PR and it works in this case, so I'm merging it.

BTW, in the future it would be nice to introduce automatic tests with some functions' data, what do you think? I could help with programming the tests under you guidance, if you want.

dschick:
oh, I feared already, that you would asked me that ;)
of course I can try doing that. I am already familiar with pytest, so I think I should be able to propose some test files

PR #1414, opened 2020-09-14
Fix upgrade_env.py to properly migrate environments from the Sardana servers

reszelaz:
Do not hard-code the DS name to MacroServer. The upgrade_env.py script should be able to migrate also the Sardana DS environments.

@sardana-org/integrators could you please take a look? It is release critical.


teresanunez (commented):
I have not the possibility to test it now, but it looks fine to me.

reszelaz:
Thanks @teresanunez for looking into it! I'm proceeding with the merge.

PR #1416, opened 2020-09-16
Reset MGConfiguration tango internals also if changes proceeds from the same object

### LINKED ISSUE #1415: showscan online does not respect the set_meas_conf macro
What I have observed is that when changing the measurement group cofiguration using `set_meas_conf` macro - disabling the plotting for a channel, the previously opened showscan online widget was not removing the disabled channel curve in the next scans.

In order to reproduce:
1. Create measurement group with two channels.
2. Enable plotting for these two channels.
3. Start `showscan online`
4. Execute a scan
5. Disable plotting for one of the channels e.g. `set_meas_conf plottype No ct01`
6. Re-run the scan

The curve for the channel is still plotted. This does not happen when one use the expconf to disable plotting.

@rhomspuron and @tiagocoutinho could you please take a look on it? It would be nice to have it for the release (both showscan online and measuremeng group configuration macros are being added in this release). Many thanks!

reszelaz:
This PR fixes #1415. What happens there is that the the same `MeasurementGroup` sardana-taurus model extension object is present in the MacroServer when one runs the `set_meas_conf` macro and when one runs a scan.
The `MGConfiguration.set_data()` is trying to optimize the objects recreation and was doing that also for Tango objects dictionaries. This dictionaries however apart from storing the Tango DeviceProxies also store the configuration info. If we don't reset these dictionaries then the configuration info gets outdated. This was affecting the

https://github.com/sardana-org/sardana/blob/ffbe5f36e6a11a2c2eb4cfd59e6f39dd1185c2de/src/sardana/macroserver/scan/gscan.py#L669

what in consecuence was delivering outdated `data_desc` in the RecordData event.

Note, my patch is very trivial, further refactoring could be done after the release. @rhomspuron, what do you think?

@sardana-org/integrators it is read for review. Thanks to all!

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/pool.py
+++ b/src/sardana/taurus/core/tango/sardana/pool.py
@@ -1499,8 +1499,6 @@ class MGConfiguration(object):
         # representing channel data as received in raw data
         self.non_tango_channels = None
 
-        self.initialized = False
-
     def _build(self):
         # internal channel structure that groups channels by tango device so
         # they can be read as a group minimizing this way the network requests
--- a/src/sardana/taurus/core/tango/sardana/pool.py
+++ b/src/sardana/taurus/core/tango/sardana/pool.py
@@ -1393,6 +1393,34 @@ class MGConfiguration(object):
         self.set_data(data)
 
     def set_data(self, data, force=False):
+        # dict<str, list[DeviceProxy, CaselessDict<str, dict>]>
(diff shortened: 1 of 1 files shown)


teresanunez (commented):
It looks fine to me

reszelaz:
Thanks Teresa for looking into it! I'm merging.

PR #1422, opened 2020-09-28
Fix macro function default settings

tiagocoutinho:
Make sure we return the exact same default settings as the Macro class
(ex: tools like the sequencer GUI expect hints to be a dict)

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -391,13 +391,24 @@ class macro(object):
         def where_moveable(self, moveable):
             self.output("Moveable %s is at %s", moveable.getName(), moveable.getPosition())"""
 
+    param_def = []
+    result_def = []
+    env = ()
+    hints = {}
+    interactive = False
+
     def __init__(self, param_def=None, result_def=None, env=None, hints=None,
                  interactive=None):
-        self.param_def = param_def
-        self.result_def = result_def
-        self.env = env
-        self.hints = hints
-        self.interactive = interactive
+        if param_def is not None:
+            self.param_def = param_def
+        if result_def is not None:
+            self.result_def = result_def
+        if env is not None:
+            self.env = env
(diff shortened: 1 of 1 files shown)


reszelaz:
@tiagocoutinho, Many thanks for the PR! Could you please provide a macro example to reproduce this issue?

tiagocoutinho:
Hi,
It should be reproducible when you have a simple macro function
```python
from sardana.macroserver.macro import macro

@macro()
def test(self):
    self.output("test")
```

and you try to launch it from the *sequencer* GUI application.

When we talked you said you cannot reproduce it. Could it be that the GUI became more robust and is now able to handle `hints=None` ?

IMHO, even if the GUI became more robust, I think is more correct that equivalent Macro class and macro function return the same macro information concerning the hints, param_def, result_def and env since there might be other clients that expect a more consistent API

tiagocoutinho:
FYI, I notice that at ALBA MSPD beamline I had to apply this patch.
I cannot test now to go back to the previous version because they are running but I could do it on the Christmas shutdown. Hope it is not too late for you.

reszelaz:
> When we talked you said you cannot reproduce it. Could it be that the GUI became more robust and is now able to handle `hints=None` ?

The sequencer GUI has not changed in this sense.
I managed to reproduce it! Before I was already using the macro function like you posted above, but I was adding it manually to the sequencer and this code was protected in the GUI.
Loading from the plain text file with the spock syntax (I think they use it a lot at BL04) was not protected and causing the error.
With your PR it gets fixed.

> 
> IMHO, even if the GUI became more robust, I think is more correct that equivalent Macro class and macro function return the same macro information concerning the hints, param_def, result_def and env since there might be other clients that expect a more consistent API

I agree with you, but I wanted to understand well the user scenario before merging the PR. Now I proceed with the merge.
Thanks a lot!

PR #1424, opened 2020-09-29
Fix double unit shown in motor widget

tiagocoutinho:
When the motor position has units defined, the taurus motor widget shows the units in two places as can be seen here:
image

This PR proposes to fix it and only show on the right side of the GUI as it was before


reszelaz (approved):
Small change, but a nice improvement to the user interface. Tested and works as expected. Many thanks! Merging...

PR #1427, opened 2020-10-14
Allow to programmatically disable *deterministic scan* optimization

### LINKED ISSUE #1426: Acquisition in pre-acq hook fails due to MeasurementGroup not prepared after updating to Sardana 3.0.3
Hi!
I was using the "hack" described in #812 to duplicate my counters after toggling a motor at every scan point (magnetic field for XMCD measurements).
Therefore a second acquisition at every point of the scan is implemented as a 'pre-acq' hook in the following way:

```
if altOn:
        magnet.move(-1*ampl)
        magnetState.magnet = -1*ampl
        
        parent = self.getParentMacro()
        if parent:
            integ_time  = parent.integ_time
            mnt_grp     = self.getObj(self.getEnv('ActiveMntGrp'), type_class=Type.MeasurementGroup)
            state, data = mnt_grp.count(integ_time)
                       
        magnet.move(+1*ampl)
        magnetState.magnet = +1*ampl
```

After updating to Sardana 3.0.3, running a scan now fails with the following error:

> An error occurred while running ascan:
PyDs_PythonError: RuntimeError: prepare is mandatory before starting acquisition

I understand that it's now necessary to explicitly prepare the measurement group before acquisition. However, simply calling `mnt_grp.prepare()` before the `mnt_grp.count(integ_time)` in my example does not help and the acquisition within the hook still throws the same error. Is there anything else I have to do?

reszelaz:
Deterministic scan optimization of measurement is not compatible with
attaching measurements on the same measurement group as hooks.

Add deterministic_scan property to allow disabling this optimization.

You can try it with the following example:

```python
import functools

from sardana.macroserver.macro import macro, Type
from sardana.macroserver.scan import SScan


def count(parent):
    parent.output("In count")
    mnt_grp = parent.getObj(parent.getEnv("ActiveMntGrp"),
                            type_class=Type.MeasurementGroup)
    state, data = mnt_grp.count(0.2)
    parent.output(data)


@macro([["moveable", Type.Moveable, None, "moveable to get position"]])
def custom_scan(self, moveable):
    """Testing custom_scan macro"""
    self.output("In custom_scan")
    ret = self.createMacro("ascan", moveable, "0", "1", "1", "0.1")
    my_scan, _ = ret
    hook_function = functools.partial(count, self)
    my_scan.appendHook((hook_function, ["pre-acq"]))
    my_scan._gScan.deterministic_scan = False
    self.runMacro(my_scan)
```
Fixes #1426

### DIFF
--- a/doc/source/devel/api/sardana/macroserver/scan.rst
+++ b/doc/source/devel/api/sardana/macroserver/scan.rst
@@ -26,8 +26,8 @@ GScan
     :show-inheritance:
     :members:
 
-GScan
------
+Scan
+----
 
 .. inheritance-diagram:: SScan
     :parts: 1
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1067,6 +1067,15 @@ class SScan(GScan):
 
     @property
     def deterministic_scan(self):
+        """Check if the scan is a deterministic scan.
+
+        Scan is considered as deterministic scan if
+        the `~sardana.macroserver.macro.Macro` specialization owning
+        the scan object contains ``nb_points`` and ``integ_time`` attributes.
+
+        Scan flow depends on this property (some optimizations are applied).
+        These can be disabled by setting this property to `False`.
+        """
         if self._deterministic_scan is None:
             macro = self.macro
             if hasattr(macro, "nb_points") and hasattr(macro, "integ_time"):
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1058,23 +1058,42 @@ class GScan(Logger):
 class SScan(GScan):
     """Step scan"""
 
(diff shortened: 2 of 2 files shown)


tox1c90:
I can confirm that using this programmatic API fixes the problem described in #1426 for us. I was also able to disable the deterministic_scan optimization on a case by case basis via a pre-scan general hook. So it's possible to continue using the standard scan macros even when the optimization is interfering with our measurement routine.

reszelaz:
@sardana-org/integrators could you please also take a look on this one? Thanks a lot!

teresanunez (approved):

teresanunez:
I could not test it in a real case, but I plan to do that. The changes looks fine and I trust the checks done by @tox1c90 in his
real case.

reszelaz:
@teresanunez many thanks for the review! Now, I just added a docsting to this property.

I will proceed to the merge. If you want to test it further, please comment afterwards on this issue, or in case you find problems open a new one.

PR #1428, opened 2020-10-15
Fix delay description in position domain

jordiandreu:
Adds the missing delay parameter in position domain in the CTScan.


reszelaz (approved):
Tested and works correctly. Many thanks for your contribution! Merging.

Issue #1433, opened 2020-11-08
Travis is discontinuing the unlimited CI service for OSS projects


cpascual:
In case you didn't know, have a look at https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

More specifically:

> For those of you who have been building on public repositories (on travis-ci.com, with no paid subscription), we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).
(...)
When your credit allotment runs out - we’d love for you to consider which of our plans will meet your needs.
We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment.

As a guide, consider that a typical sardana run uses ~25' (so the "trial" allotment will only last ~40builds)

Fortunately, there are good alternatives:
- move to github actions
- move to gitlab-ci

See:
- https://github.com/taurus-org/taurus/issues/1139  (regarding github actions)
- https://github.com/taurus-org/taurus/issues/761 (regarding gitlab-ci)
- https://github.com/taurus-org/taurus_pyqtgraph/pull/60 (regarding an already completed migration to github actions)

reszelaz:
Thanks for sharing - I was not aware of this.
For the moment GH actions wins as an alternative - we already use them for publishing on PyPI:)

reszelaz:
For the moment I'm closing this one. Whenever we run out minutes necessary for building the docs on travis we will come back to this issue.

PR #1434, opened 2020-11-09
Move testsuite from travis-ci to GH actions

reszelaz:
~For the moment just a _naive_ try to run a job on docker container.~

See https://github.com/sardana-org/sardana/issues/1433 for the rationale of the migration.


reszelaz:
For the moment fails with:
> a step cannot have both the `uses` and `run` keys

I will try with moving the run contents to entrypoint (still as a naive approach).

reszelaz:
Yesterday I tried some more things to run jobs inside a container:
1. [docker-run-action](https://github.com/marketplace/actions/docker-run-action)
2. [`job.<job_id>.container`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer)

I managed to run testsuite using the second option.

If we finally go for the GH actions, there are two more things to do:
1. Migrate documentation build and deploy to GH pages. 
2. Flake8 test.

First one should be trivial - we also use the sardana-test container to build the docs and then we could use one of many already existing actions to upload to ghpages e.g. [deploy-to-github-pages](https://github.com/marketplace/actions/deploy-to-github-pages).

Regarding 2 we have two options:
1. We could evaluate correcting all the flake8 warnings with black and then just run flake8 on the whole code base.
2. We could try to adapt the current `flake8_diff.sh` script to work on github actions.

I think that option 1 is of similar effort as 2 and it definitelly gives us more benefits - we adapt all the code to be complient with PEP8.
Anyway the flake8 test is not a must and we could migrate it at some later stage.

reszelaz:
Since yesterday we decided to post-pone eventual migration to the GitLab I will merge this PR already. This way we will not spend travis minutes on running testsuite which could already run on GitHub actions. I will keep travis however for the documentation part for the moment.

PR #1442, opened 2020-11-11
Fix continuous readout of value references in hardware synchronized acquisition

reszelaz:
@jordiandreu and @rhomspuron found a bug in hardware synchronized by start trigger acquisitions with referable channels.
The controller's `RefOne()` are called only once at the end of the acquisition but should be also called continuously during the acqusition.

This PR fixes it. I've also added tests, both, to reveal the problem and some regression tests for the `ReadOne()` calls which curently work fine.

### DIFF
--- a/src/sardana/pool/poolacquisition.py
+++ b/src/sardana/pool/poolacquisition.py
@@ -1081,6 +1081,9 @@ class PoolAcquisitionHardware(PoolAcquisitionTimerable):
                 self.read_value(ret=values)
                 for acquirable, value in list(values.items()):
                     self._process_value_buffer(acquirable, value)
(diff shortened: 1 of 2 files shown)


reszelaz:
It is green already. @jordiandreu could you please give it a try?

jordiandreu:
I have already tried this PR in python2 branch and it works fine.
@reszelaz many thanks fore the work.

aureocarneiro:
Hi tested now, I think its ready to merge in develop branch :)

aureocarneiro (approved):

reszelaz:
Thanks @jordiandreu and @aureocarneiro for the tests and the review!

PR #1446, opened 2020-11-23
Leave Tango device class attributes definition untouched

### LINKED ISSUE #1440: 1D or 2D pseudo counter cannot be created with more than one pseudo counter role / axis
I try to create 1D and 2D pseudo counter controllers which integrate multiple different ROIs along one axis of an image or treating the whole image differently depending on the axis of the pseudo counter.

In case of a 0D pseudo counter, this works perfectly by defining multiple pseudo counter roles and evaluating the axis variable within the calculation function.

However, when I try to define more than one pseudo counter roles for a 1D or 2D controller, I get an error. See the following simple example which just multiplies the image by the number of the axis:

```
class dummy2DMultipleAxisController(PseudoCounterController):

    counter_roles        = ('I',)
    pseudo_counter_roles = ('O1','O2','O3','O4',)
    
    def __init__(self, inst, props, *args, **kwargs):
        PseudoCounterController.__init__(self,inst,props, *args, **kwargs)
        
        axis_attrs = PseudoCounterController.GetAxisAttributes(self, axis)
        axis_attrs = dict(axis_attrs)
        axis_attrs['Value'][Type] = ((float, ), )
        axis_attrs['Value'][MaxDimSize] = (1024,1024)
        return axis_attrs
    
    def Calc(self, axis, counters):
        image = counters[0]        
        return axis*image
```

Doing `defctrl dummy2DMultipleAxisController dummy2DMultipleAxisCtrl I=dummy2D O1=dummy2Dx1 O2=dummy2Dx2 O3=dummy2Dx3 O4=dummy2Dx4` with a `dummy2D` counter as input, throws an error:

`An error occurred while running defctrl:
API_AttrNotFound: value attribute not found`

Only the first pseudo counter dummy2Dx1 is created properly. The second one is created but as a scalar, and trying to add it to the measurement group leads to `Attribute Value is not supported by device pc/dummy2dmultipleaxisctrl/2`.

Am I doing something wrong here, or is this a bug?

reszelaz:
When adding attributes the controller code can modify them e.g. format, type, maxdimsize. Make a copy of the standard class attributes proceeding from the class definition to avoid problems with modifying them.

Fixes #1440.

### DIFF
--- a/src/sardana/tango/pool/PoolDevice.py
+++ b/src/sardana/tango/pool/PoolDevice.py
@@ -33,6 +33,7 @@ __all__ = ["PoolDevice", "PoolDeviceClass",
 __docformat__ = 'restructuredtext'
 
 import time
+from copy import deepcopy
 
 from PyTango import Util, DevVoid, DevLong64, DevBoolean, DevString,\
     DevDouble, DevEncoded, DevVarStringArray, DispLevel, DevState, SCALAR, \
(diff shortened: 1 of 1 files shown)


tox1c90:
I can confirm that it's fixed by this commit.

reszelaz:
@sardana-org/integrators could you please review this change? The CI fails due to the problems with the flake8 check which was already removed in #1455.
Many thanks!

teresanunez (approved):

cpt-majkel (approved):

reszelaz:
Thanks @teresanunez and @cpt-majkel for the review. Merging..

PR #1461, opened 2020-12-18
Document how to properly deal with exceptions in macros

reszelaz:
In our control room our operators started using macros for the accelerators operation. They migrated a set of python scripts into macros. We have observed that in their scripts (now macros) they were using broad exception types in the `except` clause.
This may interfer with the macro stopping/aborting. 484af30 warns about it.

Since the [How to develop macros](https://sardana-controls.org/devel/howto_macros/macros_general.html) documentation already growed to a big chapter I take a profit of this PR to add a table of contats (TOC) in order to faciltate navigation. The section headers are now links to the chapter in the TOC.
It looks more or less like this:

image

@sardana-org/integrators could you please take a look on this. Thanks a lot!

### DIFF
--- a/doc/source/devel/api/api_macro.rst
+++ b/doc/source/devel/api/api_macro.rst
@@ -42,3 +42,23 @@ imacro decorator
     :members:
     :undoc-members:
 
+StopException
+-------------
+
+.. autoclass:: StopException
+    :members:
+    :undoc-members:
+
+AbortException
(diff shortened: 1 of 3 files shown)


teresanunez (approved):

reszelaz:
Thanks @teresanunez for your fast review. I'm merging.

PR #1463, opened 2020-12-21
Allow running Spock without Qt bindings

### LINKED ISSUE #1462: [BUG] Spock crashes without Qt
When running Spock in environment without Qt bindings installed, it fails with message `ImportError: No Qt bindings could be imported`. 
This error message is correct but behaviour is actual bug. 

Spock is build on IPython which can be launched from command line without Qt bindings. Spock should be able to run with the same environment (without Qt bindings) without issue. Any integration with Qt should be an opt-in mechanism.

Inspecting source code, there are lot of places where existence of Qt bindings along with Spock is incorrectly assumed. On the bright side, it looks like some minor code refactoring would fix this bug.
I've found following modules to be source of this bug:
 - `sardana.spock.ipython_01_00.genutils`
 - `sardana.spock.spockms`
 - `sardana.spock.inputhandler`

stanislaw55:
Allow launching Spock from commandline without need to have Qt bindings installed.
The behaviour is like this:
 - try to detect if Qt bindings are installed (vi `import taurus.external.qt.Qt`)
 - if Qt bindings are installed, use them
 - if Qt bindings are not installed, use command-line only input handler.

closes #1462 

I'm not an expert in IPython or Spock hence I would appreciate testing from more advanced users.

### DIFF
--- a/src/sardana/spock/ipython_01_00/genutils.py
+++ b/src/sardana/spock/ipython_01_00/genutils.py
@@ -1289,12 +1289,15 @@ def mainloop(app=None, user_ns=None):
 def prepare_input_handler():
     # initialize input handler as soon as possible
 
-    try:
-        import sardana.spock.qtinputhandler
-        _ = sardana.spock.inputhandler.InputHandler()
-    except ImportError:
-        import sardana.spock.inputhandler
-        _ = sardana.spock.inputhandler.SpockInputHandler()
+    from sardana import sardanacustomsettings
+
+    if sardanacustomsettings.SPOCK_INPUT_HANDLER == "Qt":
+
+        try:
+            import sardana.spock.qtinputhandler
+            _ = sardana.spock.qtinputhandler.InputHandler()
+        except ImportError:
+            raise Exception("Cannot use Spock Qt input handler!")
 
 
 def prepare_cmdline(argv=None):
--- a/src/sardana/spock/ipython_01_00/genutils.py
+++ b/src/sardana/spock/ipython_01_00/genutils.py
@@ -82,7 +82,10 @@ from taurus.core.taurushelper import Factory
 from taurus.core.util.codecs import CodecFactory
 
 # make sure Qt is properly initialized
-from taurus.external.qt import Qt
+try:
+    from taurus.external.qt import Qt
+except ImportError:
+    pass
 
 from sardana.spock import exception
 from sardana.spock import colors
@@ -110,7 +113,11 @@ ENV_NAME = "_E"
 
 
 def get_gui_mode():
-    return 'qt'
+    try:
+        import taurus.external.qt.Qt
+        return 'qt'
+    except ImportError:
+        return None
 
 
 def get_pylab_mode():
@@ -1159,7 +1166,8 @@ object?   -> Details about 'object'. ?object also works, ?? prints more.
     term_app = config.TerminalIPythonApp
     term_app.display_banner = True
     term_app.gui = gui_mode
-    term_app.pylab = 'qt'
+    if gui_mode == 'qt':
+        term_app.pylab = 'qt'
     term_app.pylab_import_all = False
     #term_app.nosep = False
     #term_app.classic = True
@@ -1280,8 +1288,13 @@ def mainloop(app=None, user_ns=None):
 
 def prepare_input_handler():
     # initialize input handler as soon as possible
-    import sardana.spock.inputhandler
-    _ = sardana.spock.inputhandler.InputHandler()
+
+    try:
+        import sardana.spock.qtinputhandler
+        _ = sardana.spock.inputhandler.InputHandler()
+    except ImportError:
+        import sardana.spock.inputhandler
+        _ = sardana.spock.inputhandler.SpockInputHandler()
 
 
 def prepare_cmdline(argv=None):
--- a/src/sardana/spock/magic.py
+++ b/src/sardana/spock/magic.py
@@ -42,7 +42,7 @@ def expconf(self, parameter_s=''):
     for the experiments (scans)"""
 
     try:
-        from taurus.external.qt import qt
+        from taurus.external.qt import Qt
     except ImportError:
         print("Qt binding is not available. ExpConf cannot work without it."
               "(hint: maybe you want to use experiment configuration macros? "
@@ -92,7 +92,7 @@ def showscan(self, parameter_s=''):
     """
 
     try:
-        from taurus.external.qt import qt
+        from taurus.external.qt import Qt
     except ImportError:
         print("Qt binding is not available. Showscan cannot work without it.")
         return
@@ -138,7 +138,7 @@ def showscan(self, parameter_s=''):
 
 def spsplot(self, parameter_s=''):
     try:
-        from taurus.external.qt import qt
+        from taurus.external.qt import Qt
     except ImportError:
         print("Qt binding is not available. SPSplot cannot work without it.")
         return
--- a/src/sardana/spock/magic.py
+++ b/src/sardana/spock/magic.py
@@ -40,6 +40,15 @@ from .genutils import page, get_door, get_macro_server, ask_yes_no, arg_split
 def expconf(self, parameter_s=''):
     """Launches a GUI for configuring the environment variables
     for the experiments (scans)"""
+
+    try:
+        from taurus.external.qt import qt
+    except ImportError:
+        print("Qt binding is not available. ExpConf cannot work without it."
+              "(hint: maybe you want to use experiment configuration macros? "
+              "https://sardana-controls.org/users/standard_macro_catalog.html#experiment-configuration-macros)")
+        return
+
     try:
         from sardana.taurus.qt.qtgui.extra_sardana import ExpDescriptionEditor
     except:
@@ -81,6 +90,13 @@ def showscan(self, parameter_s=''):
     Where *online* means plot the scan as it runs and *offline* means -
     extract the scan data from the file - works only with HDF5 files.
     """
+
+    try:
+        from taurus.external.qt import qt
+    except ImportError:
+        print("Qt binding is not available. Showscan cannot work without it.")
+        return
+
     params = parameter_s.split()
     door = get_door()
     scan_nb = None
@@ -121,6 +137,12 @@ def showscan(self, parameter_s=''):
 
 
 def spsplot(self, parameter_s=''):
+    try:
+        from taurus.external.qt import qt
+    except ImportError:
+        print("Qt binding is not available. SPSplot cannot work without it.")
+        return
+
     get_door().plot()
 
 
--- a/src/sardana/spock/spockms.py
+++ b/src/sardana/spock/spockms.py
@@ -44,7 +44,6 @@ CHANGE_EVTS = TaurusEventType.Change, TaurusEventType.Periodic
 
 
 if genutils.get_gui_mode() == 'qt':
-    from taurus.external.qt import Qt
     from sardana.taurus.qt.qtcore.tango.sardana.macroserver import QDoor, QMacroServer
     BaseDoor = QDoor
     BaseMacroServer = QMacroServer
--- a/src/sardana/spock/spockms.py
+++ b/src/sardana/spock/spockms.py
@@ -46,13 +46,11 @@ CHANGE_EVTS = TaurusEventType.Change, TaurusEventType.Periodic
 if genutils.get_gui_mode() == 'qt':
     from taurus.external.qt import Qt
     from sardana.taurus.qt.qtcore.tango.sardana.macroserver import QDoor, QMacroServer
-    from sardana.spock.qtinputhandler import InputHandler
     BaseDoor = QDoor
     BaseMacroServer = QMacroServer
     BaseGUIViewer = object
 else:
     from sardana.taurus.core.tango.sardana.macroserver import BaseDoor, BaseMacroServer
-    from sardana.spock.inputhandler import SpockInputHandler
     BaseGUIViewer = object
 
 
@@ -292,6 +290,8 @@ class SpockBaseDoor(BaseDoor):
         self.call__init__(BaseDoor, name, **kw)
 
     def create_input_handler(self):
+        from sardana.spock.inputhandler import SpockInputHandler
+
         return SpockInputHandler()
 
     def get_color_mode(self):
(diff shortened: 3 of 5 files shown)


guifrecuni:
Hi, just one small comment, I think that there are some other tools that rely on Qt to make spock work as expected (e.g. experiment config tool, or online scan plotting).
Could we consider that spock "not just ipython with magics" but some application which depends on Qt?

stanislaw55:
Hi @guifrecuni,
one can use Spock to just run macros. There's no obligation to use showscan or expconf. Current code base enforces user to install Qt bindings to just launch Spock, regardless of user intent - if someone just wants to run some simple macros, that's big overhead. If there's need to use expconf or showscan or something else, nothing changes - Qt bindings are still required.

I tried to make changes as much compatible with current situation as I could but testing from others is highly appreciated.

guifrecuni:
you are right, Qt is not needed to run (some) macros, because plotting is also an option available to macros:
https://sardana-controls.org/devel/examples/macro_plotting_examples.html
For me it is fine either way, maybe spock should warn the user that if no Qt is found several enumerated features (1. xxxx, 2. xxxx, ...) will not be available

src/sardana/spock/spockms.py
```diff
@@ -38,19 +38,21 @@
...
 if genutils.get_gui_mode() == 'qt':
+    from taurus.external.qt import Qt
     from sardana.taurus.qt.qtcore.tango.sardana.macroserver import QDoor, QMacroServer
+    from sardana.spock.qtinputhandler import InputHandler
     BaseDoor = QDoor
     BaseMacroServer = QMacroServer
     BaseGUIViewer = object
 else:
     from sardana.taurus.core.tango.sardana.macroserver import BaseDoor, BaseMacroServer
+    from sardana.spock.inputhandler import SpockInputHandler
```
reszelaz, line 55: We can not import it only here. The default behavior (see `sardanacustomsettings.SPOCK_INPUT_HANDLER` uses the `SpockInputHandler` even if the Qt is installed. So, `genutils.get_gui_mode()` is not enough to determine which input handler to import. I propose to move both imports to the `create_input_handler()` methods of the _doors_.

src/sardana/spock/spockms.py
```diff
@@ -290,7 +292,7 @@ def __init__(self, name, **kw):
         self.call__init__(BaseDoor, name, **kw)
 
     def create_input_handler(self):
-        return SpockInputHandler(self)
+        return SpockInputHandler()
```
reszelaz, line 295: Good spotting!

src/sardana/spock/spockms.py
```diff
@@ -38,19 +38,21 @@
...
 from sardana.spock import genutils
 from sardana.util.parser import ParamParser
-from sardana.spock.inputhandler import SpockInputHandler, InputHandler
 from sardana import sardanacustomsettings
 
 CHANGE_EVTS = TaurusEventType.Change, TaurusEventType.Periodic
 
 
 if genutils.get_gui_mode() == 'qt':
+    from taurus.external.qt import Qt
```
reszelaz, line 47: I think this import is not even necessary.

reszelaz (changes requested):
Nice work @stanislaw55! I just have some minor comments and suggest some changes. I already have them developed and will proceed to push to your branch.

reszelaz:
Spock features not available when there is no Qt:
* showscan (offline)
* plotting from macros
* spsplot (I never tried it)
* interactive macro dialogs in GUI mode

We could think on warning the user as suggest @guifrecuni.

src/sardana/spock/spockms.py
stanislaw55: Thanks - I caught it when I was testing it.

src/sardana/spock/spockms.py
stanislaw55: This import was put somewhere else in the same file with comment that Qt needs to be imported ASAP. I figured out that it would good place to keep this import in case it is required. However, I can delete it if you want.

stanislaw55:
Hi @reszelaz 
Thanks for review. Regarding warnings about unavailable Qt-related Spock features I have question about approach. Should it be one-time warning when launching Spock with list of those features? Or maybe should I modify code of those features to raise a warning when there's no Qt and they are used? Or maybe some different approach?

src/sardana/spock/spockms.py
stanislaw55: From my code analysis, current solution does not use `SpockInputHandler` at all because of two reasons:
 - `genutils.get_gui_mode()` always returns `qt` which makes `if` condition redundant
 - there was an error in when instantiating `SpockInputHanlder` in `create_input_hanlder` method. In current version, there was no way to get it to work AFAIK.

This was my analysis - I may be wrong, correct me please if I am.

I'm not entirely sure that I know what you mean so I'll push changes and you will be able to see if that's what you meant.

stanislaw55: Regarding imports - only import of input handlers, right?

src/sardana/spock/ipython_01_00/genutils.py
```diff
@@ -1280,8 +1288,13 @@ def mainloop(app=None, user_ns=None):
...
 def prepare_input_handler():
     # initialize input handler as soon as possible
-    import sardana.spock.inputhandler
-    _ = sardana.spock.inputhandler.InputHandler()
+
+    try:
+        import sardana.spock.qtinputhandler
+        _ = sardana.spock.inputhandler.InputHandler()
+    except ImportError:
+        import sardana.spock.inputhandler
```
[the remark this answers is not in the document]
stanislaw55: I've changed it

src/sardana/spock/spockms.py
reszelaz: > From my code analysis, current solution does not use `SpockInputHandler` at all because of two reasons:

The `SpockInputHandler` was instantiated correctly in the [`QSpockDoor.create_input_handler()`](https://github.com/sardana-org/sardana/blob/f2f73ac20b1f397450109a582e264577cfabd011/src/sardana/spock/spockms.py#L584).
This is a typical scenario - Qt is installed and `sardanacustomsetting.SPOCK_INPUT_HANDLER = "CLI".

You are right that the creation on `SpockInputHandler` in the [`SpockBaseDoor.create_input_handler()`](https://github.com/sardana-org/sardana/blob/f2f73ac20b1f397450109a582e264577cfabd011/src/sardana/spock/spockms.py#L293) was erroneus and you fixed it. But this scenario was rather not used.

src/sardana/spock/spockms.py
reszelaz: This import is currently here:

https://github.com/sardana-org/sardana/blob/f2f73ac20b1f397450109a582e264577cfabd011/src/sardana/spock/spockms.py#L557

and I don't see any comment around it explaining the urgency of the import.

From where you placed it, the subsequent line also imports Qt:
```python
from sardana.taurus.qt.qtcore.tango.sardana.macroserver import QDoor, QMacroServer
```

So, let's try to avoid the import. I hope it won't break anything:)

reszelaz:
> Regarding warnings about unavailable Qt-related Spock features I have question about approach. Should it be one-time warning when launching Spock with list of those features? Or maybe should I modify code of those features to raise a warning when there's no Qt and they are used? Or maybe some different approach?

I think the ideal solution would be to warn when trying to use such a feature. 
When previously listening the features depending on Qt, of course, I forgot to mention: expconf and showscan online (the most obvious ones).

So the complete list of features is:
1. expconf
2. showscan online
3. showscan (offline)
4. plotting from macros
5. spsplot (I never tried it)
6. interactive macro dialogs in GUI mode

Regarding the first three and 5. this could be at the beginning of the magic command.
Regarding the 4. I think there is already some protection implemented. So, I would start from testing on how it works now.
Regarding the 6. we already raise an exception when one selected the `InputHandler` and Qt is not installed.
Also when warning in expconf we could suggest using the [experiment configuration macros](https://sardana-controls.org/users/standard_macro_catalog.html#experiment-configuration-macros).

src/sardana/spock/ipython_01_00/genutils.py
```diff
@@ -1280,8 +1288,16 @@ def mainloop(app=None, user_ns=None):
 
 def prepare_input_handler():
     # initialize input handler as soon as possible
-    import sardana.spock.inputhandler
-    _ = sardana.spock.inputhandler.InputHandler()
+
+    from sardana import sardanacustomsettings
+
+    if sardanacustomsettings.SPOCK_INPUT_HANDLER == "Qt":
```
reszelaz, line 1294:
```suggestion
    if getattr(sardanacustomsettings, "SPOCK_INPUT_HANDLER", "CLI") == "Qt":
```

reszelaz: This is more robust in case someone removed the setting. We try to access to the `sardanacustomsettings` in this way.

src/sardana/spock/ipython_01_00/genutils.py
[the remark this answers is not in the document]
reszelaz: Thanks, I have made a suggestion on how to make access to the `sardanacustomsettings` more robust.

src/sardana/spock/spockms.py
stanislaw55: Okay, will do

src/sardana/spock/spockms.py
stanislaw55: Okay, right, I see it now.

stanislaw55:
I've added printing warning messages and other changes @reszelaz

src/sardana/spock/magic.py
```diff
@@ -40,6 +40,15 @@
 def expconf(self, parameter_s=''):
     """Launches a GUI for configuring the environment variables
     for the experiments (scans)"""
+
+    try:
+        from taurus.external.qt import qt
```
reszelaz, line 45:
```suggestion
        from taurus.external.qt import Qt
```

src/sardana/spock/magic.py
```diff
@@ -81,6 +90,13 @@ def showscan(self, parameter_s=''):
     Where *online* means plot the scan as it runs and *offline* means -
     extract the scan data from the file - works only with HDF5 files.
     """
+
+    try:
+        from taurus.external.qt import qt
```
reszelaz, line 95:
```suggestion
        from taurus.external.qt import Qt
```

src/sardana/spock/magic.py
```diff
@@ -121,6 +137,12 @@ def showscan(self, parameter_s=''):
 
 
 def spsplot(self, parameter_s=''):
+    try:
+        from taurus.external.qt import qt
```
reszelaz, line 141:
```suggestion
        from taurus.external.qt import Qt
```

reszelaz (changes requested):

src/sardana/spock/spockms.py
cpascual: I haven't checked this precise case, but usually the "urgency" for importing Qt from `taurus.external.qt` is related to avoiding indirect imports of PyQt4 API1 bindings by other modules  (e.g. pyqtgraph) which would break taurus.

reszelaz (changes requested):
Hi @stanislaw55,
I have performed tests with PyQt5 and without. Looks like everything works.
There is just one minor thing:
In https://github.com/sardana-org/sardana/pull/1463#issuecomment-760788753 I was too optimistic about:
> Regarding the 4. I think there is already some protection implemented. So, I would start from testing on how it works now.

I do not know why, but https://github.com/sardana-org/sardana/blob/c646206b3f532e717be073c4c01e33a6e6358d75/src/sardana/spock/spockms.py#L530
does not catch the error properly and what we get is:
```
Door_zreszela_1 [1]: J0_plot

Door_zreszela_1 [2]:                                                                                                                                                                         
  File "/home/zreszela/.local/bin/spock", line 11, in <module>
    load_entry_point('sardana', 'console_scripts', 'spock')()
  File "/homelocal/zreszela/workspace/sardana/src/sardana/spock/__init__.py", line 34, in main
    run()
  File "/homelocal/zreszela/workspace/sardana/src/sardana/spock/ipython_01_00/genutils.py", line 1373, in run
    launch_new_instance()
  File "/usr/lib/python3/dist-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/usr/lib/python3/dist-packages/IPython/terminal/ipapp.py", line 355, in start
    self.shell.mainloop()
  File "/usr/lib/python3/dist-packages/IPython/terminal/interactiveshell.py", line 495, in mainloop
    self.interact()
  File "/usr/lib/python3/dist-packages/IPython/terminal/interactiveshell.py", line 478, in interact
    code = self.prompt_for_code()
  File "/usr/lib/python3/dist-packages/IPython/terminal/interactiveshell.py", line 372, in prompt_for_code
    pre_run=self.pre_prompt, reset_current_buffer=True)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/interface.py", line 415, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
  File "/usr/lib/python3/dist-packages/prompt_toolkit/eventloop/posix.py", line 102, in run
    self._inputhook_context.call_inputhook(ready)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/eventloop/inputhook.py", line 74, in call_inputhook
    self.inputhook(self)
  File "/usr/lib/python3/dist-packages/IPython/terminal/interactiveshell.py", line 510, in inputhook
    self._inputhook(context)
  File "/usr/lib/python3/dist-packages/IPython/terminal/pt_inputhooks/tk.py", line 93, in inputhook
    wait_using_filehandler()
  File "/usr/lib/python3/dist-packages/IPython/terminal/pt_inputhooks/tk.py", line 70, in wait_using_filehandler
    root.createfilehandler(inputhook_context.fileno(), _tkinter.READABLE, done)
RuntimeError: Calling Tcl from different apartment

If you suspect this is an IPython bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@python.org

You can print a more detailed traceback right now with "%tb", or use "%debug"
to interactively debug it.

Extra-detailed tracebacks for bug-reporting purposes can be enabled via:
    %config Application.verbose_crash=True

Error in atexit._run_exitfuncs:
  File "/usr/lib/python3.7/tkinter/__init__.py", line 775, in after_cancel
    data = self.tk.call('after', 'info', id)
RuntimeError: main thread is not in main loop
```
So, I propose to add an earlier check as in the cases of the magic commands. I've preared directly a PR to your fork: https://github.com/synchrotron-solaris/sardana/pull/1. After solving this issue everything is ready.

stanislaw55:
I've merged your PR @reszelaz. Hope it'll work now.

reszelaz (approved):

reszelaz:
Thanks. Merging.

PR #1465, opened 2020-12-22
Feature/expconf without spyder

### LINKED ISSUE #1464: ExpConf without Spyder
Hi sardana community.
I tried to run expconf widget from Spock and bump into `ImportError` with `spyder`. After inspecting source code, it seems that `spyder` is only used to create actions in menu in `sardana.taurus.qt.qtgui.extra_sardana.sardanaeditor.SardanaEditor`.

Spyder is quite a big dependency (more than 120 indirect dependencies on PyPI) and installs lots of other tools (black, pylint, flake8, sphinx). On the other hand, current usage of spyder seems fairly simple. Would it be possible to somehow replace it with something lighter? Or maybe to make it weak, not hard, dependency?

stanislaw55:
Closes #1464 

This PR provides extra module with utilities from Spyder project that were used in `SardanaEditor`. This has effect of dropping Spyder as hard dependency for ExpConf and other widgets.


cpascual (approved):
I am in favor of removing spyder dependency. In fact I've always been in favor of the whole `SardanaEditor` removal (which would make this unnecessary)

Issue #1296, opened 2020-02-20
change of TwoDExpChannel image/array size


dschick:
Hi all,

we have implemented a TwoDExpChannel based on a Tango CDD Device Server.
(Btw. SEP2 and SEP18 are really great enhancements of the API!)
Currently, we can change the image size of the CCD in the TANGO DS via cropping and binning.

When initially adding the 2D image of the TwoDExpChannel to one of our measurement groups, the size of the 2D array will be automatically recognized.
However, when we change the size of the image array due to cropping or binning the next acquisition via the measurement will result in an error, due to the wrong image size.

Are there ways to autmatically update the size of the image to always the size of the image attribute of the underlying TangoDS?

Best

Daniel

reszelaz:
Hi Daniel,
Sorry for late answer!
We were aware of this lack of user friendliness when working on SEP2. I think we could retake this topic now.
Talking a little bit from memory, I think that this information (shape) is mainly used by the recorders? And in the case of the H5 recorder it is used to define the shape of the datasets.
If this is true, then we could evaluate these two options:

1. We want to allow scans in which experimental channels change the shape of the data during the scan. I personally don't know a use case for that, but maybe someone has? Then the shape could be understood as the maximum shape and should be maintained in the channel configuration (expconf). 
2. We assume that in all scan points the data shape won't change. In this case, we could postpone the dataset creation till the moment of receiving the first value by the recorder. Then we could eliminate the shape from the channel configuration (expconf).

Regarding point 1 maybe there are even some ways of re-shaping datasets a posteriori. Well there is one - we use it for the _scan points_ dimension, but I don't know if we could have a dataset without defining any dimension at all.

dschick:
hi Zibi,

I agree that 2. is the most common scenario. So it would be great if the recorder could adapt to the data source size after the first acquisiton of a scan.

Honestly, we are not chaning the image size too often, but for especially for unexperienced users it is a bit confusing, when an error message in sardana appears, when they change the binning of the detector.

Best

Daniel

reszelaz:
[...]

PR #1466, opened 2020-12-23
Change shape to read only axis parameter

reszelaz:
This PR changes the experimental channel's shape: from the measurement group configuration parameter to the read only controller's axis parameter - see #1296. The main benefit of this is that it is not necessary to modify the measurement group configuration everytime the shape changes e.g. RoI or binning change. The drawbacks is that the shape is consulted on every scan start (for external channels one extra read may be necessary).

@dschick, I have published it even if it is still WIP so you could comment if it goes in the right direction. It should be pretty operable - the scan already takes into account the shape reported by the controller and ignores the shape from the measurement group configuration. If you already go for testing it please follow the example of dummies from e01ce34.

Roadmap:
- [x] remove the shape from the measurement group configuration (server side and expconf)
- [ ] verify if the backwards compatibily asked by @teresanunez in https://github.com/sardana-org/sardana/issues/1296#issuecomment-596383364 works as expected
- [x] add documentation

### DIFF
--- a/doc/source/devel/howto_controllers/howto_1dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_1dcontroller.rst
@@ -30,6 +30,31 @@ To write a 1D controller class you can follow
 the :ref:`sardana-countertimercontroller` guide keeping in mind
 differences explained in continuation.
 
+.. _sardana-1dcontroller-general-guide-shape:
+
+Get 1D shape
+~~~~~~~~~~~~
+
+1D controller must provide a shape of the spectrum which will be produced by
+acquisition. The shape can be either static e.g. defined by the detector's
+sensor size or dynamic e.g. depending on the detector's (or an intermediate
+control software layer e.g. `LImA`_) configuration like :term:`RoI` or binning.
+
+In any case you must provide the shape in the format of a one-element sequence
+with the length of the spectrum using
+the :meth:`~sardana.pool.controller.Controller.GetAxisPar` method.
+
+Here is an example of the possible implementation of
+:meth:`~sardana.pool.controller.Controller.GetAxisPar`:
+
+.. code-block:: python
+
+    class SpringfieldOneDController(TwoDController):
+
+        def GetAxisPar(self, axis, par):
+            if par == "shape":
+                return self.springfield.getShape(axis)
+
 .. _sardana-1dcontroller-differences-countertimer:
 
 Differences with counter/timer controller
--- a/doc/source/devel/howto_controllers/howto_1dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_1dcontroller.rst
@@ -35,12 +35,12 @@ differences explained in continuation.
 Get 1D shape
 ~~~~~~~~~~~~
 
-1D controller must provide a shape of the spectrum which will be produced by
+1D controller should provide a shape of the spectrum which will be produced by
 acquisition. The shape can be either static e.g. defined by the detector's
 sensor size or dynamic e.g. depending on the detector's (or an intermediate
 control software layer e.g. `LImA`_) configuration like :term:`RoI` or binning.
 
-In any case you must provide the shape in the format of a one-element sequence
+In any case you should provide the shape in the format of a one-element sequence
 with the length of the spectrum using
 the :meth:`~sardana.pool.controller.Controller.GetAxisPar` method.
 
@@ -54,6 +54,10 @@ Here is an example of the possible implementation of
         def GetAxisPar(self, axis, par):
             if par == "shape":
                 return self.springfield.getShape(axis)
+
+For backwards compatibility, in case of not implementing the ``shape`` axis
+parameter, shape will be determined from the ``MaxDimSize`` of the ``Value``
+attribute, currently (4096,).
 
 .. _sardana-1dcontroller-differences-countertimer:
 
--- a/doc/source/devel/howto_controllers/howto_1dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_1dcontroller.rst
@@ -121,7 +121,7 @@ leaving the data storage at the responsibility of the detector
 just deals with the reference to the data.
 
 Please refer to :ref:`sardana-2dcontroller-valuereferencing` chapter from
-:ref:`sardana-2dcontroller-howto` in order to implement this feature for 1D
+:ref:`sardana-2dcontroller-howto` guide in order to implement this feature for 1D
 controller.
 
 .. _ALBA: http://www.cells.es/
--- a/doc/source/devel/howto_controllers/howto_2dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_2dcontroller.rst
@@ -173,7 +173,7 @@ Configure 2D value reference
 Two axis parameters: ``value_ref_pattern`` (`str`)
 and ``value_ref_enabled`` (`bool`) are foreseen for configuring where to store
 the values and whether to use the value referencing. Here you need to implement
-the :meth:`~sardana.pool.controller.Controller.SetAxiPar` method.
+the :meth:`~sardana.pool.controller.Controller.SetAxisPar` method.
 
 Here is an example of the possible implementation of
 :meth:`~sardana.pool.controller.Controller.SetAxisPar`:
--- a/doc/source/devel/howto_controllers/howto_2dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_2dcontroller.rst
@@ -27,6 +27,31 @@ To write a 2D controller class you can follow
 the :ref:`sardana-countertimercontroller` guide keeping in mind
 differences explained in continuation.
 
+.. _sardana-2dcontroller-general-guide-shape:
+
+Get 2D shape
+~~~~~~~~~~~~
+
+2D controller must provide a shape of the image which will be produced by
+acquisition. The shape can be either static e.g. defined by the detector's
+sensor size or dynamic e.g. depending on the detector's (or an intermediate
+control software layer e.g. `LImA`_) configuration like :term:`RoI` or binning.
+
+In any case you must provide the shape in the format of a two-element sequence
+with horizonatal and vertical dimensions using
+the :meth:`~sardana.pool.controller.Controller.GetAxisPar` method.
+
+Here is an example of the possible implementation of
+:meth:`~sardana.pool.controller.Controller.GetAxisPar`:
+
+.. code-block:: python
+
+    class SpringfieldTwoDController(TwoDController):
+
+        def GetAxisPar(self, axis, par):
+            if par == "shape":
+                return self.springfield.getShape(axis)
+
 .. _sardana-2dcontroller-differences-countertimer:
 
 Differences with counter/timer controller
--- a/doc/source/devel/howto_controllers/howto_2dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_2dcontroller.rst
@@ -32,12 +32,12 @@ differences explained in continuation.
 Get 2D shape
 ~~~~~~~~~~~~
 
-2D controller must provide a shape of the image which will be produced by
+2D controller should provide a shape of the image which will be produced by
 acquisition. The shape can be either static e.g. defined by the detector's
 sensor size or dynamic e.g. depending on the detector's (or an intermediate
 control software layer e.g. `LImA`_) configuration like :term:`RoI` or binning.
 
-In any case you must provide the shape in the format of a two-element sequence
+In any case you should provide the shape in the format of a two-element sequence
 with horizontal and vertical dimensions using
 the :meth:`~sardana.pool.controller.Controller.GetAxisPar` method.
 
@@ -51,6 +51,10 @@ Here is an example of the possible implementation of
         def GetAxisPar(self, axis, par):
             if par == "shape":
                 return self.springfield.getShape(axis)
+
+For backwards compatibility, in case of not implementing the ``shape`` axis
+parameter, shape will be determined frm the ``MaxDimSize`` of the ``Value``
+attribute, currently (4096, 4096).
 
 .. _sardana-2dcontroller-differences-countertimer:
 
--- a/doc/source/devel/howto_controllers/howto_2dcontroller.rst
+++ b/doc/source/devel/howto_controllers/howto_2dcontroller.rst
@@ -38,7 +38,7 @@ sensor size or dynamic e.g. depending on the detector's (or an intermediate
 control software layer e.g. `LImA`_) configuration like :term:`RoI` or binning.
 
 In any case you must provide the shape in the format of a two-element sequence
-with horizonatal and vertical dimensions using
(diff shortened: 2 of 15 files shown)


dschick:
Hi, we are finally testing the PR.
It seems that it reads the shape of our 2Dcontroller for the first scan correctly but then it is not updating it anymore for the upcoming scans.

A simple `print` command is only shown once in the Sardana terminal after starting the first scan.

It also seems that it tries do determine the shape for every element in the measurement group? at least we get several error messages for the first scan that the shape is None.

dschick:
Hi @reszelaz ,

here are the errors which happen as soon as there is pseudocounter linked to a 2D image
```python
EventTH.W001   ERROR    2021-01-21 23:02:57,790 EventTH.EventTH.W001: Uncaught exception running job 'set_attribute_push' called from thread SardanaTP.W002:
  File "/usr/lib/python3.6/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/mbi/Repos/sardana/src/sardana/sardanathreadpool.py", line 64, in run
    Worker.run(self)
  File "/home/mbi/Repos/taurus/lib/taurus/core/util/threadpool.py", line 145, in run
    callback(cmd(*args, **kw))
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/msmacromanager.py", line 1594, in __runXML
    self.__runStatelessXML()
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/msmacromanager.py", line 1610, in __runStatelessXML
    self.__runXMLMacro(xml_macro)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/msmacromanager.py", line 1617, in __runXMLMacro
    macro_obj, _ = self._prepareXMLMacro(xml)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/msmacromanager.py", line 1277, in _prepareXMLMacro
    prepare_result = self._prepareMacroObj(macro_obj, macro_params)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/msmacromanager.py", line 1300, in _prepareMacroObj
    prepare_opts=prepare_opts)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/msmacromanager.py", line 825, in prepareMacroObj
    return macro.prepare(*par_list, **prepare_opts)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/macros/scan.py", line 392, in prepare
    nr_interv, integ_time, **opts)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/macros/scan.py", line 162, in _prepare
    moveables, env, constrains, extrainfodesc)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/scan/gscan.py", line 1108, in __init__
    extrainfodesc=extrainfodesc)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/scan/gscan.py", line 400, in __init__
    self._setupEnvironment(env)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/scan/gscan.py", line 717, in _setupEnvironment
    shape = _get_shape(channel)
  File "/home/mbi/Repos/sardana/src/sardana/macroserver/scan/gscan.py", line 90, in _get_shape
    shape = channel.shape
  File "/home/mbi/.local/lib/python3.6/site-packages/tango/device_proxy.py", line 305, in __DeviceProxy__getattr
    return __get_attribute_value(self, attr_info, name)
  File "/home/mbi/.local/lib/python3.6/site-packages/tango/device_proxy.py", line 247, in __get_attribute_value
    attr_value = self.read_attribute(name).value
  File "/home/mbi/.local/lib/python3.6/site-packages/tango/green.py", line 195, in greener
    return executor.run(fn, args, kwargs, wait=wait, timeout=timeout)
  File "/home/mbi/.local/lib/python3.6/site-packages/tango/green.py", line 109, in run
    return fn(*args, **kwargs)
  File "/home/mbi/.local/lib/python3.6/site-packages/tango/device_proxy.py", line 394, in __DeviceProxy__read_attribute
    return __check_read_attribute(self._read_attribute(value, extract_as))
  File "/home/mbi/Repos/sardana/src/sardana/tango/pool/PoolDevice.py", line 378, in _read_DynamicAttribute
    return read(attr)
  File "/home/mbi/Repos/sardana/src/sardana/tango/pool/PoolDevice.py", line 949, in read_Shape
    attr.set_value(self.element.shape)
  File "/home/mbi/Repos/sardana/src/sardana/pool/poolbasechannel.py", line 656, in get_shape
    self._set_shape(shape, propagate=propagate)
  File "/home/mbi/Repos/sardana/src/sardana/pool/poolbasechannel.py", line 664, in _set_shape
    EventType("shape", priority=propagate), shape)
  File "/home/mbi/Repos/sardana/src/sardana/sardanabase.py", line 132, in fire_event
    listeners=listeners)
  File "/home/mbi/Repos/sardana/src/sardana/sardanaevent.py", line 110, in fire_event
  File "/home/mbi/Repos/sardana/src/sardana/sardanaevent.py", line 132, in _fire_event
    real_listener(self, event_type, event_value)
  File "/home/mbi/Repos/sardana/src/sardana/tango/pool/PseudoCounter.py", line 98, in on_pseudo_counter_changed
    event_value)
  File "/home/mbi/Repos/sardana/src/sardana/tango/pool/PseudoCounter.py", line 158, in _on_pseudo_counter_changed
    priority=priority, error=error, synch=False)
  File "/home/mbi/Repos/sardana/src/sardana/tango/core/SardanaDevice.py", line 317, in set_attribute
    priority=priority, synch=synch)
  File "/home/mbi/Repos/taurus/lib/taurus/core/util/threadpool.py", line 147, in run
    cmd(*args, **kw)
  File "/home/mbi/Repos/sardana/src/sardana/tango/core/SardanaDevice.py", line 329, in set_attribute_push
    error=error, priority=priority)
  File "/home/mbi/Repos/sardana/src/sardana/tango/core/SardanaDevice.py", line 349, in _set_attribute_push
    "Cannot set value of attribute '%s' with None" % (attr_name,))
Exception: Cannot set value of attribute 'shape' with None
```
I also checked that the `shape` is not requested for any scan later than the first one and hence also the errors just appear once.
I tested with `sar_demo` devices only, so it should be reproducible for you.

reszelaz:
Thanks @dschick for details about the errors. I confirm both problems. I already fixed them.
One can try it with the RoI controller proposed in https://github.com/sardana-org/sardana/pull/1482.

Here is a test that I performed:

```
Door_zreszela_1 [89]: twodroi01.roi = [1,1023,1,1023]

Door_zreszela_1 [90]: twodroi01.read_attribute("shape").value
        Result [89]: array([1022, 1022])

Door_zreszela_1 [91]: ascan mot01 0 1 1 0.1
Operation will be saved in /tmp/scan.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #2 started at Mon Jan 25 12:18:33 2021. It will take at least 0:00:00.482843
 #Pt No    mot01     twod01   twodroi01     dt   
   0         0      (1024, 1024)  (1022, 1022)  0.036572
   1         1      (1024, 1024)  (1022, 1022)  0.513074
Operation saved in /tmp/scan.h5 (HDF5::NXscan)
Scan #2 ended at Mon Jan 25 12:18:34 2021, taking 0:00:00.712399. Dead time 71.9% (motion dead time 43.8%)

Door_zreszela_1 [92]: twodroi01.roi = [1,1022,1,1022]

Door_zreszela_1 [93]: ascan mot01 0 1 1 0.1
Operation will be saved in /tmp/scan.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #3 started at Mon Jan 25 12:18:47 2021. It will take at least 0:00:00.765685
 #Pt No    mot01     twod01   twodroi01     dt   
   0         0      (1024, 1024)  (1021, 1021)  0.310781
   1         1      (1024, 1024)  (1021, 1021)  0.799457
Operation saved in /tmp/scan.h5 (HDF5::NXscan)
Scan #3 ended at Mon Jan 25 12:18:48 2021, taking 0:00:00.996008. Dead time 79.9% (motion dead time 59.0%)
```

dschick:
I can confirm that this PR is working.

Great job!

This will save a lot of time and tears!

reszelaz:
On the last follow-up meeting @dschick suggested to not remove the shape column from the expconf and leave it as read only cell with the value of the shape attribute.
It is certainly feasible to implement it. I think the easiest would be to insert in the cell a `TaurusLabel` (using `setIndexWidget()` of the table view) and connect it to the Shape Tango attribute. There is one detail that we should keep in mind. This attribute does not have events and taurus will poll it (by default every 3 s).
Any opinions about this approach?

dschick:
hm, the 3s polling might be an issue.
I am not sure if all camera APIs like it to be polled for an attribute during an acquisiton?
Could cause some trouble for some people...

teresanunez:
Hi Zibi, I think it should be fine like this. The polling every 3 s should not be a problem for
an attribute showing the shape.

teresanunez:
Sorry, I was not thinking about reading from the hardware ... only in the frequency at which the shape will be
changed. But I think it is not needed to read from hardware every time ...

dschick:
@teresanunez I am just thinking of the option to NOT use a TangoDS with its polling features as an interlayer between the hardware API and Sardana. So when directly calling the hardware API from a Sardana TwoDcontroller you would need to add some logic into the controller to avoid doing the hardware calls all the time.

We also programmed a camera TangoDS which returns the shape via a command and not as an attribute...
So all these scenarios could be problematic for the polling of the shape.

So maybe it is fine to get rid of the shape as it is also displayed after a `ct` and in the `scans`?

reszelaz:
Thanks @teresanunez and @dschick for your feedback!

As @teresanunez says it would be possible to implement some logic in the controller e.g. if no one changed the RoI nor binning (assuming these are extra axis attributes exposed by the controller), then the shape read could return the last written shape. This would not work if one changed the shape externally e.g. talking directly to the interlayer Tango device, or via a manufacturer's interface, etc. Anyway, if we go for that it would need to be well documented.

Not having the shape in the expconf (the initial idea) to avoid the above complexity is another option. As @dschick says the shape is immediatelly printed in the scan records. It is also easily accessible from the [PCTV](https://sardana-controls.org/users/taurus/poolchanneltv.html) widget.

teresanunez:
Hi Zibi, thanks, I really think that for DESY is not a problem at all if the shape is not in expconf.

reszelaz:
Based on this analysis and your answers I understand we are more in favor of eliminating this column completelly from the expconf. I will proceed with the necessary work if there are no other opinions.

reszelaz:
I have removed the shape from the expconf widget. Now we just need to:
- verify if the backwards compatibily asked by @teresanunez in #1296 (comment) works as expected. @teresanunez could you please do the necessary tests?
- add documentation - I can take care about that.

teresanunez:
Hi Zibi,
             thanks, but i do not have any camera for testing the backwards compatibility in a real case.
I would go ahead. If there are problems, we will see them in the real work.

reszelaz:
I have added the missing documentation in: ff0f2ec (2D), a4b43c7 (1D) and 43e5cb5 (pseudo counters).
It is now ready for review. @sardana-org/integrators could you please take a look? Many thanks!

teresanunez:
Now that I read the documentation ... how do you want to keep the compatibility if there is 'must' in defining
the shape in GetAxisPar?

reszelaz:
Thanks @teresanunez for reviewing it!

> Now that I read the documentation ... how do you want to keep the compatibility if there is 'must' in defining
> the shape in GetAxisPar?

This should be provided by this commit: 844de34 and more preciseslly by this [fallback mechanism](https://github.com/reszelaz/sardana/blob/f40114d41348d01796e87a906caa0163d396dd39/src/sardana/pool/poolbasechannel.py#L672).

teresanunez:
Hi ZIbi, thanks, I had something like this in mind ... But then I think that the documentation should clearly says
that you need to provide the shape in the form described there if you want to provide it, but it is not mandatory
to implement that in your controller if you don't care about the shape. For example we mainly do not care
about the shape since we do not pass the images to Sardana.

reszelaz:
When studying this issue, in https://github.com/sardana-org/sardana/issues/1296#issuecomment-595832878, I identified that shape is necessary for most of the built-in recorders:

> I think that delaying the shape determination until the first data arrives to the recorder won't be so easy. Yes, we could do it for the H5 recorder, but I reviewed the recorders that we have in Sardana and most of them use the shape information already in the _startRecordList. For example H5 do it to create the dataset, JSON do it to skip 1D and 2D channels from serialization, SPS to call sps.create(), FIO to organize channels in CT and MCA groups, SPEC to place MCA shape in the scan header, etc..
So, there are recorders which need this information when starting the record list and delaying the dataset creation in the case of H5 would be just a workaround.

How is it working now (before this PR) was briefly explained in https://github.com/sardana-org/sardana/issues/1296#issuecomment-597676363:

> When 1D/2D channel was added to the measurement group in the expconf its Value attribute was read and its shape was used in the configuration. 

This lacks one detail, when it was not possible to read the attribute, the Tango _max dim_ was used.

And it is quite common that it is not possible to read the value attribute, imagine that you have not performed any acquisition so far and you want to add a channel to the measurement group. 

The approach taken in this PR was to make it a _must_ to implement shape in the controller, providing a backwards compatibility to facilitate transition. If the shape does not change one could basically return a harcoded shape in the `GetAxisPar()`.

reszelaz:
As @teresanunez explains, at DESY they neither pass images with `ReadOne()` nor the references with `RefOne()`. What they do is to pass a minimum array filled with -1 in the `ReadOne()`. This produces some shape, neither realisting not relevant, which successfully _cheats_ the recorders.

What I have done in 77b1177db is to implement another backwards compatibility fallback and in case the shape can not be determined from the value, then use the `MaxDimSize` which is currently hardcoded at 4096 x 4096 but could be changed in the controller by implementing e.g.:
```python
        attrs = super().GetAxisAttributes(axis)
        attrs['Value'][MaxDimSize] = (512, 512)
        return attrs
```
Whenever in the future we will consider removing the backwards compatibility we could decide if we keep the shape auto-discovery from the `MaxDimSize` or at that time we could implement some other mechanism or provide a default shape which could be changed in a more _user friendly_ way than the one presented above.

If you agree to keep 77b1177db then in the documentation I would change "must" -> "should" in:

> 2D controller must provide a shape of the image which will be produced by acquisition.

And add another sentence about the backwards compatibility.

teresanunez:
Hi Zibi, many thanks for that implementation. From my point of view it can be kept and you can proceed to
change the documentation. Thanks again.

reszelaz:
> From my point of view it can be kept and you can proceed to change the documentation.

I've added info on backwards compatibility to the docs in 27124fe. 

@teresanunez I think you can proceed with the review now. Many thanks!

teresanunez (approved):

reszelaz:
Thanks @teresanunez for the review. I'm merging this one.

PR #1470, opened 2021-01-14
Complement and fix docs on setting ORBendPoint

reszelaz:
* document the possibility to fix IP address
* fix examples to use "giop:tcp:" prefix

This is a trivial PR to add some additional infos to the docs hence I will auto-merge.


reszelaz:
Auto-merging.

PR #1473, opened 2021-01-15
Make write of integration time more robust

reszelaz:
Write of the integration time may fail e.g. problems with reading latency_time of the controller, and we fill the cache anyway. This causes problems with not initialized synchronization description.

Fill integration time cache only if the write worked.


cpt-majkel (approved):
LGTM

Issue #1421, opened 2020-09-23
Cannot stop multiple motors movement in backlash direction


cpt-majkel:
Hi all,

this week we discovered a problem with stopping motors (with Ctrl + C). Problem occurs while moving multiple motors with applied backlash, in backlash direction. 

```
Door_sar3demo_1 [90]: mot06.backlash=5

Door_sar3demo_1 [91]: mot05.backlash=5

Door_sar3demo_1 [92]: umv mot05 100 mot06 100
     mot05      mot06
  100.0000   100.0000

Door_sar3demo_1 [93]: umv mot05 10 mot06 10
     mot05      mot06
   81.7474    81.7442
^C
Ctrl-C received: Stopping...
Stopping Motion(['mot05', 'mot06']) reserved by mv
Motion(['mot05', 'mot06']) stopped
Executing mv.on_stop method...
     mot05      mot06
   81.7474    81.7442
Executing umv.on_stop method...
Stopping done!

Door_sar3demo_1 [94]: wa
Current positions (user, dial) on 2020-09-23 14:10:16.097051

         gap02      ice1      ice2     mot05     mot06     mot07     mot08
User   20.0000   10.0000   20.0000   10.0000   10.0000    0.0000    0.0000
Dial   20.0000   10.0000 2619476.2000   10.0000   10.0000    0.0000    0.0000
                                                                          
      offset02
User    0.0000
Dial    0.0000
```

This happens only with multiple motors movement. When moving one motor in backlash direction or multiple motors in opposite to backlash direction, stopping works fine.

I've done tests on sardana `2.8.3` as well as `3.0.2` using `dummy` motors and `IcePAP` ones, behaviour is the same.  The main movement seems to be interrupted, but the backlash is applied nonetheless.

Cheers,
Michal

cpt-majkel:
Hi, 

After some tests with @reszelaz we have come to the following conclusions:

- #1474 has been proposed to fix this bug
- Further discussion has to be made in order to define the correct behaviour. Our suggestion is to stop the motors and apply the backlash from the current position after first `ctrl + C` and stop the movement (`abort`) after the second one. The decision will be made at the next follow-up meeting. Of course, we are open to suggestions on this issue.

Cheers
Michal

PR #1474, opened 2021-01-15
Set stop flags for each element in multiple elements movement

cpt-majkel:
Related to #1421 

Sets `stop` flag for each element in multiple elements movement. It results in correct (`interrupted`) state of the moveables and does not allow to apply backlash.

### DIFF
--- a/src/sardana/pool/poolbasegroup.py
+++ b/src/sardana/pool/poolbasegroup.py
@@ -358,6 +358,8 @@ class PoolBaseGroup(PoolContainer):
             self.debug("Stopping %s %s", ctrl.name,
                        [e.name for e in elements])
             try:
+                for el in elements:
+                    el.stop()
                 error_elements = ctrl.stop_elements(elements=elements)
                 if len(error_elements) > 0:
                     element_names = [elem.name for elem in error_elements]


src/sardana/pool/poolbasegroup.py
```diff
@@ -358,6 +358,8 @@ def stop(self):
             self.debug("Stopping %s %s", ctrl.name,
                        [e.name for e in elements])
             try:
+                for el in elements:
```
reszelaz, line 361: I would move all the loop before of the try..except. Here we are just setting an internal flag, nothing can go wrong.

reszelaz (changes requested):
Thanks @cpt-majkel for the PR. I still had no time to test it but I have a small suggestion already - see my inline comment. BTW, I think we could easily add an automatic test on the core level (without the need to use Tango).

src/sardana/pool/poolbasegroup.py
cpt-majkel: Moved

reszelaz:
@cpt-majkel, I reviewed the PR and it is looking very good. Especially the testing part with the fixtures. But I still have some questions about the testing part. Better let's organize some online meetings to discuss them.

Meanwhile, in order to not block the integration of the fix and the Jan20 release, I will proceed to manually merge the very first three commits (1bca450, f573fcc and c47263e).
These three commits successfully tests the fix for the cases of a motor and a motor group (I have also run it locally) and I have manually tested it for pseudo motors. Everything looks ok. 
Many thanks!

PR #1476, opened 2021-01-20
Add macro name and line to macro status 'start' event

tiagocoutinho:
I would be useful to have the macro name and the macro line information in the macro status "start" event.

This PR adds that.


reszelaz (approved):
LGTM, I just fixed a test that was wrongly programmed and failed here.

PR #1457, opened 2020-12-10
Allow SWMR mode and avoid flock problems with NXscanH5_FileRecorder

reszelaz:
Recently there were several issues somehow related with the H5 data storage e.g. #771, #1296, #1448, etc.. and for a developer it will be much more convenient to not have the problems with the file locking when inspecting the recorded data with tools like silx. This motivated me to dedicate some time to try to advance with #1124.

First, I wanted to dedicate the minimum possible time to achive an improvement and I discarded from the beginning more complex changes e.g. making recorders long lived objects in Sardana.

Second, the issue described in #1124 can be reproduce with automatic test added in 7747c94 or manually.

To reproduce manually:
1. Execute a scan and store its data with the `NXscanH5_FileRecorder`
    ```
    newfile /tmp/test.h5
    ascan mot01 0 1 1 0.1
    ```
2. Open the file with `silx` (I tested with version 0.12) e.g.
    ```
    silx view /tmp/test.h5
    ```
3. Execute another scan:
    ```
    ascan mot01 0 1 1 0.1
    ```
Now, very briefly, what this PR introduces is a concept of _write session_ for the NXscanH5_FileRecorder. This session can be manually started and ended using macros or programmatically (in a macro code) using a context manager.
The session opening and closing is based on the `ScanFile` and `ScanDir` environment variables so it respects different levels of environment (door and macro).

More details in the documentation: 8941294.

I have added an atexit hook to close the still opened files at the server shutdown in a466bde but most probably due to https://github.com/taurus-org/taurus/issues/982 it is never called. 

In case one forgot to end the session and already shutdown the server it is possible to clear the SWMR flag with `h5clear -s <file_name>` (not available in Debain stretch, but available in Debian buster and conda).

Some ideas on how to improve it:
1. Add `pre-newfile` hook place to newfile in order to allow automatic closing and opening sessions when changing file
2. Allow to control session from a distinct `ScanFile` and `ScanDir` level or directly using file path.
3. Automatic offering of previous session ending on a new session start (@cpascual idea, thanks!)

I have a feeling that 2 and 3 are more time consuming, while 1 would be very easy to add. Anyway, I would like to hear your opinion about all this before investing more time to this topic. 

If we agree to merge it, I would prefer to mark is as experimental feature (allowing API changes if strictly necessary).

PR #1478, opened 2021-01-21
[WIP] Generic data recorder

jordiandreu:
Implements a mechanism to register a custom-user generic data recorder. The mechanism enables the possibility to register one or more custom data recorders. The recorders get configured by setting the sardana environment variable DataRecorder with one or more recorder class names.

Documentation about the implementation and configuration of a custom data recorder still missing.

### DIFF
--- a/doc/source/devel/howto_recorders.rst
+++ b/doc/source/devel/howto_recorders.rst
@@ -42,6 +42,10 @@ Sardana provides the following standard recorders (grouped by types):
     * JsonRecorder [*]
     * OutputRecorder
 
+* generic [*]
+.. TODO Need to document the implementation and configuration
+
+
 [*] Scan Framework provides mechanisms to enable and select this recorders using
 the environment variables.
 
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -349,6 +349,9 @@ class GScan(Logger):
         # The Output recorder (if any)
         json_recorder = self._getJsonRecorder()
 
+        # The Generic Data recorders (if any)
+        data_recorders = self._getDataRecorders()
+
         # The File recorders (if any)
         file_recorders = self._getFileRecorders()
 
@@ -360,6 +363,8 @@ class GScan(Logger):
 
         data_handler.addRecorder(output_recorder)
         data_handler.addRecorder(json_recorder)
+        for data_recorder in data_recorders:
+            data_handler.addRecorder(data_recorder)
         for file_recorder in file_recorders:
             data_handler.addRecorder(file_recorder)
         data_handler.addRecorder(shm_recorder)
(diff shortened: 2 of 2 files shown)


doc/source/devel/howto_recorders.rst
```diff
@@ -42,6 +42,10 @@ Sardana provides the following standard recorders (grouped by types):
     * JsonRecorder [*]
     * OutputRecorder
 
+* generic [*]
+.. TODO Need to document the implementation and configuration
```
tiagocoutinho, line 46: Need at least one blank line before the `.. TODO` comment otherwise travis complains

reszelaz:
Thanks @jordiandreu and @tiagocoutinho for working on this PR!
Just a comment on the documentation, there is https://github.com/sardana-org/sardana/pull/1457 which moves around some documentation related to the recorders. It would be nice to get that one integrated first so you work already on the reorganized document.

Also, there is the environment variables catalgue which would need to be updated with the new environment variable.

Regarding the implementation, do you foresee to parametrize somehow the initialization of the recorder class? Like for example the file recorders they receive the path to the file?

tiagocoutinho:
> Regarding the implementation, do you foresee to parametrize somehow the initialization of the recorder class? Like for example the file recorders they receive the path to the file?

Yes, the idea was that each recorder would be responsible to figure out which parameters it needs to configure itself.
See example [here](https://github.com/ALBA-Synchrotron/sardana-streams/blob/f9c917f14bf475896c64f65a9ddc09e14390af76/sardana_streams/recorder/recorder.py#L7)

reszelaz:
Perfect! Thanks for the clarification!

reszelaz:
When reviewing issues I just found again #1274, which I think would benefit from this PR. @aureocarneiro, @13bscsaamjad how were you configuring the KafkaRecorder mentioned in https://github.com/sardana-org/sardana/issues/1274#issuecomment-579141745?

13bscsaamjad:
we have found that if you put multiple ScanFile and ScanRecorder, sardana uses them all sequentially. the catch is to have same number of ScanFile as ScanRecorder. we only tested with two though. embedding the example config.
```
senv ScanFile "['test1', 'test2']" 
senv ScanRecorder "['Scicatrecorder', 'H5Recorder']"
```

teresanunez:
Sorry if I have not understood what you mean, we use at DESY very very oft several recorders and the scanfile is the same
for all of them, only by the termination we know the format. For us it has to continue working like this.

reszelaz:
@teresanunez there are no plans to change/remove activation of file recorders based on the file extension.
The `ScanRecorder` variable was added in the past to explicitelly activate a recorder regardless of the one that Sardana would assign based on the file extension. This PR is to propose a mechanism for activating data recorders which do not write to a file.

Thanks for sharing your configuration @13bscsaamjad.
If I understands well the ScanFile value "test1" corresponding to the ScanRecorer value "Scicatrecorder" does not correspond to any real file - this recorder does not write to a file, right? So you used it as a placeholder so there is a match of elements in both lists?

13bscsaamjad:
@reszelaz yes exactly.

Issue #1472, opened 2021-01-15
access movable of parent macro


dschick:
Hi, I'd like to write a macro that is hooked in the `pre-move/post-move` hook places.
The action of the hooked macro should depend on the motor(s) moved within the parent macro (parent can be a `scan` or `mv`).

So I need a specific action only in case one specific motor is moved and not if any other is moved.

What would be the way to access the names of the movables used in the parent macro?

currently I can do

```python
parent = self.getParentMacro()
```
and then I have not found an according property/method of the parent.

Thanks in advance

Daniel

reszelaz:
Hi Daniel,
It sounds to me familiar:) I remember we discussed it in https://github.com/sardana-org/sardana/issues/891. At that time I was suggesting to go for the GScan `moveables` member cause this one is always there, even if one writes a custom scan macro using the ScanFramework. All the standard scan macros also has `motors` attribute. Since now we are discussing if adding the `pre-move/post-move` hook places to the `mv` macros, in case we add them, then we will need to create a new member in the `mv` macro. In this case it make sense to call it `motors` as well. Then, I would reconsider my recomendation from #891 and make the `motors` a standard member of all macros having the said hook places.

dschick:
Thanks for your response. Indeed I am especially asking also because of the `mv` macro and the according hooks.
I am in general wondering, why the scan framework is not simply refractoring the `mv` macro at every step?
Is this a problem with the continious scans?

reszelaz:
On the last follow-up meeting we said to add motors attribute to all built-in macros executing motion and at the same time exposing `pre-move` and `post-move` hook places. This could be already done in #1480 so the `mv` and `umv` macros have this attribute.

marouanebj:
Hi, I add my question here.
Is it possible to access from a hook the inputed params of the scan macro.
Exemple:
I have the macro "mesh" that uses 
```python
param_def = [
        ['motor1', Type.Moveable, None, 'First motor to move'],
        ['m1_start_pos', Type.Float, None, 'Scan start position for first '
                                           'motor'],
        ['m1_final_pos', Type.Float, None, 'Scan final position for first '
                                           'motor'],
        ['m1_nr_interv', Type.Integer, None, 'Number of scan intervals'],
        ['motor2', Type.Moveable, None, 'Second motor to move'],
        ['m2_start_pos', Type.Float, None, 'Scan start position for second '
                                           'motor'],
[...]

PR #1480, opened 2021-01-21
Fix #1471: add pre-move & post-move hook places to mv-macro

### LINKED ISSUE #1471: pre- and post-move hooks not available in mv macro
Hi all,

I have an issue with the pre- and post-move hook place which is similar to #770 .

In SPEC there are hook place like pre-acq/post-acq and pre-move/post-move.
They are run everytime you do an acquisition (pre/post-acq) or a movement of a motor (pre/post-move) - independent if the parent macro is a scan or a `ct` or `mv` macro.

In Sardana was solved for `ct` with #808 but is not working for `umv` or `mv`.

I would suggest to add the hook places pre/post-move also to these macros as it is more consistent and our case also necessary.

Best

Daniel

dschick:
This PR adds the `pre-move` & `post-move` hook places to `mv` macro.
A new `sardanacustomsetting ` parameter named `PRE_POST_MOVE_HOOK_IN_MV` is introduced to disable the `pre-move` and `post-move` hook places in the `mv` macro by default.

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -477,8 +477,8 @@ class mv(Macro, Hookable):
                                True)
 
         if enable_hooks:
-            for preAcqHook in self.getHooks('pre-move'):
-                preAcqHook()
+            for preMoveHook in self.getHooks('pre-move'):
+                preMoveHook()
 
         self.motors, positions = [], []
         for m, p in motor_pos_list:
@@ -495,8 +495,8 @@ class mv(Macro, Hookable):
             self.info("\n".join(msg))
 
         if enable_hooks:
-            for postAcqHook in self.getHooks('post-move'):
-                postAcqHook()
+            for postMoveHook in self.getHooks('post-move'):
+                postMoveHook()
 
 
 class mstate(Macro):
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -576,9 +576,10 @@ class mvr(Macro, Hookable):
     ]
 
     def run(self, motor_disp_list):
-        motor_pos_list = []
+        self.motors, motor_pos_list = [], []
         for motor, disp in motor_disp_list:
             pos = motor.getPosition(force=True)
+            self.motors.append(motor)
             if pos is None:
                 self.error("Cannot get %s position" % motor.getName())
                 return
@@ -597,9 +598,10 @@ class umvr(Macro, Hookable):
     param_def = mvr.param_def
 
     def run(self, motor_disp_list):
-        motor_pos_list = []
+        self.motors, motor_pos_list = [], []
         for motor, disp in motor_disp_list:
             pos = motor.getPosition(force=True)
+            self.motors.append(motor)
             if pos is None:
                 self.error("Cannot get %s position" % motor.getName())
                 return
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -459,9 +459,10 @@ class pwm(Macro):
         self.execMacro('wm', motor_list, **Table.PrettyOpts)
 
 
-class mv(Macro):
+class mv(Macro, Hookable):
     """Move motor(s) to the specified position(s)"""
 
+    hints = {'allowsHooks': ('pre-move', 'post-move')}
     param_def = [
         ['motor_pos_list',
          [['motor', Type.Moveable, None, 'Motor to move'],
@@ -470,6 +471,9 @@ class mv(Macro):
     ]
 
     def run(self, motor_pos_list):
+        for preAcqHook in self.getHooks('pre-move'):
+            preAcqHook()
+
         motors, positions = [], []
         for m, p in motor_pos_list:
             motors.append(m)
@@ -483,6 +487,9 @@ class mv(Macro):
             for motor in motors:
                 msg.append(motor.information())
             self.info("\n".join(msg))
+
+        for postAcqHook in self.getHooks('post-move'):
+            postAcqHook()
 
 
 class mstate(Macro):
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -45,6 +45,7 @@ from sardana.macroserver.macro import Macro, macro, Type, ViewOption, \
 from sardana.macroserver.msexception import StopException, UnknownEnv
 from sardana.macroserver.scan.scandata import Record
 from sardana.macroserver.macro import Optional
+from sardana import sardanacustomsettings
 
 ##########################################################################
 #
@@ -470,9 +471,15 @@ class mv(Macro, Hookable):
          None, 'List of motor/position pairs'],
     ]
 
+    def prepare(self):
+        selsardanacustomsettings
+
     def run(self, motor_pos_list):
-        for preAcqHook in self.getHooks('pre-move'):
-            preAcqHook()
+        enable_hooks = getattr(sardanacustomsettings, 'PRE_POST_MOVE_HOOK_IN_MV')
+
+        if enable_hooks:
+            for preAcqHook in self.getHooks('pre-move'):
+                preAcqHook()
 
         motors, positions = [], []
         for m, p in motor_pos_list:
@@ -488,8 +495,9 @@ class mv(Macro, Hookable):
                 msg.append(motor.information())
             self.info("\n".join(msg))
 
-        for postAcqHook in self.getHooks('post-move'):
-            postAcqHook()
+        if enable_hooks:
+            for postAcqHook in self.getHooks('post-move'):
+                postAcqHook()
 
 
 class mstate(Macro):
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -471,9 +471,6 @@ class mv(Macro, Hookable):
          None, 'List of motor/position pairs'],
     ]
 
-    def prepare(self):
-        selsardanacustomsettings
-
     def run(self, motor_pos_list):
         enable_hooks = getattr(sardanacustomsettings, 'PRE_POST_MOVE_HOOK_IN_MV')
 
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -473,7 +473,8 @@ class mv(Macro, Hookable):
 
     def run(self, motor_pos_list):
         enable_hooks = getattr(sardanacustomsettings,
-                               'PRE_POST_MOVE_HOOK_IN_MV')
+                               'PRE_POST_MOVE_HOOK_IN_MV',
+                               True)
 
         if enable_hooks:
             for preAcqHook in self.getHooks('pre-move'):
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -472,6 +472,11 @@ class mv(Macro, Hookable):
     ]
 
     def run(self, motor_pos_list):
+        self.motors, positions = [], []
+        for m, p in motor_pos_list:
+            self.motors.append(m)
+            positions.append(p)
+
         enable_hooks = getattr(sardanacustomsettings,
                                'PRE_POST_MOVE_HOOK_IN_MV',
                                True)
@@ -480,11 +485,9 @@ class mv(Macro, Hookable):
             for preMoveHook in self.getHooks('pre-move'):
                 preMoveHook()
 
-        self.motors, positions = [], []
-        for m, p in motor_pos_list:
-            self.motors.append(m)
-            positions.append(p)
+        for m, p in zip(self.motors, positions):
             self.debug("Starting %s movement to %s", m.getName(), p)
+
         motion = self.getMotion(self.motors)
         state, pos = motion.move(positions)
         if state != DevState.ON:
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -479,17 +479,17 @@ class mv(Macro, Hookable):
             for preAcqHook in self.getHooks('pre-move'):
                 preAcqHook()
 
-        motors, positions = [], []
+        self.motors, positions = [], []
         for m, p in motor_pos_list:
-            motors.append(m)
+            self.motors.append(m)
             positions.append(p)
             self.debug("Starting %s movement to %s", m.getName(), p)
-        motion = self.getMotion(motors)
+        motion = self.getMotion(self.motors)
         state, pos = motion.move(positions)
         if state != DevState.ON:
             self.warning("Motion ended in %s", state.name)
             msg = []
-            for motor in motors:
+            for motor in self.motors:
                 msg.append(motor.information())
             self.info("\n".join(msg))
 
@@ -521,6 +521,7 @@ class umv(Macro):
             pos, posObj = motor.getPosition(force=True), motor.getPositionObj()
             self.all_pos.append([pos])
             posObj.subscribeEvent(self.positionChanged, motor)
+        self.motors = self.all_names
 
     def run(self, motor_pos_list):
         self.print_pos = True
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -520,13 +520,14 @@ class umv(Macro, Hookable):
     def prepare(self, motor_pos_list, **opts):
         self.all_names = []
         self.all_pos = []
+        self.motors = []
         self.print_pos = False
         for motor, pos in motor_pos_list:
             self.all_names.append([motor.getName()])
+            self.motors.append(motor)
             pos, posObj = motor.getPosition(force=True), motor.getPositionObj()
             self.all_pos.append([pos])
             posObj.subscribeEvent(self.positionChanged, motor)
-        self.motors = self.all_names
 
     def run(self, motor_pos_list):
         self.print_pos = True
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -511,9 +511,10 @@ class mstate(Macro):
         self.info("Motor %s" % str(motor.stateObj.read().rvalue))
 
 
-class umv(Macro):
+class umv(Macro, Hookable):
     """Move motor(s) to the specified position(s) and update"""
 
+    hints = {'allowsHooks': ('pre-move', 'post-move')}
     param_def = mv.param_def
 
     def prepare(self, motor_pos_list, **opts):
@@ -530,7 +531,9 @@ class umv(Macro):
     def run(self, motor_pos_list):
         self.print_pos = True
         try:
-            self.execMacro('mv', motor_pos_list)
+            mv, _ = self.createMacro('mv', motor_pos_list)
+            mv.appendHook(self.hooks)
+            self.runMacro(mv)
         finally:
             self.finish()
 
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -472,7 +472,8 @@ class mv(Macro, Hookable):
     ]
 
     def run(self, motor_pos_list):
-        enable_hooks = getattr(sardanacustomsettings, 'PRE_POST_MOVE_HOOK_IN_MV')
+        enable_hooks = getattr(sardanacustomsettings,
+                               'PRE_POST_MOVE_HOOK_IN_MV')
 
         if enable_hooks:
             for preAcqHook in self.getHooks('pre-move'):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,8 @@ This file follows the formats and conventions from [keepachangelog.com]
 * `sequencer` loading of plain text sequences in spock syntax with macro functions (#1422)
 * Recorders tests helpers (#1439)
 * Disable flake8 job in travis CI (#1455)
-* `createMacro()` and `prepareMacro()` docstring (#1460, #1444) 
+* `createMacro()` and `prepareMacro()` docstring (#1460, #1444)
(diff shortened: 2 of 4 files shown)


dschick:
please let me know, if you have a better idea for the name of the `sardanacustomsettings` parameter

dschick:
According to the dicussion in #1472 I would simply add a 
```
self.motors = [...]
```
to the `mv` and `umv` macros saving the list of involved motors in the `self.motors` attribute?

Do I have to put in the `prepare` method or is it sufficient to put in in the `run` method?

reszelaz:
> Do I have to put in the `prepare` method or is it sufficient to put in in the `run` method?

I would say that in the `run()` is enough. This attribute will be accessed in the hooks mainly. So it for the moment I would put it there.

teresanunez:
Hi, I have tested it and it works nicely. The only point, I think we have decided to put True as default.
In the file in this PR is set to False. A part from that, i would say that it can be merged.

teresanunez (approved):

reszelaz:

> The only point, I think we have decided to put True as default.
> In the file in this PR is set to False.

See my reasoning in https://github.com/sardana-org/sardana/issues/1471#issuecomment-767439805, I already talked with @teresanunez privatelly about it and she agrees on using `True` as default. @dschick If you are also fine with that and there are no other opinions, could you please make the necessary change and then we could proceed with the merge.

Many thanks to all!

dschick:
Just changed the default from False to True.
Shall I also update the docs within this PR?

teresanunez:
Hi Daniel, thanks. Do you mean to change the comments in the file sardanacustomsettings.py ?. Yes, please change that.

dschick:
Hi Teresa, I think I added the correct comments in the  file sardanacustomsettings.py
I was more thinking of added some information to the general Sardana documentation and this new custom setting?

teresanunez:
Hi Daniel, let´s see what Zibi says, but I think the idea is that by default the pre/post-move hooks will run with
mv/umv  and they have to be explicitly disabled in the sardanacustomsettings if one does not want to have them
active.

teresanunez:
So the default behaviour has to be True (not only the default value in the sardanacustomsettings)

dschick:
thanks for the hint, I added the default value `True` to `getattr`

teresanunez:
Thanks, IMO is fine now. The documentation would be great. But I think it could
be another PR, as you prefer.

dschick:
okay, then do it in another PR and first merge this one here

reszelaz (commented):
Thanks to both for working on this PR!

I agree to add documentation in another PR. I can start working on that in parallel.

I have still one cooment about this one, where `umv` macro is not `Hookable` and does not announce `allowsHooks` hint. I've not checked it but I suspect that it won't be possible to:
1. Attach [sequencer hooks](https://sardana-controls.org/users/taurus/sequencer.html#sequencer-ui) hooks to it
2. Attach [programmatic hooks](https://sardana-controls.org/devel/howto_macros/macros_general.html#sardana-macro-adding-hooks-support) to it
3. In the future, whenever we add more granular configuration of general hooks as described in https://github.com/sardana-org/sardana/issues/1471#issuecomment-767439805, it won't be possible to attach [general hooks](https://sardana-controls.org/users/macro_hooks.html#sardana-macros-hooks-general) only to the `umv` macro.

I think we would overcome all these points if wee make it inherit from `Hookable`, add the `allowsHooks` hint and then in the `run()` first create the `mv` macro and then transfer to it the hooks from the `umv`. What do you think? Should that work?

teresanunez:
One should try and test, I think.

reszelaz:
I tried point 1 from the above and it does not work.

dschick:
Hi @reszelaz ,

I am still struggling with accessing the `motors` from within a hooked macro at e.g. the `pre-move` hook place.
I have hooked a simple macro there and would like to output the moved motors with something like this:
```python
@macro()
def user_pre_move(self):
    parent = self.getParentMacro()
    if parent:
        self.output(parent._name)
        self.output(parent.motors)
```

But I get the following error in spock: `AttributeError: 'mv' object has no attribute 'motors'`

What is the correct syntax here?

src/sardana/macroserver/macros/standard.py
```diff
@@ -470,20 +472,32 @@ class mv(Macro):
...
+                               True)
+
+        if enable_hooks:
+            for preMoveHook in self.getHooks('pre-move'):
+                preMoveHook()
+
+        self.motors, positions = [], []
         for m, p in motor_pos_list:
-            motors.append(m)
+            self.motors.append(m)
```
reszelaz, line 485: We need to fill the `self.motors` before executing `pre-move` hooks. They may need to access to it.

In this case I would make two iterations. In the first one fill the `self.motors` and in the second one print the debug message. IMHO the debug message should stay as close as possible to the move itself.

reszelaz (changes requested):
Hi @dschick,

Sorry for replying so late.

> But I get the following error in spock: AttributeError: 'mv' object has no attribute 'motors'

> What is the correct syntax here?

See my code review comment about it.

dschick:
thanks, I changed it and now it is working fine for me.

dschick:
is there still something that I should change?
I havn't noticed that you requested changes yourself

reszelaz (commented):
Perfect! Thanks for the fix of `self.motors`. I think that these https://github.com/sardana-org/sardana/pull/1480#pullrequestreview-581102332 were still not addressed? IMHO it is better to fix these issues now, so the hooks could be used consistently between `mv` and `umv`.

If you prefer, I could take a look on them, but I will still need some time. So, if you advances earlier, then this would speed up the PR integration and the Jan21 release.

Maybe better keep the "Changes requested" so we don't forget that there is still something to be done.

dschick:
sorry, I missed that.

> first create the `mv` macro and then transfer to it the hooks from the `umv`

How do I atually transfer hooks from one macro to the other?

reszelaz:
> How do I atually transfer hooks from one macro to the other?

I have never tried it, but one naive approach could be something like this:

```python
class umv(Macro, Hookable):
    [...]
    def run(self, motor_pos_list):
            [...]
            mv, _ self.prepareMacro('mv', motor_pos_list)
            mv.hooks = self.hooks
            self.runMacro(mv)
            [...]
    [...]
```

dschick:
hm, not so easy ...

```
This macro defines its own hooks. Previously defined hooks, including the general ones, would be only called if these own hooks were added using the appendHook method or appended to the self.hooks.
umv
An error occurred while running umv:
TypeError: must be real number, not NoneType
```

dschick:
ah, actually the same then applies for `mvr` and `umvr`, right?
How about other macros, such as `ubr`?

dschick:
@reszelaz are you continuing this, or shall I update the PR using the direct hook assignment instead of the `appendHooks`?

reszelaz:
> @reszelaz are you continuing this, or shall I update the PR using the direct hook assignment instead of the `appendHooks`?

Yes, I'm reviewing it now. Give me a little more time I'm preparing a comment. Thanks!

reszelaz:
Thanks for working on that @dschick ! I agree, it is not so easy.

I tried your solution from d4cf62b but it does not work. This is because the interface between the `Hookable.hooks` property and `Hookable.appendHook()` method does not match. The `Hookable.hooks` property returns a list of all hooks where each of the hooks is a sequence of a callable and a list of _hook places_ where the callable should be called. While the `Hookable.appendHook()` expects only one hook and not a list. Following your approach, I have tried something like this:

```python
class umv(Macro, Hookable):
    [...]
    def run(self, motor_pos_list):
            [...]
            mv, _ self.prepareMacro('mv', motor_pos_list)
            for hook in self.hooks:
                mv.appendHook(hook)
            self.runMacro(mv)
            [...]
    [...]
```

and it works.

I wonder why you went for using the `Hookable.appendHook()`. I see two possibilities:
1. You were guided by this warning message: 
    > This macro defines its own hooks. Previously defined hooks, including the general ones, would be only called if these own hooks were added using the appendHook method or appended to the self.hooks."
2. You wanted to workaround:
    > An error occurred while running umv:
TypeError: must be real number, not NoneType

So, point 2 is a bug which I already fixed in #1522 so it should not be a problem anymore.
And point 1 is actually very important here. I try to explain why.

Let's consider a case where we execute an `umv` macro and we have a general `pre-move` hook defined. If we use `Hookable.appendHook()` on the `mv` macro object then we will execute the said `pre-move` hook twice. One will come from the `umv` and one from the `mv` (general hooks are attached to the macro at the macro object preparation step i.e. when you call `createMacro()`). Setting the `Hookable.hooks` overrides any previously attached hook and I think this is the behavior we would like to achieve. Use of the `mv` macro inside of the `umv` is just an implementation detail. In this case it will never have programmatic nor sequencer hooks, so overriding is safe here. But in this case, we don't want the warning to appear.
If you agree on this reasoning I will think on how to avoid the warning here.

dschick:
actually I was wondering about the first point. I thought I could not overwrite the hooks. But of course this is what we want.
Actually I just tested it myself with one macro in the general pre-move hook place and there the `appendHook` worked fine. I guess not with multiple macros.
Also I checked that the hooked macro is not executed twice, when calling the `umv` macro and this was somehow okay with for me. I thought that `appendHook` already takes care for not hooking duplicates.
Anyhow you could go on with this as you suggested.

Please remember my former comment about `mvr` and `umvr` macros as well.

reszelaz:
I have advanced with this as we agreed. dd6db360f avoids the warning message. 

I have also made the necessary changes for `mvr`, `umvr`, `br` and `ubr`. @teresanunez please could you test the hkl macros if they work as expected with this new hook places, please?

In the case of the `br` macro there is the non-blocking move option. I have not implemented the hooks execution in this case. Well, the `pre-move` hooks could be executed, but I don't see a way to execute the `post-move` ones, is it fine like this? 

@teresanunez, do you know other hkl macros where we should do the same?

What do you think if we make the macro's documentation to list the _allowed hook places_? I think that all (not only the Sardana built-in macros but also the third party macros) would benefit from this. I can investigate how to do it if you agree and propose another PR.

dschick:
hi @reszelaz many thanks for bringing this to an end.
I did not know about the non-blocking feature of the `br` macro.
so as I do not know about the use-case I would go for implementing at least the pre-move hook and throw a warning if there is a post-move hooked but cannot be executed.

I would appreciate the documentation of the allowd hook places!

dschick:
I just noted a small issue for `umv` and alike macros which transfer their hoosk to the underlying `mv` macro.
In case the hooks contain an output this leeds to the nasty behaviour in spock:
```
Door_scattering_1 [46]: umv mag_curr 0                                                                                
pre-move
  mag_curr
   -0.0000
post-move
  mag_curr
   -0.0000
```
So the updated position is printed once after the post-move hook macros are executed.
I think this has to do with the `finish()` method of the `umv` macro which is called after the hooks and prints the final position.

reszelaz:
> So commenting the self.printAllPos() will do the job but will not update the printed positions after the last movement event was thrown.

I already proposed a #1526 with this suggestion. Hope my reasoning is correct and we could improve the current behavior. Thanks!

dschick:
I have merged all of your changes as well as #1526 
Finally I noticed that I have assigned only the names and not the motor objects to `self.motors` in `umv`.

I also added the `self.motors` to `mvr` and `umvr`.

This is maybe anther issue, but how about the `self.motors` in `br` and `ubr`?
Should we add there `hkl` or the real physical angle motors to `self.motors`?
I would suggest to use `hkl` as it is also though to get the angles in the `ubr` command.

Other wise everythin is working fine now for me with `mv`, `umv`, `mvr`, `umvr`.
I was not able to test the `br` and `ubr`.

reszelaz:
> Finally I noticed that I have assigned only the names and not the motor objects to `self.motors` in `umv`.

Very good, I missed that when reviewing it. Yes, the scan macros deposit the moveables objects in `self.motors` so I would continue doing it this way.

> I also added the self.motors to mvr and umvr.

From the point of view of the `pre-move` or `post-move` hook developer, these are not strictrly necessary. At the end, what will be used are the `self.motors` of the `mv` macro object. But, it is OK to fill it, and it will be more consistent with what was 
proposed in https://github.com/sardana-org/sardana/issues/1472#issuecomment-76314859.

dschick:
> From the point of view of the `pre-move` or `post-move` hook developer, these are not strictrly necessary. At the end, what will be used are the `self.motors` of the `mv` macro object. But, it is OK to fill it, and it will be more consistent with what was
> proposed in [#1472 (comment)](https://github.com/sardana-org/sardana/issues/1472#issuecomment-76314859).

I was actually about to ask about that. If I ask for the `parent = self.getParentMacro()` in a `pre-move` hooked macro, when executing `umv` I do get `umv` returned and not `mv`.
In case would get the `mv` macro returned as `grant parrent` I could access `mv.motors` and I would not need to add the `self.motors` to `mvr, umvr, umv`. I like this idea very much, as this would also solve the issue for `ubr, br` and essentially every other macro that involves the `mv` macro.

So is this still a but or do I use the wrong systax to access the corret parent?

reszelaz:
Sorry about my previous message which created confusion. I was speaking without actually testing it. If it already works as you explain, then I think it is correct.

As I mentioned somewhere earlier, the fact that the `umv` internally executes the `mv` IMO an implementation detail. We must define `umv` (and others) as `Hookable`, and define the `allowsHooks`, and also, as you recently corrected, fill the `self.motors`. Otherwise, it won't be possible to attach hooks to the `umv`.

dschick:
thanks for clarification, however, it would be much nicer for the future if the hooks from the internally called `mv` macro will be called automatically without the workaround with `Hookable`, `allowsHooks`, and `self.motors`.
I guess this could be solved to some extent if all move-macros inherit from the same base class, which does above mentioned logic.

Still it will be more difficult for `br` and `ubr`.

I also wrote simple `uan` and `an` macros in analogy to SPEC to move `tth` and `th` simulataneously.
In this case I always have to remember to add all the logic for the hooks and motors.
So at least inheriting would make things much eaiser for macro developer here.

dschick:
I tested and updated `br` and `ubr` and it works as expected.
@teresanunez I decided to add the physical angles here as `self.motors`, instead of the `hkl` pseuodmotors.
Maybe we should have all the involved physical and the hkl pseudomotors listed here?
Since all of them can be changed It should not make a large difference.

teresanunez:
I think it is fine only with the physical motors.

reszelaz:
Thanks to both @dschick and @teresanunez for reviewing the HKL part!
I think it is then ready to be merged. @dschick can you confirm?

> I decided to add the physical angles here as self.motors, instead of the hkl pseuodmotors.
Maybe we should have all the involved physical and the hkl pseudomotors listed here?

This is a very interesting point. For example, if we execute a scan or a `mv` on a pseudo motor, then it is the pseudo motor in the `motors` list. If I understand correctly the `br` macro is like a `mv` on _hard coded_ pseudo motors (H, K and L). In this case, for symmetry, shouldn't these be the pseudo motors in the `motors` lists? The fact that underneath we execute a `mv` macro on the physical axes (angles) it is more implementation detail. I think that the `pre-move` and `post-move` hook developer would still be able to access to the physical motors from the pseudo motor object. Also, in principle, he would never get access to the `mv` macro object because the _parent macro_ will be the `br`. 

This is just my way of interpreting it, I may be wrong (I've not made any tests of that) and I rely on your decission cause you have much more experience with the HKL.

reszelaz:
@dschick, just a side comment/question on the PRs.

I think it would be more clear to the reviewer if we avoid merging other PRs, let's call them dependency PRs, into the one of the PR, let's call it dependent PR. Whenever the dependency PRs are getting merged into the develop branch, then we could update the dependent PR with the changes from the develop branch.

This way the list of the commits will only include the ones from the dependent PR, and the list of the changes as well. 

I understand the need to reflect on this kind of dependencies. So, what we could do instead, is to keep an updated list of dependency PRs in the dependent PR description (initial comment) and make comments in the dependent PR discussion whenever this list changes (so we get notifications on the updates). Or maybe there are better mechanisms/policies to solve this?

What you, and others, think about it?

teresanunez:
Hi Zibi, the macros br and ubr do not move the pseudomotors but the real motors directly. This is why I agree with
the implemention from Daniel.

reszelaz:
> the macros br and ubr do not move the pseudomotors but the real motors directly

Yes, this is what happens inside of the macro. But from the user point of view, the parameters that you pass are the H, K and L positions, so pseudo motors' positions.

Now, let's put aside the `br` and analyze the `mv` macro example (not the `mv` macro executed from inside of the `br`), and we move a pseudo motor with the `mv` macro, we also pass the pseudo motor's position as a target, and in this case in the `motors` we store the pseudo motor and not physical motors.

dschick:
Hi @reszelaz 
 
> @dschick, just a side comment/question on the PRs.
> 
> I think it would be more clear to the reviewer if we avoid merging other PRs, let's call them dependency PRs, into the one of the PR, let's call it dependent PR. Whenever the dependency PRs are getting merged into the develop branch, then we could update the dependent PR with the changes from the develop branch.
> 
> This way the list of the commits will only include the ones from the dependent PR, and the list of the changes as well.
> 
> I understand the need to reflect on this kind of dependencies. So, what we could do instead, is to keep an updated list of dependency PRs in the dependent PR description (initial comment) and make comments in the dependent PR discussion whenever this list changes (so we get notifications on the updates). Or maybe there are better mechanisms/policies to solve this?
> 
> What you, and others, think about it?

that is a very good idea, to keep the dependencies listed in the header. And I am sorry for the mess I produce sometimes, when fighting with git :)

I am just not so sure about, how I should use the dependencies in my PR branch without merging them from another PR branch?
What would be the correct git command for that?

> Now, let's put aside the `br` and analyze the `mv` macro example (not the `mv` macro executed from inside of the `br`), and we move a pseudo motor with the `mv` macro, we also pass the pseudo motor's position as a target, and in this case in the `motors` we store the pseudo motor and not physical motors.

I actually used the physical motors just because of what @teresanunez said, as the `hkl` are not used as pseudomotors.
However, I agree that one might be consistent here.
BUT for the case of `br` the `pre-move` and `post-move` developers will not be able to access the real physical motors from the pseudo motors, so in this specific case, it is a though decision.

> I think it is then ready to be merged. @dschick can you confirm?

I am happy with it. if you have tested it, it can be merged

reszelaz:
> that is a very good idea, to keep the dependencies listed in the header. And I am sorry for the mess I produce sometimes, when fighting with git :)

Don't worry! It happens to all of us:) IMHO, the best thing about Git is that it is so easy to redo things when something goes wrong.

> I am just not so sure about, how I should use the dependencies in my PR branch without merging them from another PR branch?
What would be the correct git command for that?

I'm not an expert here, but I would keep two branches: the one for the PR needs. And another one where I would put it together with the dependencies just for doing tests. There may be other ways, like for example https://github.com/isaacs/github/issues/959#issuecomment-642838613, but I think it may need that all the branches are in the upstream repo and not in the forks. Anyway, from the Sardana point of view, it is not worth to explore it since we will be moving to GitLab soon.

> BUT for the case of br the pre-move and post-move developers will not be able to access the real physical motors from the pseudo motors, so in this specific case, it is a though decision

Not directly, but the following should work :

```python
@macro()
def pre_move_hook(self):
    motors = []
    for moveable in self.parent_macro.motors:
        if moveable.type == "PseudoMotor":
            motors.extend([self.getMotor(name) for name in moveable.physical_elements])
        else:
            parent_physical_motors.append(moveable)
```

You could also use `moveable.elements` instead of `moveable.physical_elements` if you are sure that there is only one _pseudo layer_.

Again, I'm happy with your decission. I just wanted to share with you this observations. Thanks again @teresanunez and @dschick !

teresanunez:
Hi Zibi, the point is that the macro br does not use the pseudomotors at all ... As Daniel and myself said, the moveables
used are the motors of the angles, but not the motors corresponding to the coordinates h, k, l. So it is not that we move
the physical motors through the pseudomotors (like all the pseudomotors do) but we move directly the motors, the
pseudomotors are not used in the macro at all.

dschick:
in addition what @teresanunez said, you cannot move only `L` by `umv L` as it is not a typical pseudomotor, as `HKL` dependent many physical motors and the dependency between `hkl` and physical motors are not one-to-one.

reszelaz:
@teresanunez many thanks again for the clarification. I see the point about the implementation. If possible, I still would like to ask you some questions about it (just to understand better how the diffractometers are being used by the users), but since it is not urgent let's postpone it to the follow-up meeting. Could you please remind me about it if I forget to ask the next time we meet:)

I'm merging, thanks to all involved in this work!

reszelaz (approved):

cpascual:
Just a note: gitlab already has a specific field for dependencies in MRs. I am just mentioning it in case you prefer waiting for migration instead of deciding on an ad hoc solution

On 29 March 2021 13:06:05 CEST, reszelaz ***@***.***> wrote:
***@***.***, just a side comment/question on the PRs.
>
>I think it would be more clear to the reviewer if we avoid merging
>other PRs, let's call them dependency PRs, into the one of the PR,
>let's call it dependent PR. Whenever the dependency PRs are getting
>merged into the develop branch, then we could update the dependent PR
>with the changes from the develop branch.
>
>This way the list of the commits will only include the ones from the
>dependent PR, and the list of the changes as well. 
>
>I understand the need to reflect on this kind of dependencies. So, what
>we could do instead, is to keep an updated list of dependency PRs in
>the dependent PR description (initial comment) and make comments in the
>dependent PR discussion whenever this list changes (so we get
>notifications on the updates). Or maybe there are better
>mechanisms/policies to solve this?
>
>What you, and others, think about it?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/sardana-org/sardana/pull/1480#issuecomment-809290894

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

reszelaz:
> Just a note: gitlab already has a specific field for dependencies in MRs. I am just mentioning it in case you prefer waiting for migration instead of deciding on an ad hoc solution

Very good news! Thanks for letting us know. Yes, it is better to wait for the GitLab.

Issue #1159, opened 2019-06-25
Step scan generator with the possibility to not move motors


reszelaz:
This issue was created internally by @jairomoldes, and I just copy&paste here the report:

User wants to measure several points at the same energy. Sardana will always instruct the involved motor(s) to go to the target position(s) (even if they are the same).

Unfortunately currently there is no way to do this without modifying sardana gscan framework itself.

What I did was modifying gscan.py so step scans don't move the motors if the target position(s) are the same as the previous ones:

```
```

teresanunez:
Hi @reszelaz, we also needed something like this at DESY some years ago and I had implemented
 the 'general condition' feature, repeating points based on a condition defined by the user in a function.
 It is used at DESY. I don't know if it could be interesting for everybody

dschick:
Wouldn't you call that a timescan?
However, I noticed that Sardana timescan is not a normal step scan but a
continuous one.

teresanunez <notifications@github.com> schrieb am Di., 25. Juni 2019, 15:37:

teresanunez:
> Wouldn't you call that a timescan? However, I noticed that Sardana timescan is not a normal step scan but a continuous one. 
No, it is not a timescan, it is a step scan where some points have to be repeated.

dschick:
@teresanunez if only _some_ points have to be repeated, you are right.
But from what @reszelaz initially wrote, it sounds more like a timescan.
However, how would you start a scan with only  _some points to be repeated_?

teresanunez:
> @teresanunez if only _some_ points have to be repeated, you are right.
> But from what @reszelaz initially wrote, it sounds more like a timescan.
> However, how would you start a scan with only _some points to be repeated_?

Hi @dschick, the one from Zibi is also not a timescan because they need to move a motor (in
this case the energy) but then instead of taking only one data set in each point they want
to take more. It could be done with a step scan + a timescan in a post-acq hook but in that
 case you would not have the data recorded like for a single scan. 
In my case, there is a condition function, defined for the user, that decides if the point has to be repeated and how many times. The condition function usually looks at the accelerator current or
the status of the shutters and take a decision based on that.

jairomoldes:
Hi @teresanunez . Could you please provide further information about this "general condition"? I think it is of general interest

teresanunez:
Hi @jairomoldes, the implementation is done in gscan and is activated if an environement variable
called GeneralCondition is defined. This variable has to contain the name of a macro, and this macro
[...]

PR #1481, opened 2021-01-21
[WIP] Add general condition feature

reszelaz:
_General condition_ is a feature used successfully at DESY for some years already.
It allows the users to easily repeat the acquisitions if certain conditions are not satisfactory.
The user simply writes a macro which return value decides if the acquisition should be repeated and set it via `GeneralCondition` environment variable. You can find the original implementation authored by @teresanunez in https://github.com/desy-fsec/sardana (mainly `gscan.py` module).

Today with @teresanunez we had a pair-programming session in which we came with the following ideas:
- The [step scan generator](https://sardana-controls.org/devel/howto_macros/scan_framework.html#a-basic-example-on-writing-a-step-scan) seems to already fullfill most of the requirements (not the part of not repeating the motion to the same position but this one would be covered if we go for https://github.com/sardana-org/sardana/issues/1159#issuecomment-764702323). The _step scan generator_ however requires from the user to write a new scan macro. The _general condition_ feature is for users with less programming skills who do not want to learn how to program a new scan macro.
- The _general condition_ feature is not for pausing the scan and waiting for optimal conditions to continue - for these kind of needs it is better to use `pre-step`, `pre-move` or `pre-acq` hooks.

This PR changes the approach from the original implementation and moves the condition execution to the step scan generator of the built-in scan macros (for the moment only `aNscan`) which we believe fits better to the current design of the _scan framework_ - the `SScan` simply obeys what the generator ask it for. 

Even if the PR is marked as WIP it is already operational but it will keep moving the motors to the same position in case of repeating the acquisition - again see https://github.com/sardana-org/sardana/issues/1159#issuecomment-764702323. 

Before integrating it, apart from adding the documentation, I propose that we also clarify the following points:
1. Think if simply having another hook-place e.g. `repeat-step` (still executed from the generator) wouldn't be simpler instead of adding a new environment variable `GeneralHooks`. We could still keep the environment variable for the _backwards compatibility_ for DESY. This is more about recommendations for the new users of this feature.
2. At DESY some users already asked about extending this feature with the possibility of not saving the acquisitions points which do to fullfill the condition. Maybe this should actually be the default behavior of this feature?
3. Do we add this feature to the `fscan` and `mesh` macros as well?

It would be greate to hear your opinions about these points and this feature in general.

One important thing is that using the _general condition_ automatically disables the [deterministic scans optimizations](https://sardana-controls.org/devel/howto_macros/scan_framework.html#deterministic-scans). 

Thanks @teresanunez for your patience! Sorry you had to wait so long until we finally found some time to deeply analyze the requirements and review your great work!

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -224,10 +224,31 @@ class aNscan(Hookable):
         step["post-step-hooks"] = self.getHooks('post-step')
 
         step["check_func"] = []
-        for point_no in range(self.nb_points):
-            step["positions"] = self.starts + point_no * self.interv_sizes
-            step["point_id"] = point_no
+        try:
+            general_condition = self.getEnv("GeneralCondition")
+            self._gScan.deterministic_scan = False
+        except UnknownEnv:
+            general_condition = None
+        repeated_points = 0
+        for point_nb in range(self.nb_points):
+            step["point_id"] = point_nb + repeated_points
+            step["positions"] = self.starts + point_nb * self.interv_sizes
             yield step
+            if not general_condition:
+                continue
+            while True:
+                try:
+                    condition_macro, _ = self.createMacro(general_condition)
+                    repeat = self.runMacro(condition_macro)
+                except Exception as e:
+                    repeat = False
+                if not repeat:
+                    break
+                repeated_points += 1
+                step["point_id"] = point_nb + repeated_points
+                # avoid moving motors to the same position
+                # step["positions"] = [None] * self.N
+                yield step
 
     def _waypoint_generator(self):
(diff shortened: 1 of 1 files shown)


dschick:
Hi @teresanunez and @reszelaz , many thanks for sharing your work from DESY and this new PR.

Regarding the open questions:

1. I would go for the hook-place `repeat-acq` as this is again more constistent, but of course this rather special hook-place must be wel documented.
2. I think not saving the data should be the default setting, because otherwise it will be tough to get your analysis software running smoothly, e.g. PyMCA
3. Again for consistency I would appreciate, if this feature is working for all kind of step-scan, also `fscan` and `mesh`

PR #1482, opened 2021-01-25
[WIP] Add RoI pseudo counters

reszelaz:
Hi all,

This `TwoDRoI` controller helped me in reproducing #1466 errors and now I think that it may be usefull for others, hence I opened this PR - still in WIP.

First question that we would need to answer is if RoI (Region of Interest) pseudo counter controllers should be redistrributed with the Sardana project or maybe should be a third-party plugin project and be just federated in the plugins catalogue. What do you think? 

When selecting the data type and format of the `RoI` attribute (the `RoI` configuration) I see there two options:
1. Spectrum of integers (2 elements for 1D RoI and 4 elements for 2D RoI
2. String with representation of the spectrum as in point 1.

When thinking about using Sardana with Tango, option 2 has an advantage that it may be a memorized attribute.
To choose between this two options it may be helpful to decide on how to save the RoI configuration in the scan file. This was asked by @dschick (now I can not find in which issue/comment) and we still need to advice on this point.
This decission should be then also applied of the _hardware RoI of dummy 2D controller_.

And finally this PR still misses the `OneDRoI` controller.

### DIFF
--- a/src/sardana/pool/poolcontrollers/RoI.py
+++ b/src/sardana/pool/poolcontrollers/RoI.py
@@ -1,0 +1,103 @@
+##############################################################################
+##
+# This file is part of Sardana
+##
+# http://www.sardana-controls.org/
+##
+# Copyright 2011 CELLS / ALBA Synchrotron, Bellaterra, Spain
+##
+# Sardana is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+##
+# Sardana is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+##
+# You should have received a copy of the GNU Lesser General Public License
+# along with Sardana.  If not, see <http://www.gnu.org/licenses/>.
+##
+##############################################################################
+
+from sardana.pool.controller import (
+    DefaultValue,
+    Description,
+    FGet,
+    FSet,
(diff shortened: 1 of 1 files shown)


dschick:
Thanks @reszelaz for bringing this up.

> To choose between this two options it may be helpful to decide on how to save the RoI configuration in the scan file. This was asked by @dschick (now I can not find in which issue/comment) and we still need to advice on this point.

I think this was discussed in #771

So in general, I appreciate to put the ROI pseudo counter into the Sardana project since it is a very general controller and is also a good example for beginners.

However, I would like to dicuss the intended functionality. As I see it the current implementation simply cuts a ROI out of a 2D image an creates a new smaller 2D image.
1. Regarding the option of having only 2 intergers set as ROI: I think it would be nice to define the axis along which the 1D ROI is set.
2. The current implementation is a simple data duplication, right? So at least at MBI, we do not have this use case. Instead we always integrate/average the rois either to spectra (2D->1D) or to floats (2D->0D).
3. It would be nice to make any implementation of the ROI PC compatible with https://github.com/lavue-org/lavue, maybe @jkotan could comment on this as well.

Best

Daniel

jkotan:
Hi,

in lavue (our online viewer) we've just started to use the RoIs attribute convention designed by @teresanunez  for theLambdaOnlineAnalysis tango server.  Lavue is used to select the RoI regions on the GUI widget and passed to the Analysis server. 

According to the convention RoIs attribute is a spectrum with a number of elements multiplied by 4, i.e. a lot of our users ask for possibility to have more than one RoI sum calculated. Thus, e.g. if you want from your tango server to calculate two RoI sums you provide a 4 x 2 integer spectrum.  It looks like the new attribute 'RoI' is not compatible with the lavue because:
1) its name: `RoIs` -> `RoI`
2) the sardana new attribute may contain only one RoI definition (i.e. four int number).

Issue #1209, opened 2019-10-02
Crash on spock startup on buster (matplotlib?)


reszelaz:
To reproduce you can use [this VM](https://filesender.rediris.es/?s=download&token=b1ef4d3b-b77f-9b62-8e13-909b0f403d39). It has all the Sardana and Taurus dependencies installed according to this [vagrant configuration](https://github.com/sardana-org/sardana-training/blob/master/vagrant/provision.sh). You will need to install Taurus and Sardana on it, for example using:
```
pip3 install --user <url>
```

Then first start `Sardana demo1` server and then start spock:

```
vagrant@buster:~/sardana-training$ spock
MainThread     INFO     2019-10-02 17:09:44,175 TaurusRootLogger: Using PyQt5 (v5.11.3 with Qt 5.11.3 and Python 3.7.3)
Profile 'spockdoor' does not exist. Do you want to create one now ([y]/n)? malloc_consolidate(): invalid chunk size
KCrash: crashing... crashRecursionCounter = 2
KCrash: Application Name = python3.7 path = /usr/bin pid = 1734
KCrash: Arguments: 
y
Available Door devices from buster.localdomain:10000 :
Door_demo1_1 (a.k.a. Door/demo1/1) (running)
Door name from the list? Door_demo1_1
Storing ipython_config.py in /home/vagrant/.ipython/profile_spockdoor... [DONE]
Spock 3.0.2-alpha -- An interactive laboratory application.

help      -> Spock's help system.
object?   -> Details about 'object'. ?object also works, ?? prints more.

IPython profile: spockdoor

Connected to Door_demo1_1
MainThread     INFO     2019-10-02 17:10:33,918 matplotlib.font_manager: Could not open font file /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf
```

here it hangs, and after Ctrl+C:

```
^CTraceback (most recent call last):
  File "/usr/lib/python3/dist-packages/matplotlib/font_manager.py", line 1356, in <module>
    fontManager = json_load(_fmcache)
  File "/usr/lib/python3/dist-packages/matplotlib/font_manager.py", line 891, in json_load
    with open(filename, 'r') as fh:
FileNotFoundError: [Errno 2] No such file or directory: '/home/vagrant/.cache/matplotlib/fontlist-v300.json'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/vagrant/.local/bin/spock", line 10, in <module>
    sys.exit(main())
[...]

PR #1484, opened 2021-01-25
Fix example in docstring of QtSpockWidget

schooft:
I forgot to update the example to reflect the changes in ba692bc9c1c2273da6c2d56b08bb2c9890f870d9.

It would be nice to have a better error message when no kernel is started. Currently, only `...` is shown after pressing enter in a widget without a running kernel.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
@@ -181,6 +181,8 @@ class QtSpockWidget(RichJupyterWidget, TaurusBaseWidget):
         self._door_name = None
         self._door_alias = None
 
+        self.append_stream("Waiting for kernel to start")
+
         self.kernel_manager = SpockKernelManager(kernel_name=kernel)
         self.kernel_manager.kernel_about_to_launch.connect(
             self._handle_kernel_lauched)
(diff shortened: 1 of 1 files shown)


src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
```diff
@@ -146,8 +146,9 @@ class QtSpockWidget(RichJupyterWidget, TaurusBaseWidget):
         from taurus.external.qt import Qt
         from sardana.taurus.qt.qtgui.extra_sardana.qtspock import QtSpockWidget
         app = Qt.QApplication([])
```
reszelaz, line 148:
```suggestion
        app = Qt.QApplication(["qtspock"])
```

I just tried it on Debian Buster with PyQt v5.11.3 and with Qt 5.11.3 and the Qt.QApplication requires a list with a string in order to not crash. We hit the same thing in the past already with spock https://github.com/sardana-org/sardana/issues/1209. I imagine newer versions do not have this problem but maybe it is better to pass a string to be on the safe side.

reszelaz (commented):
Thanks for the PR @schooft ! I agree with the changes you propose but before merging I found one thing that could be improved - see my inline comment.

> It would be nice to have a better error message when no kernel is started. Currently, only ... is shown after pressing enter in a widget without a running kernel.

+1. How would you suggest to implement it?

schooft:
> How would you suggest to implement it?

I pushed a simple approach that should just show a "Waiting for kernel to start" message in the widget until a kernel is started. This could already be enough to point the user to the problem. What do you think?

reszelaz (approved):
Perfect! I tested it without starting the kernel and the "Waiting for kernel to start" info appears in the widget. I think it is a nice improvement.
Merging.
Thanks for your contribution!

PR #1485, opened 2021-01-26
Fix NXscanH5_FileRecorder.addCustomData() with strings

reszelaz:
https://github.com/sardana-org/sardana/issues/771#issuecomment-734743646 reports a bug with saving string _custom data_. This PR fixes it. @dschick, the workaround from https://github.com/sardana-org/sardana/issues/771#issuecomment-736360038 is not necessary anymore. @sardana-org/integrators could you please take a look on this one?

Note that in h5py 3.x they changed some recommendations for storing string data, but this PR still uses the recommendations from h5py 2.x that we used in the past.


teresanunez (approved):

reszelaz:
Thank you @teresanunez for reviewing! Merging.

Issue #1315, opened 2020-03-25
Remove deprecated features in Sardana 3


reszelaz:
Since we are approaching a major version number release we would like to remove some features deprecated long time ago.

This issue is to discuss if we all agree on it.

I have reviewed the whole code and found the following features marked as deprecated.
I think we can remove all of them since the deprecation period is very long.
The _data source_ deprecations (32-37) are maybe too early to be removed - only one release of deprecation warning. But if no one comments here I would remove them as well.

|Number | Deprecated | Deprecated since | Sugested substitute |
|---|---|---|---|
| 1 | sardana.release.version_info | 2.2.0 | sardana.release.version |
| 2 | sardana.release.revision | 2.2.0| sardana.release.version |
| 3 | sardana.release.revision | 2.2.0| sardana.release.version |
| 4 | ParamRepeat | 2011| param_def in lists |
| 5 | GSF generator yields hooks | 2011| GSF generator yields specific hooks e.g. post-acq-hooks|
| 6 | pre-start hook place in ascanct scans | 2018| pre-acq hook place in ascacnt scans|
| 7 | FileRecorder macro hints | Jan16 | ScanRecorder environment variable |
| 8 | Controller.class_prop | 1.0 | Controller.ctrl_properties |
| 9 | Controller.ctrl_extra_attributes | 1.0 | Controller.axis_attributes |
| 10 | Controller.inst_name | 1.0 | Controller.GetName |
| 11 | Controller.SetPar | 1.0 | Controller.SetAxisPar |
| 12 | Controller.GetPar | 1.0 | Controller.GetAxisPar |
| 13 | Controller.GetExtraAttributePar | 1.0 | Controller.GetAxisExtraPar |
| 14 | Controller.SetExtraAttributePar | 1.0 | Controller.SetAxisExtraPar |
| 15 | CounterTimerController._trigger_type | Jan17 (SEP6) | CounterTimerController._synchronizarion |
| 16 | CounterTimerController.PreStartAllCT | 1.0 | CounterTimerController.PreStartAll |
| 17 | CounterTimerController.PreStartOneCT | 1.0 | CounterTimerController.PreStartOne |
| 18 | CounterTimerController.StartAllCT | 1.0 | CounterTimerController.PreStartAll |
| 19 | CounterTimerController.PreStartOneCT | 1.0 | CounterTimerController.PreStartOne |
[...]

PR #1488, opened 2021-01-28
Add script to upgrade mntgrp from Sardana 2 to Sardana 3

reszelaz:
When upgrading one of our beamlines to Sardana 3 we realized that not all measurement group configurations were already converted to use the URIs in the measurement group configuration.

Sardana 3 removed the backwards compatibility as agreed in #1315.

The upgrade script from this PR may be helpful during the migration process.

### DIFF
--- a/doc/source/devel/guide_migration/2to3.rst
+++ b/doc/source/devel/guide_migration/2to3.rst
@@ -45,6 +45,10 @@ Sardana v3 was reduced by long time ago deprecated features. You can
 find a list of them together with the suggested substitutes in this
 `table <https://github.com/sardana-org/sardana/issues/1315>`_.
 
+For migrating the measurement group configurations (non-URI model names) you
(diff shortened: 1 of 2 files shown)


jairomoldes:
I used this script in BL29 at ALBA and seemed to work fine.

reszelaz:
Thanks @jairomoldes for the confirmation!
I will auto-merge.

jairomoldes:
I forgot to mention that this seems not to fix snapshots configuration. However that can be easily fixed by running expconf and applying (even if no changes are made): seems like expconf sets fqdn correctly.

reszelaz:
Yes, I confirm that the snapshot is not covered in this script.
Maybe even this script from the wiki recipies: https://gist.github.com/reszelaz/c20ff2cf8c505cf7622c6e6537219da9 could be used.

PR #1494, opened 2021-02-02
Document how to write 1D and 2D controllers

reszelaz:
This PR adds documentation on how to write 1D and 2D controllers.

Since at the end 1D and 2D are _timerable_ controllers as counter/timer controller, I decided to not duplicate the documentation and just reference to the counter/timer controller documentation and list the differences and additional features.

Similarly the 1D controller documentation does not duplicate the _value referencing_ documentation which is placed in the 2D controller documentation.

These documentation does not follow the entire narration of developing "Springfield" hypothetical hardware controller as it is done for the motor and counter/timer. This would require more time and this PR just adds the strictly necessary information. This way we could advance with the last missing point from #1466.

@sardana-org/integrators could you please take a look on this one? Many thanks!

### DIFF
--- a/doc/source/devel/api/sardana/pool/controller.rst
+++ b/doc/source/devel/api/sardana/pool/controller.rst
@@ -25,6 +25,7 @@
     :columns: 3
 
     * :class:`Readable`
+    * :class:`Referable`
     * :class:`Startable`
     * :class:`Stopable`
     * :class:`Loadable`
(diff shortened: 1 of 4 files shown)


teresanunez (approved):

reszelaz:
Thank you @teresanunez for the review! I proceed with the merge.

PR #1505, opened 2021-02-19
Ensure order of moveables is preserved in Motion object

reszelaz:
Move of motion object sends motors to wrong positions. This happens when using moveables from different Pools and is not 100 % reproducible.

This is due to the fact that we rely on simple dictionaries (their order is not preserved in Python 3.5) when constructing the Motion object from the list of moveables requested by the user. Further move requests on that object assume that the positions correspond to the order of moveables requested at the Motion object creation time. Currently the order may be lost.

Use OrderedDict instead to avoid such problems.

We discovered this bug after migrating one of our beamlines to Python 3, while executing continuous scan (a2scanct). I looks like with Python 2 we were more lucky and we had not suffered it. Before reviewing it, let's wait until @jairomoldes confirms it fixes the issue for him.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/motion.py
+++ b/src/sardana/taurus/core/tango/sardana/motion.py
@@ -30,6 +30,7 @@ __all__ = ["Moveable", "MoveableSource", "Motion", "MotionGroup"]
 __docformat__ = 'restructuredtext'
 
 import time
+from collections import OrderedDict
 
 from taurus.core.util.containers import CaselessDict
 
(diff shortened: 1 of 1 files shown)


jairomoldes:
This fix seems to work fine in BL29 at ALBA

reszelaz:
Thanks @jairomoldes for the confirmation!
@sardana-org/integrators could you please review this one? Many thanks!

13bscsaamjad:
looks good!

13bscsaamjad (approved):

reszelaz:
Thanks @jairomoldes and @13bscsaamjad for the review! I proceed with the merge.

PR #1510, opened 2021-02-24
Fixes for string datasets with h5py >= 3

tiagocoutinho:
Some fixes required for having hdf5 with VDS feature working.
Identified so far:
* if no vds name is given in ValueRefPattern it should default to "dataset" but instead is defaulting to None which raises error
* in some versions of h5py (example: 3.1) you get "bytes" instead of "str" as value for a key which raises an error

A more serious error is that when vds is active for a specific channel, the vds shape is taken from the channel (which gets it from the controller) but the dtype is the same of the original even which is "str" since the channel provides the data source in form of a URI. hdf5 raises an error saying that it cannot create a vds for type "<U" as seen below:
```
spock --profile=myth2
Spock 3.0.3 -- An interactive laboratory application.

help      -> Spock's help system.
object?   -> Details about 'object'. ?object also works, ?? prints more.

IPython profile: myth2

Connected to Door_myth2_1

Door_myth2_1 [1]: _myth2ct 0.1 2   # this is basically a timescan
Operation will be saved in /tmp/mythen2/scan.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #25 started at Wed Feb 24 18:50:55 2021. It will take at least 0:00:00.200000
 #Pt No   test_myth2     dt   
   0      h5file:///tmp/mythen2/myth2_024.h5  0.0392895
   1      h5file:///tmp/mythen2/myth2_025.h5  0.211432
Operation saved in /tmp/mythen2/scan.h5 (HDF5::NXscan)
Scan #25 ended at Wed Feb 24 18:50:55 2021, taking 0:00:00.375989. Dead time 100.0% (motion dead time 0.0%)
```
In spock we get no error or warning but the hdf5 has wrong data. Specifically instead of the channel key we get "_" + channel name and the data is the original list of paths instead of the expected VDS.
Here is the sardana log:

```
SardanaTP.W001 WARNING  2021-02-24 18:50:55,949 NXscanH5_FileRecorder: Could not create a Virtual Dataset. Reason: TypeError("No conversion path for dtype: dtype('<U')")
```

### DIFF
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -397,7 +397,9 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
                         continue
                     uri_groups = group.groupdict()
                     filename = uri_groups["filepath"]
-                    dataset = uri_groups.get("dataset", "dataset")
+                    dataset = uri_groups["dataset"]
+                    if dataset is None:
+                        dataset = "dataset"
                     vsource = h5py.VirtualSource(filename, dataset,
                                                  shape=(dim_1, dim_2))
                     layout[i] = vsource
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -66,7 +66,11 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
     """
     formats = {'h5': '.h5'}
     # from http://docs.h5py.org/en/latest/strings.html
-    str_dt = h5py.special_dtype(vlen=str)  # Variable-length UTF-8
+    try:
+        str_dt = h5py.string_dtype()
+    except AttributeError:
+        # h5py < 3
+        str_dt = h5py.special_dtype(vlen=str)  # Variable-length UTF-8
     byte_dt = h5py.special_dtype(vlen=bytes)  # Variable-length UTF-8
     supported_dtypes = ('float32', 'float64', 'int8',
                         'int16', 'int32', 'int64', 'uint8',
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -366,10 +366,12 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
             # If h5file scheme is used: Creation of a Virtual Dataset
             if dd.value_ref_enabled:
                 measurement = nxentry['measurement']
-                first_reference = measurement[label][0]
-                # in some versions of h5py we get bytes
-                if isinstance(first_reference, bytes):
-                    first_reference = first_reference.decode()
+                try:
+                    dataset = measurement[label].asstr()
+                except AttributeError:
+                    # h5py < 3
+                    dataset = measurement[label]
+                first_reference = dataset[0]
                 group = re.match(self.pattern, first_reference)
                 if group is None:
                     msg = 'Unsupported reference %s' % first_reference
@@ -395,10 +397,7 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
                                             dtype=dd_env.dtype)
 
                 for i in range(nb_points):
-                    reference = measurement[label][i]
-                    # in some versions of h5py we get bytes
-                    if isinstance(reference, bytes):
-                        reference = reference.decode()
+                    reference = dataset[i]
                     group = re.match(self.pattern, reference)
                     if group is None:
                         msg = 'Unsupported reference %s' % first_reference
@@ -406,10 +405,10 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
                         continue
                     uri_groups = group.groupdict()
                     filename = uri_groups["filepath"]
-                    dataset = uri_groups["dataset"]
-                    if dataset is None:
-                        dataset = "dataset"
-                    vsource = h5py.VirtualSource(filename, dataset,
+                    remote_dataset_name = uri_groups["dataset"]
+                    if remote_dataset_name is None:
+                        remote_dataset_name = "dataset"
+                    vsource = h5py.VirtualSource(filename, remote_dataset_name,
                                                  shape=(dim_1, dim_2))
                     layout[i] = vsource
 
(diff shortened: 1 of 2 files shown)


tiagocoutinho:
The first two issues should be solved.

For the third, my proposal is to extend what was done in #1466 to return not only the shape but also the dtype in the controller `GetAxisPar()` method.
Before I go down this rabbit hole, if someone can comment against, in favor or an alternative it could be helpful

reszelaz:
Thanks @tiagocoutinho for testing the VDS support. Just to let you know that this feature is not yet commonly used in the community, I think that MaxIV was interested when we announced it, but I don't know if they finally tried it. Maybe @13bscsaamjad  could comment here.

First of all, I have tried the two issues you describe at the beginning. I confirm they happen with h5py 3.1.0 from conda-forge but not with h5py 2.9.0. Luckliy we have developed some tests when implementing SEP2:), I used them to verify this issue:

h5py 3.1.0:

```console
> pytest sardana/macroserver/recorders/test/test_h5storage.py
[...]
================================================================================== short test summary info ==================================================================================
FAILED sardana/macroserver/recorders/test/test_h5storage.py::TestNXscanH5_FileRecorder::test_VDS - TypeError: cannot use a string pattern on a bytes-like object
FAILED sardana/macroserver/recorders/test/test_h5storage.py::TestNXscanH5_FileRecorder::test_value_ref - TypeError: cannot use a string pattern on a bytes-like object
FAILED sardana/macroserver/recorders/test/test_h5storage.py::test_addCustomData[str_custom_data] - AttributeError: 'Dataset' object has no attribute 'value'
FAILED sardana/macroserver/recorders/test/test_h5storage.py::test_addCustomData[8] - AttributeError: 'Dataset' object has no attribute 'value'
FAILED sardana/macroserver/recorders/test/test_h5storage.py::test_addCustomData[True] - AttributeError: 'Dataset' object has no attribute 'value'
========================================================================== 5 failed, 1 passed, 3 warnings in 0.60s ==========================================================================

```
h5py 2.9.0:

```console
> pytest sardana/macroserver/recorders/test/test_h5storage.py
[...]
=============================================================================== 6 passed, 9 warnings in 0.49s ===============================================================================
```

> For the third, my proposal is to extend what was done in #1466 to return not only the shape but also the dtype in the controller GetAxisPar() method.

To overcome the third problem currently you need to configure the `dtype` in expconf to the data type of the original dataset.
I checked it and the VDS are correctly created. You could do the following to reproduce it:

```
Door_vds_1 [31]: defmeas mntgrpvds02 twod04
Created mntgrpvds02

Door_vds_1 [32]: set_meas mntgrpvds02

Door_vds_1 [33]: set_meas_conf ValueRefPattern h5file:///tmp/foo.h5::bar
ActiveMntGrp = mntgrpvds02

Door_vds_1 [34]: set_meas_conf ValueRefEnabled True
ActiveMntGrp = mntgrpvds02

Door_vds_1 [35]: mkdir -p /tmp/vds

Door_vds_1 [36]: senv ScanDir /tmp/vds/
ScanDir = /tmp/vds/

Door_vds_1 [37]: senv ScanFile vds.h5
ScanFile = vds.h5

Door_vds_1 [38]: twod04.savingenabled = True

Door_vds_1 [39]: expconf  # here you set the dtype to float64

Door_vds_1 [40]: ascan mot13 0 1 1 0.1
Operation will be saved in /tmp/vds/vds.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #4 started at Thu Feb 25 23:07:16 2021. It will take at least 0:00:00.765685
 #Pt No    mot13     twod04      dt   
   0         0      h5file:///tmp/foo.h5::bar  0.33071 
   1         1      h5file:///tmp/foo.h5::bar  0.809083
Operation saved in /tmp/vds/vds.h5 (HDF5::NXscan)
Scan #4 ended at Thu Feb 25 23:07:17 2021, taking 0:00:00.989400. Dead time 79.8% (motion dead time 61.2%)
```

I also agree that it does not have too much sense to have the dtype in the expconf (at least I don't know anyone who uses this). I think it is mostly fixed for a detector. I could imagine only detector's dynamic range configuration that could eventually benefit from changing the dtype and in this case, similarly to the shape, the dtype would be a result of the configuration - so again in favor of removing it from the expconf. But, I'm not very experienced in this matter, so better to hear other opinions before we go for this change.

reszelaz:
I tried to run the above mentioned tests using h5py 3.1.0 with your branch and there are other places to be fixed (now only 4 tests fail) and from the error message I would say that maybe it is the `TestNXscanH5_FileRecorder::test_value_ref` that needs to be adapted:

```
================================================================================== short test summary info ==================================================================================
FAILED sardana/macroserver/recorders/test/test_h5storage.py::TestNXscanH5_FileRecorder::test_value_ref - AssertionError: b'file:///tmp/test.edf' != 'file:///tmp/test.edf' : data does not...
FAILED sardana/macroserver/recorders/test/test_h5storage.py::test_addCustomData[str_custom_data] - AttributeError: 'Dataset' object has no attribute 'value'
FAILED sardana/macroserver/recorders/test/test_h5storage.py::test_addCustomData[8] - AttributeError: 'Dataset' object has no attribute 'value'
FAILED sardana/macroserver/recorders/test/test_h5storage.py::test_addCustomData[True] - AttributeError: 'Dataset' object has no attribute 'value'
========================================================================== 4 failed, 2 passed, 3 warnings in 0.59s ==========================================================================
``` 

The good thing is that with h5py 2.9.0 they still work! On CI (ALBA production stack) it is also OK.

Also, I wonder if it wouldn't be better to limit this PR to only adaptations to h5py 3.1.0 and eventlually refactor the `dtype` concept in another PR - just to simplify review and speed-up its integration.

cpascual:
+1 on removing dtype from expconf

13bscsaamjad:
we are using hdf5 external linking to link to external sources for datasets and not vds because they are not yet supported by matlab hdf5 bindings i think.

reszelaz:
@tiagocoutinho many thanks for proposing this PR and for raising the question of removing the dtype from the measurement group configuration.
I have finished the missing fixes for the h5py >= 3. I used a slightly different approach as it is described in [Strings in HDF5](https://docs.h5py.org/en/stable/strings.html). 
Regarding the second question, I have opened #1561 to decouple it from this PR. 
I also changed the target branch to release-Jan21 so it gets included in the upcoming release. Merging.

PR #1512, opened 2021-03-08
Read value in serial mode

reszelaz:
Value is read in a worker thread, however the read synchronizes with the worker
thread anyway in order to return the read value.

Read it in serial mode to avoid extra complexity of involving a thread.
It is also done the same way for the motor's dial position readout.


13bscsaamjad (approved):

reszelaz:
@13bscsaamjad, again thanks for the review. Merging.

PR #1514, opened 2021-03-08
Open msenvironment shelve file always with 'c' flag

### LINKED ISSUE #1425: dumb shelve backend fails when no environment variables were stored
In order to reproduce it:
1. Start fresh Sardana (or MacroServer) server.
2. Stop the server without storing any environment variables.
3. Start the server again.

In the server logs from point 3 we can observe:

```
MainThread     ERROR    2020-10-01 21:47:27,461 MS_dumb_1.EnvironmentManager: Failed to access environment in /tmp/tango/Sardana/dumb/macroserver.properties
MainThread     DEBUG    2020-10-01 21:47:27,461 MS_dumb_1.EnvironmentManager: Details:
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/msenvmanager.py", line 148, in setEnvironmentDb
    self._env = shelve.open(f_name, flag='w', writeback=False)
  File "/usr/lib/python3.5/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.5/shelve.py", line 227, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.5/dbm/__init__.py", line 85, in open
    raise error[0]("need 'c' or 'n' flag to open new db")
dbm.error: need 'c' or 'n' flag to open new db
MainThread     ERROR    2020-10-01 21:47:27,509 MacroServer/dumb/1: Failed to set environment DB to /tmp/tango/Sardana/dumb/macroserver.properties
MainThread     DEBUG    2020-10-01 21:47:27,509 MacroServer/dumb/1: Details:
  File "/home/zreszela/workspace/sardana/src/sardana/tango/macroserver/MacroServer.py", line 99, in init_device
    macro_server.set_environment_db(self.EnvironmentDb)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/macroserver.py", line 185, in set_environment_db
    self.environment_manager.setEnvironmentDb(environment_db)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/msenvmanager.py", line 148, in setEnvironmentDb
    self._env = shelve.open(f_name, flag='w', writeback=False)
  File "/usr/lib/python3.5/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.5/shelve.py", line 227, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.5/dbm/__init__.py", line 85, in open
    raise error[0]("need 'c' or 'n' flag to open new db")
dbm.error: need 'c' or 'n' flag to open new db
```

This works perfectly with `gdbm` backend is used (on Debian install python3-gdbm).
To workaround, simply created a dummy environment variabled before point 2.

reszelaz:
Empty environment files used with dumb backend fails to
open with 'w' flag. This does not happen for gdbm backend.

Avoid this problem by always using 'c' flag, which according to
docs [1]: "Open database for reading and writing, creating it if it doesn’t exist".

Tested and it does not harm when the environment shelve files is not empty - its
content is preserved.

Fixes #1425.

[1] https://docs.python.org/3/library/dbm.html#dbm.open

### DIFF
--- a/src/sardana/macroserver/msenvmanager.py
+++ b/src/sardana/macroserver/msenvmanager.py
@@ -145,7 +145,7 @@ class EnvironmentManager(MacroServerManager):
                 raise ose
         if os.path.exists(f_name) or os.path.exists(f_name + ".dat"):
             try:
-                self._env = shelve.open(f_name, flag='w', writeback=False)
+                self._env = shelve.open(f_name, flag='c', writeback=False)
             except Exception:
                 self.error("Failed to access environment in %s", f_name)
                 self.debug("Details:", exc_info=1)


teresanunez:
Hi Zibi, I have just looked at this, and for me (in debian10, where we had actually the problems, in debian9 is for us
working) the problem is not solved. If the DB is removed the environment is empty (even if now the property
MacroserverPath is correctly filled).

teresanunez:
Hi Zibi, it works now, I had to remove the wrong files that were previously created and now it works.

teresanunez (approved):

reszelaz:
Thanks for the tests and review. Merging.

jangarrevoet:
Just as a side note, I am not sure if it is sardana that falls back to the path in "/tmp" for the shelve but the path set in the tango database will point to "/tmp" when this was triggered at desy. One has to go in and change the path again otherwise your environment will not persist upon reboot.

teresanunez:
Hi Zibi, thanks for the explanation. In fact, we already set it in our sardana starting up tools to another directory (but
it is a DESY specific directory). The problem Jan hat (and I had before this PR) is that it is set to that  "random" tmp
directory ... But we will look into that. Many thanks.

teresanunez:
Hi Zibi, we look into this. The problem is now solved in the computer where we had it .. but the behaviour was different
to the one in the other computers (included mine) and we don't know the reason.
So, i try to explain it:
-> after applying this PR we remove the old macroserver.properties file
-> starting sardana from scratch the MS environment was, as it should be, empty
-> setting something in the environment two files were created:
macroserver.db (with size different from 0)
macroserver.dat (with size 0)

in the computers without problems the files created were with extensions .dat, .bak, .dir
-> restarting sardana from scratch the spock MS were again empty, and changing something
in the environment was not having any effect in the .dat and .db file.

The solution: we copied the files .dat, .bak, .dir from another computer where everything is working
and removed the two files (.dat, .db) from the directory set in MacroPath.
Now it is working  and the changes in the environment are stored and 
restored when starting again from scratch. The file .db is not created any more.

reszelaz:
With @teresanunez we made some tests on this issue. We found that in python3 the following files are created:
- `gnu` backend: `macroserver.properties` (size > 0); on adding entries no more files are created
- `dumb` backend: `macroserver.properties.dat` (size = 0); on adding entries new files are created: `macroserver.properties.dir` and `macroserver.properties.bak`
- `ndbm` backend: `macroserver.properties.db` (size > 0) ; on adding entries no more files are created

Since at DESY they observed that `.db` file (size > 0) and `.dat` (size = 0) were created it made us think that both `ndbm` and `dumb` backends were somehow involved and it should not be - in the Sardana code we try to use only `gnu` or `dumb`.

Issue #1516, opened 2021-03-11
timer and monitor wrongly set in MG Configuration


jkotan:
Our users reported a problem with setting the master timer and monitor, i.e. if you sent to MeasurementGroup configuration with timer and monitor which is not on the first place e.g.
```python
mg.Configuration = \
{'controllers': {
  'tango://haso228jk:10000/controller/sis3820ctrl/sis3820_exp': {'channels': {
    'tango://haso228jk:10000/expchan/sis3820_exp/1': {
     '_controller_name': 'tango://haso228jk:10000/controller/sis3820ctrl/sis3820_exp',
     '_unit_id': '0',
     'conditioning': '',
     'data_type': 'float64',
     'data_units': '',
     'enabled': True,
     'full_name': 'tango://haso228jk:10000/expchan/sis3820_exp/1',
     'index': 0,
     'instrument': None,
     'label': 'exp_c01',
     'name': 'exp_c01',
     'ndim': 0,
     'nexus_path': '',
     'normalization': 0,
     'output': True,
     'shape': [],
     'plot_axes': ['<mov>'],
     'plot_type': 1,
     'source': 'tango://haso228jk:10000/expchan/sis3820_exp/1/Value'},
    'tango://haso228jk:10000/expchan/sis3820_exp/2': {
     '_controller_name': 'tango://haso228jk:10000/controller/sis3820ctrl/sis3820_exp',
     '_unit_id': '0',
     'conditioning': '',
     'data_type': 'float64',
     'data_units': '',
     'enabled': True,
     'full_name': 'tango://haso228jk:10000/expchan/sis3820_exp/2',
     'index': 1,
     'instrument': None,
     'label': 'exp_c02',
     'name': 'exp_c02',
     'ndim': 0,
     'nexus_path': '',
     'normalization': 0,
     'output': True,
     'shape': [],
     'plot_axes': ['<mov>'],
     'plot_type': 1,
     'source': 'tango://haso228jk:10000/expchan/sis3820_exp/2/Value'}},
   'synchronization': 0,
   'monitor': 'tango://haso228jk:10000/expchan/sis3820_exp/1',
   'timer': 'tango://haso228jk:10000/expchan/sis3820_exp/1',
   'synchronizer': 'software'},
  'tango://haso228jk:10000/controller/dgg2ctrl/dgg2_exp_02': {'channels': {
    'tango://haso228jk:10000/expchan/dgg2_exp_02/1': {
[...]

PR #1521, opened 2021-03-12
respect timer/monitor passed in measurement group configuration

reszelaz:
timer/monitor configuration parameters are still kept in the measurement
group configuration for the backwards compatibility. Respect the user
configuration values (if present) otherwise fill it with the software
master timer/monitor or a random timerable channel (last fallback).

### DIFF
--- a/src/sardana/pool/poolmeasurementgroup.py
+++ b/src/sardana/pool/poolmeasurementgroup.py
@@ -880,7 +880,8 @@ class MeasurementConfiguration(object):
         # timer & monitor are not used
         mnt_grp_timer = cfg.get('timer')
         if mnt_grp_timer:
-            timerable_channels = self.get_timerable_channels(enabled=True)
+            timerable_channels = _get_timerable_channels(timerable_ctrls,
+                                                         enabled=True)
             if mnt_grp_timer in [ch.full_name for ch in timerable_channels]:
                 user_config['timer'] = mnt_grp_timer
             else:
@@ -897,7 +898,8 @@ class MeasurementConfiguration(object):
 
         mnt_grp_monitor = cfg.get('monitor')
         if mnt_grp_monitor:
-            timerable_channels = self.get_timerable_channels(enabled=True)
+            timerable_channels = _get_timerable_channels(timerable_ctrls,
+                                                         enabled=True)
             if mnt_grp_monitor in [ch.full_name for ch in timerable_channels]:
                 user_config['monitor'] = mnt_grp_monitor
             else:
--- a/src/sardana/pool/poolmeasurementgroup.py
+++ b/src/sardana/pool/poolmeasurementgroup.py
@@ -525,6 +525,32 @@ class MeasurementConfiguration(object):
 
         return self._filter_ctrls(timerable_ctrls, enabled)
 
+    def get_timerable_channels(self, acq_synch=None, enabled=None):
+        """Return timerable channels.
+
+        Allow to filter channels based on acquisition synchronization or
+        whether these are enabled/disabled.
+
+        :param acq_synch: (optional) filter controller based on acquisition
+         synchronization
+        :type acq_synch: :class:`~sardana.pool.pooldefs.AcqSynch`
+        :param enabled: (optional) filter controllers whether these are
+         enabled/disabled:
+
+         - :obj:`True` - enabled only
+         - :obj:`False` - disabled only
+         - :obj:`None` - all
+
+        :type enabled: :obj:`bool` or :obj:`None`
+        :return: timerable channels that fulfils the filtering criteria
+        :rtype: list<:class:`~sardana.pool.poolmeasurementgroup.ChannelConfiguration`>  # noqa
+        """
+        channels = []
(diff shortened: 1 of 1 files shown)


jkotan:
Hi Zibi,
One needs to add 
```patch
diff --git a/src/sardana/pool/poolmeasurementgroup.py b/src/sardana/pool/poolmeasurementgroup.py
index 98675a96e..951bed937 100644
--- a/src/sardana/pool/poolmeasurementgroup.py
+++ b/src/sardana/pool/poolmeasurementgroup.py
@@ -868,6 +868,7 @@ class MeasurementConfiguration(object):
         # Fill user configuration with measurement group's timer & monitor
         # This is a backwards compatibility cause the measurement group's
         # timer & monitor are not used
+        self._timerable_ctrls = timerable_ctrls
         mnt_grp_timer = cfg.get('timer')
         if mnt_grp_timer:
             timerable_channels = self.get_timerable_channels(enabled=True)
```
to fix the issue mentioned in #1516

reszelaz:
Thanks for pointing this out. You are right `MeasurementConfiguration._timerable_ctrls` are still not updated at the point where we are looking for the timer/monitor. IIRC this was done on purpose to update the local members at the very end of the setting process. What we wanted to achive is that in the case the new configuration is not correct we raise an exception and leaves the object in an untouched state. Because of that, instead of directly applying your suggestion, I decided to refactor the code a little bit and use another approach in order to avoid prematurely setting the `MeasurementConfiguration._timerable_ctrls`. But I think it should aboid the problem you experienced.

Before you invest your time in testing it again, let's wait for the CI results. When this get's green, go ahaead with your tests (I have only tested very simple cases).

Many thanks for your help!

reszelaz:
@jkotan CI gives us green light already. 

Please remember:

> But if I understands well, what is currently in that PR (now it simply returns in the configuration what was passed when settin it) won't be enough. What you actually would like is that the channels set as master timer/monitor are started as the last ones? In this case we would also need to set them as master timer/monitor for the software synchronization. If you confirm I will make an additional commit.

from https://github.com/sardana-org/sardana/issues/1516#issuecomment-797535167

jkotan:
Great. Thanks! I've just tested and it works it fixes the issue #1516. 

> What you actually would like is that the channels set as master timer/monitor are started as the last ones?

Here, I'm confused. I don't see the reason why timer should be started at the end unless you have big differences between starting different channels.

jkotan:
I've talked to Teresa and we don't need the feature which the channels set as master timer/monitor are started as the last ones

reszelaz:
I thought that you need to specify the master timer/monitor of the measurement group for the need of the ordered start. If this is not the case, what do you use this information for?

I don't know your setup and I may be completely wrong, but I was imagining that in the same measurement group you have some counter channels proceeding from one hardware controller (is it SIS3820?) and a timer channel proceeding from a different hardware controller (is it DGG2?). I imagine that counters count only when the timer gives a gate signal to them. In this case, counters need to be started (armed) for the acquisition in the first order and the timer must be only started as the last one.
Please correct my thinking. Many thanks!

jkotan:
Hi Zibi, 
All the details  can explain to you by Teresa online.
The master timer/monitor configuration is used in our user applications, It is need to know what is the master timer. Also the right order in measurement group is required by some text-file data recorders. Also in spock some our users want to have specific order of channels.  

With counters and timers everything works OK. Teresa has implemented it in sardana.
Our counters (sis3820) are connected to timer (dgg2) via cable. The timer sends to counters a gate signal. The counter don't have to be started, only reset is needed. The general procedure is:
```python
resetCounter()
timer.sampleTime = sampleTime
startTimer()
waitForTimer()
readCounter()
```

reszelaz:
> The master timer/monitor configuration is used in our user applications, It is need to know what is the master timer. Also the right order in measurement group is required by some text-file data recorders. Also in spock some our users want to have specific order of channels.

Thanks for the clarification! Now I see the use cases.
So, you use measurement group configuration, as a place to store information about the master timer channel. 

I also agree that the order of channels used as the mechanism to select the master timer is limiting us. It's better to add a more explicit and flexible way to select it. And actually, we need two master timers: one for the software trigger and gate, and another one for the software start. But, better to open a dedicated issue to discuss it. 


> With counters and timers everything works OK. Teresa has implemented it in sardana.
> Our counters (sis3820) are connected to timer (dgg2) via cable. The timer sends to counters a gate signal. The counter don't have to be started, only reset is needed. The general procedure is:
> 
> ```python
> resetCounter()
> timer.sampleTime = sampleTime
> startTimer()
> waitForTimer()
> readCounter()
> ```

Yes, it works well, I was looking into these controllers' code and I see that you use the `PreStartOne()` to call the reset on the sis3820 and the `StartAll()` to call start on the dgg2. According to [Multiple acquisition synchronization](https://sardana-controls.org/devel/howto_controllers/howto_countertimercontroller.html#multiple-acquisition-synchronization) this is enought to ensure the order that you need. 

With your confirmation that it works as expected, I proceed to auto-merge.

PR #1530, opened 2021-03-18
also apply position formatting to the limits

### LINKED ISSUE #1529: also set position format on the limits
Currently the position format is only applied to the current motor position, which can give unwanted results since the motor limits might have crazy precision.

jangarrevoet:
Should fix #1529 in the wm macro.

Are there other locations where one would like to apply this?

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -369,8 +369,17 @@ class wm(Macro):
             except:
                 val1 = str_fmt % motor.getPosition(force=True)
 
-            val2 = str_fmt % posObj.getMaxRange().magnitude
-            val3 = str_fmt % posObj.getMinRange().magnitude
+            try:
+                val2 = fmt % posObj.getMaxRange().magnitude
+                val2 = str_fmt % val2
+            except Exception:
+                val2 = str_fmt % posObj.getMaxRange().magnitude
+
+            try:
+                val3 = fmt % posObj.getMinRange().magnitude
+                val3 = str_fmt % val3
+            except Exception:
+                val3 = str_fmt % posObj.getMinRange().magnitude
 
             if show_ctrlaxis:
                 valctrl = str_fmt % (ctrl_name)
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -328,6 +328,36 @@ class wm(Macro):
          None, 'List of motor to show'],
     ]
 
+    @staticmethod
+    def format_value(fmt, str_fmt, value):
+        """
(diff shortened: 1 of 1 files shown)


reszelaz (approved):
Thanks for the PR @jangarrevoet!
LGTM, I also tested and works well.

Was it that you simply wanted to get all the three positions being formatted in the same way. So, instead of:

```
Door_zreszela_1 [10]: setvo PosFormat 4

Door_zreszela_1 [11]: wm mot01
                  mot01
User               
 High        1.23456789
 Current         1.2346
 Low        -1.23456789
Dial               
 High               inf
 Current         1.7654
 Low               -inf
```

To get:

```
Door_zreszela_1 [16]: setvo PosFormat 4

Door_zreszela_1 [17]: wm mot01
                  mot01
User               
 High            1.2346
 Current         1.2346
 Low            -1.2346
Dial               
 High               inf
 Current         1.7654
 Low               -inf
```

Or you experienced some other behavior?

One idea to improve it even more, but I repeat that it already works very well, could be to refactor it and make the formatting with a helper function, something like this:

```python
def format_value(fmt, str_fmt, value):
    try:
        fmt_value = fmt % value
        fmt_value = str_fmt % fmt_value
    except Exception:
        fmt_value = str_fmt % value
    return fmt_value

position = motor.getPosition(force=True)
val1 = format_value(fmt, str_fmt, position)
# etc for the max_range and min_range
```
This way, we would remove technical debt and avoid reading the position twice with the request to the hardware.

> Are there other locations where one would like to apply this?

No, I think that the other macros `wa` and `wu` which use the `ViewOption.PosFormat` does not print the limits.

jangarrevoet:
@reszelaz, I have updated the PR with the requested changes.

reszelaz:
Thanks @jangarrevoet for the refactoring. In cba176c I just corrected, I assume, a copy&paste error.

I see you use the numpy style in the docsting of the newly added method. This accelerated a discussion about alternative docstring styles: #1535. I kindly invite you to participate in it if you would like to share your experience with us. 

Meanwhile, I think that this PR is ready to be merged but, due to the numpy docstring, the sphinx build [fails with this error](https://travis-ci.org/github/sardana-org/sardana/builds/763512810): 

```
Warning, treated as error:
/sardana/src/sardana/macroserver/macros/standard.py:docstring of sardana.macroserver.macros.standard.wm.format_value:4:Unexpected section title.
```

What we could do now:
1. Convert the docsting to reST using pyment (I can do it)
2. Wait for #1533 (which I don't know if would be accepted)

What do you prefer?

reszelaz:
Thanks! Since #1533 is already merged, I updated this branch with the necessary changes, and using the napoleon extension, the documentation build does not show any warnings. I'm merging.
Thanks again!

PR #1301, opened 2020-03-02
Revert "Fix Tango inheritance"

reszelaz:
Reverts sardana-org/sardana#1214.
The above commit [broke](https://travis-ci.org/sardana-org/sardana/jobs/657212175?utm_medium=notification&utm_source=email) tests on Travis.
The PR tests were green but the PR branch was very outdatad and the tests were run agains docker image with outdated taurus.
In the current situation the tests unfortuatelly fails. I will automatically merge this one whenever it is green.

PR #1531, opened 2021-03-18
Use LatestDeviceImpl (Device_5Impl)

reszelaz:
This is a retry of #1214 proposed by @tiagocoutinho which was initially merged and then reverted in #1301.


reszelaz:
Merging. This was initially reviewed by myself and got accepted once. The revert was due to a probably spurious failure of the testsuite not reproducible anymore. Thanks @tiagocoutinho!

Issue #1509, opened 2021-02-24
Run post-scan hooks even when the scan fails


13bscsaamjad:
Currently if the scan fails due to some reason, the post-scan hooks does not run. There can be important stuff that needs to be done after the scan, for example, closing the x-ray shutter. This issue asks for doing the implementation to run the `post-scan` hooks even when the scan fails.

Needs implementation discussion for the hooks that only needs to be run when the scan succeeds and not fails.

dschick:
very good issue!
We actually have the same problem, that we need very important hooks at the end of a scan.

However, I am not sureif this is necessary:
> Needs implementation discussion for the hooks that only needs to be run when the scan succeeds and not fails.

I see your point, but it would require some more diversification of the hook-places.
I guess you could also check within your macros in the `post-scan` hook, if they need to be run or not?

13bscsaamjad:
Thanks @dschick. Yes, I agree with your comment and it is not really required for our use case for now. It came to the discussion during the follow-up meeting and I mentioned it if someone has a strong point in favour for implementing this.

reszelaz:
I think it makes a lot of sense to always execute the post-scan hook, even if the scan failed or was stopped/aborted.

I remember we found the current implementation limitting us, I could find only the following comment: https://github.com/sardana-org/sardana/issues/771#issuecomment-736834678. But I think we were discussing it in other occassions as well.

Yes, on the follow-up we found a backwards compatibility issue (I think mostly affecting DESY for the moment) since as @teresanunez mentions in https://github.com/sardana-org/sardana/issues/1471#issuecomment-761212266, they use post-scan hooks.

I would also prefer to leave it to the hook developer to figure out if the scan failed or not. All the necessary information is already gathered: https://github.com/sardana-org/sardana/blob/bc6f55179d4c45f854b8f702ebdca546baab62da/src/sardana/macroserver/scan/gscan.py#L1066
We would just need to add a public API to access to it.

@teresanunez could you please investigate what would be the impact at DESY if we change that?

teresanunez:
Hi Zibi, thanks, I have sent an e-mail to our beamline scientists asking about that. Let´s wait a couple of days
for the answers.

teresanunez:
I have already some feedbacks and they are positive, so I think that it will not be any problem.

jangarrevoet:
@reszelaz, question regarding the "endstatus", is this populated for every macro that inherits Macro or is this just for gscan based macros?
[...]

PR #1538, opened 2021-03-24
Execute post-scan also in case of an exception

reszelaz:
post-scan hook is only executed in case of normal end of the scan.
In some case it will be convenient to execute it also in case the
scan failed or was stopped/aborted. The hook could identify
the scan endstatus from its environment.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1066,10 +1066,9 @@ class GScan(Logger):
             self._env["endstatus"] = endstatus
             self.end()
             self.do_restore()
-            if endstatus == ScanEndStatus.Normal:
-                if hasattr(macro, 'getHooks'):
-                    for hook in macro.getHooks('post-scan'):
-                        hook()
+            if hasattr(macro, 'getHooks'):
+                for hook in macro.getHooks('post-scan'):
+                    hook()
 
     def scan_loop(self):
         raise NotImplementedError('Scan method cannot be called by '


teresanunez:
Hi Zibi,
           I have tested this, but for me it doesn´t work, If I make one scan  with a post-scan hook and I stop it with
ctrl-C the program goes to the point where the hook should be called (so the change you did is fine)  but the macro
doesn´t run ... Could be that this is related to the fact that after a ctrl-c not macros can be run?.
This is in fact the problem I had when I implemented the general_on_stop function (it is in the DESY version of Sardana
and still running), it has to be a function and not a macro because if a macro is stopped none macros can run after that.
                                     Best Regards,
                                                                 Teresa

reszelaz:
@teresanunez, many thanks for the tests!

> Could be that this is related to the fact that after a ctrl-c not macros can be run?.

Exactly! My similar observation in https://github.com/sardana-org/sardana/issues/1509#issuecomment-805746579.

13bscsaamjad:
Tested and works! I think we can merge this and keep the discussion open for how to deal with ctrl+c and hooks!

13bscsaamjad (approved):

reszelaz:
Thanks for the tests @teresanunez and @13bscsaamjad. I'm merging.

reszelaz:
> Tested and works! I think we can merge this and keep the discussion open for how to deal with ctrl+c and hooks!

I agree, let's keep https://github.com/sardana-org/sardana/issues/1509 opened to not forget that the hooks are not being executed after stop/abort.

PR #1326, opened 2020-04-27
Fix regscans points repeat

cpt-majkel:
Hi,
we've discovered a bug in regscans. When starting point of a region is the same as the starting position, scan point is repeated twice (basically one of them is not ommited).  

Example of this bug below (see points 7-8 and 18-19):

```
Door_test_1 [34]: rscan mot01 0.5 0 0.5 2 0 5 -0.5 5 0 5 0.5 5 0 2
Scan #18 started at Mon Apr 27 16:10:16 2020. It will take at least 0:00:15.981031
 #Pt No    mot01      ct01      ct05     epoch       dt   
   0         0        0.5       0.5     1.588e+09  0.174259
   1        0.25      0.5       0.5     1.588e+09  0.891304
   2        0.5       0.5       0.5     1.588e+09  1.60174 
   3        0.4       0.5       0.5     1.588e+09  2.26886 
   4        0.3       0.5       0.5     1.588e+09  2.93647 
   5        0.2       0.5       0.5     1.588e+09  3.61776 
   6        0.1       0.5       0.5     1.588e+09   4.2919 
   7         0        0.5       0.5     1.588e+09  4.96892 
   8         0        0.5       0.5     1.588e+09  5.54238 
   9        -0.1      0.5       0.5     1.588e+09  6.19282 
   10       -0.2      0.5       0.5     1.588e+09   6.8597 
   11       -0.3      0.5       0.5     1.588e+09  7.51575 
   12       -0.4      0.5       0.5     1.588e+09  8.18407 
   13       -0.5      0.5       0.5     1.588e+09  8.84132 
   14       -0.4      0.5       0.5     1.588e+09  9.49642 
   15       -0.3      0.5       0.5     1.588e+09  10.1601 
   16       -0.2      0.5       0.5     1.588e+09  10.8231 
   17       -0.1      0.5       0.5     1.588e+09  11.4903 
   18        0        0.5       0.5     1.588e+09  12.1458 
   19        0        0.5       0.5     1.588e+09  12.7085 
   20       0.1       0.5       0.5     1.588e+09  13.3906 
   21       0.2       0.5       0.5     1.588e+09   14.076 
   22       0.3       0.5       0.5     1.588e+09  14.7484 
   23       0.4       0.5       0.5     1.588e+09  15.4043 
   24       0.5       0.5       0.5     1.588e+09  16.0802 
   25       0.25      0.5       0.5     1.588e+09  16.8118 
   26        0        0.5       0.5     1.588e+09   17.541 
[...]

PR #1540, opened 2021-03-24
Fix1537 move regscan to scans

dschick:
I have moved the `regscan, reg2scan, reg3scan` from th example folder into the `macro/scan.py` module.
At the same time I have renamed the scan macros to `rscan, r2scan, r3scan` in order to be consistent with SPEC and with all other scan macros, e.g. `ascan, dscan, fscan` etc. I also added the `self.motors` property.

I further added the general hook places and fixed some bugs with the `Type` of the `region_nr_intervals` and the missing `self._data` which some lead to the case that the original `regscan`s where not saving their scanned data.

I am also thinking of changing the order of arguments in order to move the `integration_time` to the end, in order to be consistent with all other scan macros. On the other hand, this will require to use spocks advanced syntax with brackets, which on the other hand could be not to bad here in order to improve usability of the macro.


dschick:
arg sorry, I just noticed, that I started this branch off the #1471 PR.
Sorry for the mess, but hopefully it will be soon merged ...

guifrecuni:
> At the same time I have renamed the scan macros to `rscan, r2scan, r3scan` in order to be consistent with SPEC

I'm not a SPEC user, but looking at the documentation of [rscan](https://certif.com/spec_help/rscan.html) and [r2scan](https://certif.com/spec_help/r2scan.html), those are devoted to reflectivity scans with different arguments.

So, I'd say that is better not to use the same names.

dschick:
hi @guifrecuni , thanks for pointing this out.
At least in all SPEC versions that I was using, `rscan` was always linked to region scan.
But in case we do not care so much about SPEC, then `rscan` would be more consistent with `ascan` and `dscan` etc in Sardana.
Also in #1326 @cpt-majkel useses `rscan` in his example.

So of course we can keep the name `regscan` also for backwards compatibility and everyone who like `rscan` better can create an alias macro for it.

guifrecuni:
Searching a little bit further, I've found [another CSS - spec help page](https://certif.com/spec_help/chg6_04.html) where rscan is mentioned for the same purpose. So, it would be fine by me to rename them.
We could also add "regscan aliases" for backwards compatibilty with a deprecation message so we can remove it later on.

dschick:
sure I can do add the regscan alias with deprecation warning

dschick:
in todays Sardana follow-up meeting we decided not to add the alias for `regscan` unless someone really insist on it, because the `regscan` has not been part of the Sardana core macros so far.

for now I updated the order of arguments and moved the `integ_time` to the end.
This requires now to use brackets around the `regions`, which might improve the readability a bit:
`rscan dummy 0 [[1 10] [2 20]] 0.5`

From the current code base, I think that `r2scan` and `r3scan` are not working properly, since they would require individual position/region definitions for each of the motors. I think the `generator` is also just a copy from the single-motor `rscan`.
As these scans are currently not implemented correctly and I am actualy pretty sure that it will happen quite seldom that someone needs an `rscan` for multiple motors, I would simply delete them and readd them if required.

reszelaz (commented):
@dschick many thanks for this PR!
I just reviewed it and it looks OK to me.

Just some comments:

- As we agreed on the follow-up meeting we won't maintain `regscan`, etc. macros and will provide only the `rscan` macros. These use different _macro parameters API_. If there is anyone preferring to maintain the old ones for the backwards compatibility, even if these were just examples, it is a good moment to scream:).

- I tested hooks and these works fine!

- `r2scan` and `r3scan` are working, but I also agree that these are probably not very useful - cause they move all the motors to the same positions. So a question to all, unless someone finds them useful, I would also remove them, what do you think?
    ```
    Door/zreszela/1 [11]: reg2scan mot01 mot02 0.1 0 1 1 2 2
    Scan range is not defined for mot01 and could not be verified against motor limits.
    Scan range is not defined for mot02 and could not be verified against motor limits.
    Operation will be saved in /tmp/foo.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
    Scan #6084 started at Fri May 28 09:50:31 2021. It will take at least 0:00:01.365685
     #Pt No    mot01     mot02      ct01       dt   
       0         0         0        0.1     0.32308 
       1         1         1        0.1     1.17031 
       2        1.5       1.5       0.1     1.97372 
       3         2         2        0.1     2.77406 
    Operation saved in /tmp/foo.h5 (HDF5::NXscan)
    Scan #6084 ended at Fri May 28 09:50:34 2021, taking 0:00:02.896879. Dead time 86.2% (motion dead time 82.0%)
    
    Door/zreszela/1 [12]: reg3scan mot01 mot02 mot03 0.1 0 1 1 2 2
    Scan range is not defined for mot01 and could not be verified against motor limits.
    Scan range is not defined for mot02 and could not be verified against motor limits.
    Scan range is not defined for mot03 and could not be verified against motor limits.
    Operation will be saved in /tmp/foo.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
    Scan #6085 started at Fri May 28 09:52:41 2021. It will take at least 0:00:01.482843
     #Pt No    mot01     mot02     mot03      ct01       dt   
       0         0         0         0        0.1     0.457273
       1         1         1         1        0.1     1.33959 
       2        1.5       1.5       1.5       0.1     2.13361 
       3         2         2         2        0.1     2.93874 
    Operation saved in /tmp/foo.h5 (HDF5::NXscan)
    Scan #6085 ended at Fri May 28 09:52:44 2021, taking 0:00:03.062142. Dead time 86.9% (motion dead time 83.0%)
    ```

- Thanks for fixing the `region_nr_intervals` parameter. This started to fail with the recent numpy versions due to this:

    ```
    ╰─>$ ipython                                                                                                                                                                                                                                                         (sardana) 
    Python 3.9.1 | packaged by conda-forge | (default, Jan 26 2021, 01:34:10) 
    Type 'copyright', 'credits' or 'license' for more information
    IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.
    
    In [1]: import numpy
    
    In [2]: numpy.linspace(0, 1, 1.0)
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-2-6aa401c35593> in <module>
    ----> 1 numpy.linspace(0, 1, 1.0)
    
    <__array_function__ internals> in linspace(*args, **kwargs)
    
    ~/miniconda3/envs/sardana/lib/python3.9/site-packages/numpy/core/function_base.py in linspace(start, stop, num, endpoint, retstep, dtype, axis)
        118 
        119     """
    --> 120     num = operator.index(num)
        121     if num < 0:
        122         raise ValueError("Number of samples, %s, must be non-negative." % num)
    
    TypeError: 'float' object cannot be interpreted as an integer
    
    In [3]: numpy.__version__
    Out[3]: '1.20.0'
    ```
    ```
    ╰─>$ ipython3                                                                                                                                                                                                                                                           (base) 
    Python 3.7.3 (default, Jan 22 2021, 20:04:44) 
    Type "copyright", "credits" or "license" for more information.
    
    IPython 5.8.0 -- An enhanced Interactive Python.
    ?         -> Introduction and overview of IPython's features.
    %quickref -> Quick reference.
    help      -> Python's own help system.
    object?   -> Details about 'object', use 'object??' for extra details.
    
    In [1]: import numpy
    
    In [2]: numpy.linspace(0, 1, 1.0)
    /usr/bin/ipython3:1: DeprecationWarning: object of type <class 'float'> cannot be safely interpreted as an integer.
      #! /bin/sh
    Out[2]: array([0.])
    
    In [3]: numpy.__version__
    Out[3]: '1.16.2'
    
    ```

- I still could not  verify the _macro data_ feature. I think we introduced some regression in the `macrodata` magic command. I will investigate it.

reszelaz:
> * I still could not  verify the _macro data_ feature. I think we introduced some regression in the `macrodata` magic command. I will investigate it.

`macrodata` works fine if there are no hooks attached to the scan macro. But I think it was like this already, so I think there is no regression. This problem should disappear after implementing the _last scan data_ concept introduced in https://github.com/sardana-org/sardana/issues/290#issuecomment-819440784/

reszelaz (approved):
I'm proceeding to merge this PR.

There are still two points which we could change, so please comment:
- provide `regscan` _backwards compatibility_ if necessary e.g. at ALBA we were not using them programmatically (by calls to `Macro.execMacro()`, `Macro.runMacro()`, etc), so for us, it is not necessary.
- remove `r2scan` and `r3scan` - @dschick and myself do not see them very useful

Issue #540, opened 2017-08-09
Python crash pop up when launching the testsuite


mrosanes:
Win7: Pop up when launching the testsuite (sardana 2.3.0)

On Windows 7, when launching the testsuite with  ```python testsuite.py``` many times, it appears an error popup indicating:

```python.exe has stopped working```

mrosanes:
Twelve popups have been opened during a single testsuite execution.

reszelaz:
Info from: https://github.com/sardana-org/sardana/issues/906#issue-348821722

> For example the following tests:
> 
> `C:\Users\IEUser>python -m unittest sardana.macroserver.macros.test.test_scanct`
> 
> works perfectly (well maybe apart of the last message:)):
> 
> ```
> ----------------------------------------------------------------------
> Ran 7 tests in 346.428s
> 
> OK
> Assertion failed: Successful WSASTARTUP not yet performed (..\..\..\src\signaler.cpp:377)
> ```
> 
> But there are several python crash pop-up raised during the test execution.

reszelaz:
We could try this hint: https://github.com/tango-controls/pytango/issues/390#issuecomment-746995003 (the part of the client and the part of the server).

reszelaz:
> We could try this hint: [tango-controls/pytango#390 (comment)](https://github.com/tango-controls/pytango/issues/390#issuecomment-746995003) (the part of the client and the part of the server).

I was trying this, based on [taurus-org/taurus!1185](https://gitlab.com/taurus-org/taurus/-/merge_requests/1185) but it does not help with the python crashes popups when running the Sardana tests. I have tried to call `Util.instance().server_cleanup()` at exit in a DS. It does no harm but it was not fixing the issue either. But from this [comment](https://gitlab.com/tango-controls/pytango/-/issues/390#note_496383071) we can see that it is not cleaning the ZMQ context but just the ORB context so maybe the issue we observe has to do with the ZMQ context not cleaned.

I have naively tried some other things but neither of them worked:
- use `multiprocessing` instead of subprocess (as PyTango's `DeviceTestContext` does) to start DS
- use spawn instead of fork with `multiprocessing.set_start_method`

I will exclude this issue from the Jan21 milestone and will mark the problematic tests to be skipped for Windows so we could correctly execute the rest of the tests during the release process.

reszelaz:
I'm not experienced at all with debugging such problems at Windows, but what I tried is to:
1. Go to the Task Manager when such python process crashes
[...]

Issue #1548, opened 2021-03-30
Jan21 manual release tests: Debian 9


reszelaz:
This is an issue for checking the Jan21 release in Debian 9

This is a check-list of manual tests. It is just orientative. Expand it
at will. This list assumes a clean environment with all Sardana dependencies
already installed and access to a Tango system with the TangoTest DS running.

Hint: this list can be used as a template to be copy-pasted on a release manual test issues

### Installation
- [x] Install Sardana (on Linux from the tar.gz : `pip3 install <tarball_artifact_URL>` (in conda use `pip`) 
  and on Windows from MSI)

### Create testing environment and run testsuite
- [x] Start Pool demo2. In a console do `Pool demo2`.
- [x] Start MacroServer demo2 and connect to the Pool demo2.
  In another console do: `MacroServer demo2`
- [x] Set MacroServer's MacroPath to point to the macro examples.
  In another IPython console do:
  `tango.DeviceProxy('macroserver/demo2/1').put_property({'MacroPath':'<path_to_sardana_installation_dir>/macroserver/macros/examples'})`
  **Note:** Remember to use OS path separator e.g. `/` on Linux and `\` on 
  Windows
- [x] Restart MacroServer e.g. Ctrl+C in the MacroServer's console and
  start it again.
- [x] Create spock profile demo2. In another console do `spock --profile=demo2`
- [x] In spock run `sar_demo` macro.
- [x] Edit `<path_to_sardana_installation_dir>/sardanacustomsettings.py`
  to point to the demo2 door e.g. `UNITTEST_DOOR_NAME = "door/demo2/1"`
- [x] Run testsuite. In another console do `pytest <path_to_sardana_installation_dir>`
  **Note:** On openSuse 11.1 and Windows there are known problems with 
  testsuite. Check previous release comments.

### Test Sardana using Spock and expconf
- [x] Test interactive macros from spock e.g. `ask_for_moveable`, `ask_peak`
  **Note**: On Windows there is printed RuntimeWarning about difference between
  stdin and stdout encoding.
- [x] Execute `umvr` macro and verify that the position updates arrives.
- [x] In expconf configure scan files by setting ScanDir to: `/tmp/` on Linux
  `C:\Users\<username>\tmp` on Windows and ScanFile to: `demo1.h5, demo1.dat`.
- [x] Configure online plot to show counters: On expconf GUI select for all
  the counter channels, Plot Type 'Spectrum' and Plot Axes '\<mov\>'
[...]

Issue #1546, opened 2021-03-30
Jan21 manual release tests: Win10


reszelaz:
This is an issue for checking the Jan21 release in Windows 10

This is a check-list of manual tests. It is just orientative. Expand it
at will. This list assumes a clean environment with all Sardana dependencies
already installed and access to a Tango system with the TangoTest DS running.

Hint: this list can be used as a template to be copy-pasted on a release manual test issues

### Installation
- [x] Install Sardana: `pip install https://ci.appveyor.com/api/buildjobs/ihe5ontu9a569y5y/artifacts/dist%2Fsardana-3.1.0.tar.gz`

### Create testing environment and run testsuite
- [x] Start Pool demo2. In a console do `Pool demo2`.
- [x] Start MacroServer demo2 and connect to the Pool demo2.
  In another console do: `MacroServer demo2`
- [x] Set MacroServer's MacroPath to point to the macro examples.
  In another IPython console do:
  `tango.DeviceProxy('macroserver/demo2/1').put_property({'MacroPath':'<path_to_sardana_installation_dir>/macroserver/macros/examples'})`
  **Note:** Remember to use OS path separator e.g. `/` on Linux and `\` on 
  Windows
- [x] Restart MacroServer e.g. Ctrl+C in the MacroServer's console and
  start it again.
- [x] Create spock profile demo2. In another console do `spock --profile=demo2`
- [x] In spock run `sar_demo` macro.
- [x] Edit `<path_to_sardana_installation_dir>/sardanacustomsettings.py`
  to point to the demo2 door e.g. `UNITTEST_DOOR_NAME = "door/demo2/1"`
- [x] Run testsuite. In another console do `pytest <path_to_sardana_installation_dir>`
  **Note:** To avoid problems with crashing testsuite, it was executed using taurus from [hotfix-4.7.2](https://gitlab.com/taurus-org/taurus/-/merge_requests/1188) branch

### Test Sardana using Spock and expconf
- [x] Test interactive macros from spock e.g. `ask_for_moveable`, `ask_peak`
- [x] Execute `umvr` macro and verify that the position updates arrives.
- [x] In expconf configure scan files by setting ScanDir to: `/tmp/` on Linux
  `C:\Users\<username>\tmp` on Windows and ScanFile to: `demo1.h5, demo1.dat`.
- [x] Configure online plot to show counters: On expconf GUI select for all
  the counter channels, Plot Type 'Spectrum' and Plot Axes '\<mov\>'
[...]

PR #1551, opened 2021-03-30
Release 3.1.0 (Jan21)

reszelaz:
This is a PR to the master branch for the official [Jan21 release](https://github.com/sardana-org/sardana/milestone/8). Final tests (automatic and manual) should be done based on it before accepting the merge.

So, this is a call for help from @sardana-org/integrators and any other person willing to help to test this branch in their respective systems.

As we agreed on the follow-up meetings, these are the systems to be tested thoroughly:

- [x] Debian 9 (#1548)
- [x] Debian 10 (#1547)
- [x] CentOS (#1549)
- [x] Windows 10 (#1546)

The list of the testing platform is quite extensive. If you finally would like to skip some of them, or you think you won't have time to do the tests, IMO it is still fine.

**Outdated - tests on the other platforms can already start:** I will start from Windows and please wait with starting the other tests until this one is done. This is because during the tests I will try to fix https://github.com/sardana-org/sardana/issues/540.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 3.1.0-alpha
+current_version = 3.1.0
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
 serialize = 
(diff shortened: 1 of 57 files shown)


teresanunez:
Hi Zibi, I could try to test debian10, if nobody is already planning to do it.

reszelaz:
Great @teresanunez! Thanks for the offer! You are the first volunteer so I think you can go ahead. Just let me know if you need any help.

PR #1556, opened 2021-04-12
(test) use os.pathsep as recorder paths separator in test.

reszelaz:
Fix test failing on Windows due to using wrong path separator.
Use os.pathsep instead of hardcoding ":" in recorder paths
(test input). For windows the separator must be ";".


teresanunez:
Hi Zibi, I can not test it in Windows, for my computer is fine (debian10), so I agree with the change.

teresanunez (approved):

reszelaz:
@teresanunez  thanks for the review! I'm merging

PR #1557, opened 2021-04-12
Fix measurement group tango test

reszelaz:
Wrong full_name composition makes the test to fail
(surprisingly only on Windows)

TangoDeviceNameValidator is not capable to interpret the
device names containing authority which does does not start
with "//". Add this prefix to avoid the problems.

Sardana does not support Taurus 3. Also remove unnecessary fallback.


teresanunez (approved):
Not tested in Windows, but for debian10 I agree with the change

reszelaz:
@teresanunez , again many thanks for the review! Merging. (also the CI tests works on Debian 9 it is just the docs build on Travis fails due to some problems with dowloading the Docker image)

PR #1558, opened 2021-04-12
(test) Avoid crashes of tests at process exit

reszelaz:
Some combination of tests (when not executed with other tests
using Taurus >= 4.7.1) crash at exit due to [PyTango#390](https://gitlab.com/tango-controls/pytango/-/issues/390).

Register (at class level) atexit hook to call `tango.ApiUtil.cleanup()` in:
- `MacroExecutor` (tango implementation) - used when testing
  macro execution
- MeasurementGroup (tango) tests

Multiple executions of `ApiUtil.cleanup()` does not harm.

### DIFF
--- a/src/sardana/tango/macroserver/test/macroexecutor.py
+++ b/src/sardana/tango/macroserver/test/macroexecutor.py
@@ -25,6 +25,7 @@
 
 import copy
 import time
+import atexit
 import threading
 import PyTango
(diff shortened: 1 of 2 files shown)


reszelaz:
I'm auto-merging since it is necessary for the release process to advance. Feel free to comment in this PR afterward if changes are necessary.

PR #1564, opened 2021-04-17
(test) skip qtspock test_get_value if qtconsole >= 4.4.0

reszelaz:
qtspock's `test_get_value()` fails with newer versions of qtconsole. 
This is because `RichJupyterWidget.blocking_client` (`BaseFrontendMixin.blocking_client`) was removed in [qtconsole#174](https://github.com/jupyter/qtconsole/issues/174) and is not available in qtconsole >= 4.4.0.

Skip the test for these versions.

@schooft do you have other ideas on how to deal with that? Many thanks in advance!

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py
@@ -1,13 +1,14 @@
 import re
 import os
 import tempfile
+import qtconsole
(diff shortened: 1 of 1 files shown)


teresanunez (approved):

reszelaz:
@teresanunez many thanks for the review! I'm merging. 
@schooft feel free to comment, we could still introduce changes after the release if you have other ideas on how to fix it.

PR #1566, opened 2021-04-19
(fix) prompt in qtspock when used with new traitlets

reszelaz:
Setting qtspock's prompt (door's alias) fails when traitlet >= 5.0.0 and Taurus models are used instead of configuration stored in the profile. This is because the traitlets library instead of returning a string door_alias now returns DeferredConfigString object.

Fix it by obtaining a string value stored in this trait.

@schooft, first many thanks for providing unit tests together with the qtspock. These made us realize that with the new versions of the traitlets package the qtspock prompt was not correctly set (currently our CI does not run the tests with the newest versions of the packages and I just run them manually as part of the Jan21 release process).
It looks like I fixed it (to be honest it was not an easy one for me) but it would be nice if you could comment.

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
@@ -180,6 +180,7 @@ class QtSpockWidget(RichJupyterWidget, TaurusBaseWidget):
         self._macro_server_alias = None
         self._door_name = None
         self._door_alias = None
+        self._config_passed_as_extra_arguments = False
 
         self.append_stream("Waiting for kernel to start")
 
@@ -207,6 +208,7 @@ class QtSpockWidget(RichJupyterWidget, TaurusBaseWidget):
 
         if not self.use_model_from_profile:
             if self._macro_server_name and self._door_name:
+                self._config_passed_as_extra_arguments = True
                 extra_arguments.append("--Spock.macro_server={}".format(
                     self._macro_server_name))
                 extra_arguments.append("--Spock.macro_server_alias={}".format(
@@ -273,7 +275,14 @@ class QtSpockWidget(RichJupyterWidget, TaurusBaseWidget):
             self._macro_server_alias = None
 
     def _set_prompts(self):
-        var = "get_ipython().config.Spock.door_alias"
+        # If traitlets >= 5.0.0 then DeferredConfigString is used for values
+        # that are not listed in the configurable classes. Get its value.
+        if self._config_passed_as_extra_arguments:
+            self.kernel_client.execute(
+                "from sardana.spock.config import Spock", silent=True)
+            var = "get_ipython().config.Spock.door_alias.get_value(Spock.door_alias)"  # noqa
+        else:
+            var = "get_ipython().config.Spock.door_alias"
         self._silent_exec_callback(
             var, self._set_prompts_callback)
 
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
@@ -36,6 +36,7 @@ import sys
 import pickle
 import ast
 
+import traitlets
 from IPython.core.profiledir import ProfileDirError, ProfileDir
 
 from taurus.external.qt import Qt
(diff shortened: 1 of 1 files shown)


src/sardana/taurus/qt/qtgui/extra_sardana/qtspock.py
```diff
@@ -273,7 +275,14 @@ def _set_macro_server_name(self, door):
...
     def _set_prompts(self):
-        var = "get_ipython().config.Spock.door_alias"
+        # If traitlets >= 5.0.0 then DeferredConfigString is used for values
+        # that are not listed in the configurable classes. Get its value.
+        if self._config_passed_as_extra_arguments:
+            self.kernel_client.execute(
+                "from sardana.spock.config import Spock", silent=True)
+            var = "get_ipython().config.Spock.door_alias.get_value(Spock.door_alias)"  # noqa
+        else:
+            var = "get_ipython().config.Spock.door_alias"
```
schooft, line 285: I didn't have a deeper look yet, but a quick test indicates that this simpler change might work as well
```suggestion
            var = "str(get_ipython().config.Spock.door_alias)"
```
Would there be a problem with this approach?

reszelaz: Thanks for your comments!

I assume you want to use the simple cast to `str` it unconditionally i.e. when both, passing configuration as Taurus model and in Spock's profile. In this case it fails in the following tests:

```
FAILED workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py::MissingProfileTestCase::test_ipython_prompt - AssertionError: False is not true
FAILED workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py::CorrectProfileAfterRestartTestCase::test_ipython_prompt - AssertionError: False is not true
FAILED workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py::ProfileErrorAfterRestartTestCase::test_ipython_prompt - AssertionError: False is not true
FAILED workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py::QtSpockNoModelTestCase::test_ipython_prompt - AssertionError: False is not true
FAILED workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py::QtSpockNoModelAfterRestartTestCase::test_ipython_prompt - AssertionError: False is not true
FAILED workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py::QtSpockModelMissingProfileTestCase::test_ipython_prompt - AssertionError: False is not true
```

I debugged only the test_qtspock.py::MissingProfileTestCase::test_ipython_prompt and this produces `output_bytes='\'<LazyConfigValue {}>\''` instead of: `output_bytes='<LazyConfigValue {}>'` (in `QtSpockWidget._set_prompts_callback()`)

When I tried it conditionally, only for `self._config_passed_as_extra_arguments=True` then there are still some tests which fail.

schooft: Sorry for not running the tests locally. When I try, the first error I get is
```
Could not create HW object: device door/demo1/1 not defined in the database !
```
I did run `sar_demo` in spock without any error message but I cannot find the door in jive. Any idea?

reszelaz: As you already discovered the QtSpock tests require the [`sar_demo` test environment](https://sardana-controls.org/devel/howto_test/test_run_commands.html#sar-demo-test-environment). 

I just tried it reusing the[ reszelaz/sardana-test](https://hub.docker.com/r/reszelaz/sardana-test/) Docker image (as we do in the CI process) and this worked for me:

```
docker run -d --name=sardana-test -h sardana-test -v /homelocal/zreszela/workspace/sardana:/sardana reszelaz/sardana-test
docker exec sardana-test bash -c "cd /sardana && python3 setup.py develop"
docker exec sardana-test supervisorctl start Pool
docker exec sardana-test supervisorctl start MacroServer
docker exec sardana-test bash -c "xvfb-run -s '-screen 0 1920x1080x24' /bin/bash -c 'pytest /sardana/src/sardana/taurus/qt/qtgui/extra_sardana/test/test_qtspock.py'"
```
Note that `/homelocal/zreszela/workspace/sardana` is the path to my local Sardana Git clone.

Don't hesitate to ask if you still find problems, unfortunately the whole Sardana testing thing is not user-friendly.

reszelaz:
@schooft, I will proceed with the merge cause I think it already improves the situation at the same time not breaking anything and this would allow us to continue with the release process. If you think that this change should be reverted for the release don't hesitate to say so. If you find a more optimal solution we could introduce it after the release.
Many thanks again for reviewing these changes!

PR #1567, opened 2021-04-20
(test) avoid executing Pool's DeleteElement() cmd on Windows

### LINKED ISSUE #540: Python crash pop up when launching the testsuite

reszelaz:
Calling Pool's DeleteElement() command at the end of the tests very often crashes the server on Windows (it may have to do with pushing change event on the Elements attribute just before the server exit, note that this is not problematic on Linux).
This deletion is not necessary cause at the end the device server will be deleted anyway.
Skip execution of this command on Windows.

Fixes #540.

### DIFF
--- a/src/sardana/tango/pool/test/base_sartest.py
+++ b/src/sardana/tango/pool/test/base_sartest.py
@@ -23,6 +23,8 @@
 ##
 ##############################################################################
 
+import os
+
 import PyTango
 import taurus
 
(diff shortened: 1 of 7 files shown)


teresanunez (approved):

reszelaz:
@teresanunez many thanks for the review. Merging.

PR #1568, opened 2021-04-22
(test) increase timeout for qtspock tests

reszelaz:
On Windows it was observed that the qtspock tests that
have timeout set to 5 s sporadically fail.
The ones with the timeout set to 10 s works well.
This happened only when executed within the testsuite,
when executed separately it was not observed.
As an immediate solution, unify all timeouts in all tests
to 10 s.

@schooft hope you don't mind this change. For me, it was easier like this to have a reliable testsuite on Windows. I will auto-merge whenever the CI on Linux turns green but feel free to leave comments here.


schooft:
Sure, no problem. This is only a maximum timeout. Increasing it should not increase the test duration in case the test passed before within the shorter timeout.

PR #1569, opened 2021-04-22
(fix) sardana GUIs launchers on Windows

reszelaz:
During the Jan21 release it was observed that the Sardana GUIs launchers,
when sardana is installed on Windows with pip, does not launch
the applications at all.
In the previous releases Sardana was never tested on Windows when installing with
pip, but with MSI, and in this case the launchers were working correctly.
Since version Jan21, the recommendation is to use Sardana on Windows with pip.
In order to overcome this problem change the GUIs launchers from gui_script to
console_scripts entry points. Similar solution was already applied for Taurus
in taurus-org/taurus#541.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,9 +72,6 @@ console_scripts = [
     "Sardana = sardana.tango:main",
     "sardanatestsuite = sardana.test.testsuite:main",
     "spock = sardana.spock:main",
-]
-
-gui_scripts = [
     "diffractometeralignment = sardana.taurus.qt.qtgui.extra_hkl.diffractometeralignment:main",
     "hklscan = sardana.taurus.qt.qtgui.extra_hkl.hklscan:main",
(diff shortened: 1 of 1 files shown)


teresanunez (approved):
Not tested, but checked that it is done like in Taurus.

reszelaz:
I have done some further tests when using `gui_scripts` vs. `console_scripts` and these are the results:

|  | app run in cmd | app run with `Popen` e.g. `showscan online` in Spock |
| -------- | -------- | -------- |
| MSI with `console_scripts`     | OK     | OK (logs directly in the cmd where Spock runs, the same as on Linux)    |
| pip with `console_scripts`     | OK     | OK     |
| MSI with `gui_scripts`     | OK     | OK (but opens an additional cmd to print logs)     |
| pip with `gui_scripts`     | NOT OK     | NOT OK     |

So, I will proceed to merge this one since it fixes the issue at the same time not breaking the MSI, and IMO, even improving it since we get the same behavior as on Linux for `showscan online`.

PR #1572, opened 2021-04-26
(fix) make PMTV relative move combobox accept only positive numbers

### LINKED ISSUE #1571: Only positive values in PMTV relative move write widget
In the PMTV widget, one can introduce negative values in the relative move write widget. IMHO is ambiguous since the relative movement is commanded with a move negative or move positive buttons - there is no need to add a sign to the _delta_ displacement.

reszelaz:
In the PMTV widget, one can introduce negative values in the relative move write
widget. It is ambiguous since the relative movement is commanded with a move negative
or move positive buttons - there is no need to add a sign to the delta displacement.

Set bottom limit to 0 for the validator. `sys.float_info.min` was also evaluated but it
still allows moving the motor with 0 delta.

Fixes #1571.


cpascual:
LGTM. Merging

reszelaz:
@cpascual , many thanks for the review!

PR #1575, opened 2021-05-03
Fix restoring velocity in software (ascanc) continuous scans

### LINKED ISSUE #1574: dscanc not restoring motor velocity when finished
I do this scan:
```python
dscanc mot01 -1000 1000 0.06 0.1
```

with this motor:
```python
Door[7]: sar_info mot01
Properties:
-----------
      pool  =  Pool_py3_hang_1
    source  =  tango://tbl29.cells.es:10000/motor/motctrl01/1/Position
controller  =  tango://tbl29.cells.es:10000/controller/dummymotorcontroller/motctrl01
interfaces  =  ['Element', 'Object', 'PoolObject', 'Moveable', 'Motor', 'PoolElement', 'Acquirable']
      name  =       mot01
        id  =           2
 full_name  =  tango://tbl29.cells.es:10000/motor/motctrl01/1
      axis  =           1
      type  =       Motor
    parent  =   motctrl01
```

and this measurement group:
```python
Door [9]: %sar_info mntgrp01
Properties:
-----------
             pool  =  Pool_py3_hang_1
physical_elements  =  ['ct02', 'ct04', 'ct03', 'ct01']
       interfaces  =  ['Element', 'Object', 'MeasurementGroup', 'PoolObject', 'Countable', 'PoolElement']
             name  =    mntgrp01
               id  =          30
         elements  =  ['ct01', 'ct02', 'ct03', 'ct04']
        full_name  =  tango://tbl29.cells.es:10000/mntgrp/pool_py3_hang_1/mntgrp01
             type  =  MeasurementGroup
           parent  =  Pool_py3_hang_1
```

reszelaz:
8815c91 introduced a hack to avoid recursive decrease of
velocity in continuous scans. Most probably this hack was needed
to avoid some synchronization problems between the software
continuous scan threads (motion and acquisition are done in
separate threads).
When using Taurus 4 this hack is not applied
cause the attribute's range returns -inf and inf floats
instead of "Not specified" string and we don't fall into the
except clause anymore.
Refactor the `CScan.get_max_top_velocity()` to consider only
Taurus 4 and above case (we don't support Taurus < 4 anymore)
and maintain the hack.

Also, log failed setting of maximum top velocity in continuous scans.

Fixes #1574.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1607,7 +1607,9 @@ class CScan(GScan):
         try:
             motor.setVelocity(v)
         except Exception:
-            pass
+            self.warning(
+                "failed setting max top velocity {} for {}".format(
+                    v, motor.name), exc_info=True)
 
     def get_min_pos(self, motor):
         '''Helper method to find the minimum position for a given motor.
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1551,9 +1551,8 @@ class CScan(GScan):
 
         top_vel_obj = motor.getVelocityObj()
         _, max_top_vel = top_vel_obj.getRange()
-        try:
-            max_top_vel = float(max_top_vel)
-        except ValueError:
+        # Taurus4 reports -inf or inf when no limits are defined (NotSpecified)
+        if np.isinf(max_top_vel):
             try:
                 # hack to avoid recursive velocity reduction
                 self._maxVelDict = getattr(self, '_maxVelDict', {})
@@ -1562,9 +1561,6 @@ class CScan(GScan):
                 max_top_vel = self._maxVelDict[motor]
             except AttributeError:
                 pass
-        # Taurus4 reports -inf or inf when no limits are defined (NotSpecified)
-        if np.isinf(max_top_vel):
-            max_top_vel = motor.getVelocity()
         return max_top_vel
 
     def get_min_acc_time(self, motor):
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1595,10 +1595,22 @@ class CScan(GScan):
             min_dec_time = motor.getAcceleration()
         return min_dec_time
 
-    def set_max_top_velocity(self, motor):
-        """Helper method to set the maximum top velocity for the motor to
-        its maximum allowed limit."""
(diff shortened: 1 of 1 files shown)


reszelaz:
I discovered that there was a difference in the algorithm of restoring the motor's velocity when using Taurus 3 and Taurus 4. @jairomoldes there is a big chance that this will solve the issue you are experiencing now. Could it be that it was not there before you upgrade to Sardana 3 and Taurus 4? 
Anyway, I'm testing this issue in my environment using your testing macro (first trying to reproduce the issue with the develop branch, and later I will try the proposed fix). It would be nice if you could test this fix in your environment, this is where you found it reproducible. 
Also, it would help in the review process if you test it for pseudo motors.
Many thanks!

reszelaz:
@jairomoldes points a bug in this PR. `dscanc` does not work with pseudo motors. I will try to fix it.

jairomoldes:
The pseudo motors fix seems to work fine.

However I found another problem, but this does not seem related to this fix, since it also happens when the patch is not applied. Probably it was already there. The problem is that motors are not moved to their original positions when the scan is cancelled (ctrl+c) while it is returning to the original motor positions (motors are just left whereever they are when you ctrl+c). It works fine when you ctrl+c in the middle of the scan. As I said, I think this is a different bug, not related with this one.

reszelaz:
> The pseudo motors fix seems to work fine.

Thanks, @jairomoldes for the tests!

> However I found another problem, but this does not seem related to this fix, since it also happens when the patch is not applied. Probably it was already there. The problem is that motors are not moved to their original positions when the scan is cancelled (ctrl+c) while it is returning to the original motor positions (motors are just left whereever they are when you ctrl+c). It works fine when you ctrl+c in the middle of the scan. As I said, I think this is a different bug, not related with this one.

In https://github.com/sardana-org/sardana/issues/8 it was agreed that interrupting a dscan (and its continuous scan equivalents) in the middle should return to the initial position. Interrupting after the scan has finished but when the motor is returning to the initial position is slightly different. In this case, the user receives this info message: `Returning to start positions...` before going to the initial position. I do not have a strong opinion, but I think that interrupting this process should not re-try to go to the initial position. Any other opinions?

Anyway, I proceed with the merge, cause as you point, this are unrelated issues.

jairomoldes:
I agree that it makes sense to stop the motors when ctrl+c is pressed while already going back to initial positions. The user is already warned that the motors are going to initial positions: if another ctrl+c is done then it makes sense to stop immediately.

Issue #1043, opened 2019-02-12
Deactivate overshooting correction in ascanct


fbecheri:
Hi all,
Currently at the end of the "ascanct" macro the "correction overshooting" is performed.
We should be able to activate or deactivate this correction according to a parameter.
It is important, for example,
* To optimize the round trip time.
* To not make extra movements when the mechanics is not too good.

Regards,
Fulvio

reszelaz:
Will be fixed when https://github.com/sardana-org/sardana/pull/1576 gets integrated.

PR #1576, opened 2021-05-04
Allow to disable the overshoot correction on continuous scan

rhomspuron:
Add environment variable ScanOvershootCorrection to activate or deactivate this procedure. By default if the variable is not defined the system will do the correction. 

Fixes https://github.com/sardana-org/sardana/issues/1043


reszelaz (approved):
Thanks for the PR @rhomspuron!
There was an issue with `ascanc` but it is already fixed in 56e49c419f.
I also added the `ScanOvershootCorrection` environment variable to the catalogue.
Also, this PR fixes an issue with the overshoot correction velocities in `ascanc` - when there are no maximum velocity limits it at least uses the velocities configured before the scan.

PR #1577, opened 2021-05-05
fix: do not reserve instruments in GScan

reszelaz:
Today @tiagocoutinho reported that scans, when get stopped, stop all the instruments present in the Pool. Then in the Spock we could see something like this:

```
Stopping /monitor reserved by dscanc
Stopping /mirror reserved by dscanc
Stopping /slit reserved by dscanc
```

GScan reserve instruments for the scanning macro when it simply
looks for the instruments information e.g. full names. This makes
the MacroExecutor to stop the instruments when the macro is
being stopped as it would do with any other reserved object.

Do not reserve instruments cause it does not make sense to stop
them when stopping the macro.

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -720,8 +720,8 @@ class GScan(Logger):
                 except Exception:
                     instrument = ''
             try:
-                instrumentFullName = self.macro.findObjs(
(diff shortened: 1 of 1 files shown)


tiagocoutinho (approved):
LGTM

reszelaz:
Thanks @tiagocoutinho for the review! Please be aware that I discovered another bug when stopping scans, it is described in #1578.

PR #1580, opened 2021-05-07
Improve installation instructions

### LINKED ISSUE #727: Improve installation instructions & automation for windows
This is a spin-off from #724, in order to focus on the discussion of the best way to provide instructions for sardana installation on Windows, which is [also related to how we do the installation on our CI scripts](https://github.com/sardana-org/sardana/issues/724#issuecomment-376810492):

@reszelaz [wrote](https://github.com/sardana-org/sardana/issues/724#issuecomment-378298555)
>> you should publish this in the sardana documentation.

>I totally agree, we should have it somewhere public. But for the moment I don't have any clear idea where to put it. Just to open the brainstorming:
>  -  DirectlyIn the documentation
>  -   In the release manual test markdown document
>  -  As powershell script in the CI directory

IMHO, the best option is to publish it as a powershell script in the CI dir. Then both the vagrant file and the Appveyor.yml could use it. And **also the "getting started" section of the docs could link to it**.

Now regarding to implementation details, I have two comments:

- most of this powershell script could be converted to a single pip call with a requirements file, which would be neater IMHO
- it would be good to **not** to base our recipe on pythonxy (it is too heavy for CI, and does not seem to be updated anymore). The problem is that then we need to install some stuff that is not available in PyPI (PyQt4, Qt4?, PyQwt5?) . For this we could either try to do it "manually" or we could use conda (AFAIK, pytango is now available for windows in a conda channel).

reszelaz:
Improve installation instructions in the docs:
- add conda option
- recommend pip on Windows
- update Debian installation
- rewrite dependencies section

Fixes #727

### DIFF
--- a/doc/source/users/getting_started/installing.rst
+++ b/doc/source/users/getting_started/installing.rst
@@ -50,31 +50,6 @@ You can test the installation by running::
 (see more detailed instructions in `this step-by-step howto
 <https://sourceforge.net/p/sardana/wiki/Howto-Sardana-on-Debian8/>`__)
 
-
-Windows
--------
-
-#. Download the latest windows binary from https://github.com/sardana-org/sardana/releases
-#. Run the installation executable
-#. test the installation::
-
-       C:\Python35\python3 -c "import sardana; print(sardana.Release.version)"
-
-Windows installation shortcut
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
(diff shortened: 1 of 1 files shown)


reszelaz:
Now I realized how many commits I generated.. I suggest squash before merging :)

teresanunez (approved):

reszelaz:
@teresanunez , many thanks for the review, merging!

Issue #1581, opened 2021-05-07
PoolAcquisition keeps polling the state of channels even after aborting the acquisition


13bscsaamjad:
It is a side-effect of poorly-implemented controller and/or the underlying device of the controller. But it renders Pool process useless until restarted.

### Steps to reproduce
- make experimental channel always return Moving in State calls.
- run any scan and then abort it.
- the pool will still be polling on the channel state and will never recover even after abort because its doing so in wild loop

### Suggested solution
exit condition from state polling based on the scan abortion

reszelaz:
This should be solved by #1582

PR #1582, opened 2021-05-10
exit condition on abort for state polling of channels

13bscsaamjad:
Solves #1581. Only aborting sw and hw acquisitions for now. Can abort the synch as well.


reszelaz:
I have made the first tests. I used an example of a _buggy_ dummy counter/timer controller which takes 3 s to complete `StopOne()` and does not really stop the acquisition and which takes 3 s to complete `AbortOne()` which finally stops.

I tried it in two architectures:
- One with a single Sardana server. This one fails but most probably due to some other issues...
- Another one with a separate Pool and MacroServer. This one looks better!

What I observed when using this PR is, that in this particular case, we lose the end acquisition event (MOVING -> ON). This could be easily explained since we now exit from the poll loop as soon as the measurement group was aborted - not waiting for the elements to finish the aborting process, hence there is no change to emit the last state event.

Furthermore, I observed that it makes that the first subsequent acquisition does not push state events (neither MOVING nor ON) on the experimental channel at all. Later acquisitions push events correctly. 

Together with @13bscsaamjad  we were thinking about an alternative solution and maybe instead of simply ending the poll loop on abort, we could add a 3rd type of request to be sent to an element (the first two are: stop and abort) called _release_. This one would be an explicit request to abandon the action's poll loop. This request could be sent with a 3rd Ctrl+C in Spock introduced in https://github.com/sardana-org/sardana/pull/1256. What do you think?

We found other issues related to this PR. Solutions to these issues were already proposed in: #1591 and #1592.

reszelaz:
Together with @13bscsaamjad we advanced with this PR and I think it is ready for the review and eventual merge.

We consider the following three cases:
1. Element which takes some time to abort and the abort request is blocking
2. Element which takes some time to abort and the abort request is not blocking
3. Element which is hung and will never abort

> Together with @13bscsaamjad we were thinking about an alternative solution and maybe instead of simply ending the poll loop on abort, we could add a 3rd type of request to be sent to an element (the first two are: stop and abort) called release. This one would be an explicit request to abandon the action's poll loop. This request could be sent with a 3rd Ctrl+C in Spock introduced in #1256. What do you think?

Since there were no other comments, we made a quick evaluation of the following options:
a. Simply exit the action's loop on abort (as it was initially implemented by @13bscsaamjad in this PR).
b. Add a static timeout which would start counting on the abort request and exit from the action's loop when it expires.
c. Add an explicit release request as described in the above-quoted comment.

We discarded option a. because it does not work with case 2.
We had chosen option c. over option b. because is more flexible and leaves the decision to the user on whether and when to release the operation. So it is a kind of *dynamic* timeout decided by the user for each case e.g. after visual inspection of the hardware if it is really hung.

Just to complement this discussion, note that all the options work well with case 1 thanks to this code https://github.com/sardana-org/sardana/blob/68316f982e2e1ba0a4ef87547d826876133b48f0/src/sardana/tango/pool/PoolDevice.py#L125

We do not have a strong opinion if this code is correct, but for the moment we do not propose to change it.

In summary, I have added:
- the release element and release action concepts to the Sardana core
- `Release` Tango command to Pool element devices
- `release()` method to the Taurus extensions
- macro release will automatically release the element hung on aborting (3rd Ctrl+C)

The following sketch was helpful to me when working on this, I upload it for future needs.

Note, I also discovered that the current implementation of stopping, aborting, and the newly added releasing elements is not perfect. It will work correctly only for the first element which is hung. So, for example, a continuous scan where both measurement group and moveables hung (I think very unfavorable coincidence and quite unlikely) won't be correctly released. To overcome this we would need to execute Stop/Abort Tango command in an asynchronous way and then collect the replies and wait for the states to change to different than Moving instead of this code: https://github.com/sardana-org/sardana/blob/68316f982e2e1ba0a4ef87547d826876133b48f0/src/sardana/macroserver/msmacromanager.py#L1445

Also note, that I used a separate Pool and MacroServer for testing it and I tested:
- motion of physical motors
- motion of motor groups
- motion of pseudo motors
- acquisition of counter/timers (requires #1595)
- acquisition of measurement group

teresanunez:
Many thanks for the implementation. The decision you took is fine to me. There is something related to that that came up in
one of our beamlines. If they have defined a general_on_stop function (the DESY implementation for calling a function when
the scan is stopped, for example with a first ctrl-C) this function is not called if a second ctrl-C is executed. It seems that some
of the users can not wait for a first ctrl-C and execute immediately a second one, causing problems because this general_on_stop function is not called. We have to take this into account when implementting this general on_stop feature officially in Sardana.

reszelaz:
@teresanunez, many thanks for the comment! I imagine the _general on stop feature_ to be implemented with the hooks mechanism. After fixing https://github.com/sardana-org/sardana/issues/10 this should be possible. The hook could be placed either before or after `Macro.on_stop()` execution. Then, the 2nd Ctrl+C (abort) if issued while still waiting until the reserved elements are stopped (or maybe waiting because one of them is hung) would prevent `Macro.on_stop()` or _general on stop feature_ to be executed according to this [design](https://github.com/sardana-org/sardana/pull/1256#issuecomment-843180578). 
If the 2nd Ctrl+C (abort) is issued during `Macro.on_stop()` or _general on stop feature_  then it these methods will be aborted. If we don't like this behavior we could change it.

13bscsaamjad (commented):
tested and works!

reszelaz:
Thanks for the review @13bscsaamjad . Merging.

Issue #1077, opened 2019-03-11
Servers does not stop correctly on Windows


reszelaz:
Sardana servers e.g. Pool, MacroServer does not stop correctly on Windows. They leave the device server process running (python) attached to the cmd and the device exported flag maintains as True in the Tango database:
![image](https://user-images.githubusercontent.com/6735649/54135939-0fb77d80-441b-11e9-86f2-355f766150c9.png)

I use the following code to reproduce it (note that the demo1 Pool must be already defined in the database):

```python
executable = 'C:\\Users\\IEUser\\AppData\\Roaming\\Python\\Scripts\\Pool.exe'
s = ProcessStarter(executable, 'Pool/demo1')
devname = 'pool/demo1/1'
s.startDs(wait_seconds=30)
try:
    print('Is running:', s.isRunning())
    print("ping:", PyTango.DeviceProxy(devname).ping())
except Exception as e:
    print(e)
s.stopDs()
s.cleanDb(force=False)
```

Which gives the following output:

```
MainThread     INFO     2019-03-11 08:28:14,526 Starter: Starting server Pool/demo1...
MainThread     INFO     2019-03-11 08:28:30,187 Starter: Server Pool/demo1 has been started
Is running: True
ping: 0
MainThread     INFO     2019-03-11 08:28:30,191 Starter: Stopping server Pool/demo1...
MainThread     WARNING  2019-03-11 08:28:40,196 Starter: Server Pool/demo1 did not stop within 10 seconds
MainThread     WARNING  2019-03-11 08:28:40,197 Starter: Pool/demo1 was not registered by this starter. Not removing. Use ProcessStarter.cleanDb(force=True) to force cleanup
```

However when I change to not use the executable and run it via python the device server process ends up correctly. The output is exactly the same.

To try it this way I need to change the executable variable in the aboce code to: `"C:\\Users\\IEUser\\AppData\\Roaming\\Python\\Scripts\\Pool-script.py"` and modify the `taurus.core.tango.starter.ProcessStarter.start` method to insert `'python'` string at the beginning of args list.

This makes me think that there is something wrong with the executables created for console_scripts.

Note that this problem may be the root cause of #540.

reszelaz:
In the case of the MacroServer the crashes at exit were fixed with #1559. Now Jan21 release tests I can not reproduce the issue with the Pool. Closing.

Issue #724, opened 2018-03-27
windows python crash after exiting spock


dschick:
Hej,

when I exit spock under windows 10, python always crashes with the following error message in the spock terminal

```
MainThread     INFO     2018-03-27 12:31:28,565 TaurusRootLogger:
*********************
< Deprecation Counts (525):
        407 * "_get_value is deprecated since 4.0. Use .rvalue instead"
        67 * "getHWObj is deprecated since 4.0. Use getDeviceProxy() instead"
        25 * "getState is deprecated since 4.0. Use .stateObj.read().rvalue [Tango] or .state [agnostic] instead"
        25 * "getSWState is deprecated since 4.0. Use state instead"
        1 * "getStateObj is deprecated since 4.0. Use .stateObj [Tango] or .factory.getAttribute(state_full_name) [agnostic] instead" >
Assertion failed: Successful WSASTARTUP not yet performed (..\..\..\src\signaler.cpp:377)
```

cpascual:
This part is normal (just debugging messages from the compatibility layer for taurus3)

```
< Deprecation Counts (525):
        407 * "_get_value is deprecated since 4.0. Use .rvalue instead"
        67 * "getHWObj is deprecated since 4.0. Use getDeviceProxy() instead"
        25 * "getState is deprecated since 4.0. Use .stateObj.read().rvalue [Tango] or .state [agnostic] instead"
        25 * "getSWState is deprecated since 4.0. Use state instead"
        1 * "getStateObj is deprecated since 4.0. Use .stateObj [Tango] or .factory.getAttribute(state_full_name) [agnostic] instead" >
``` 

But regarding the last line  (`Assertion failed: (...)`) I do not know where it comes from or if it is really problematic.

dschick:
despite of these messages in the terminal there is also a popup claiming that python has crashed and needs to be closed.
Nothing serious but in order to convince my colleagues to use sardana it would be helpful to get rid of most of the dramatic error messages ;)

thanks for the quick reply

cmft:
We had seen this kinds of problems in some Windows setups. In order to try to reproduce these errors can you report us which version (python, tango, qt, pytango, taurus and sardana ) and how did you install them?.

dschick:
hi @cmft ,
I used WinPython 2.7.13.1 from [here](https://sourceforge.net/projects/winpython/files/WinPython_2.7/2.7.13.1/)

[...]

PR #1583, opened 2021-05-13
bump taurus requirement to >= 4.7.1.1 on windows

reszelaz:
#724, #540 requires taurus >= 4.7.1
#1077, #540 requires taurus >= 4.7.1.1

Tested:
- on Linux, installing with pip (in a conda environment) where taurus 4.7.0 was available and pip considered this version as satisfactory
- on Windows, installing with pip (on Python installed with the exe installer from python.org) where taurus 4.7.0 was available and pip considered this version as **not satisfactory** and performed the upgrade

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ install_requires = [
     'PyTango>=9.2.5',
     'itango>=0.1.6',
     'taurus>=4.7.0',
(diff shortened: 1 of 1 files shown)


teresanunez (approved):
I can not test it (windows) but the change looks fine to me.

reszelaz:
@teresanunez many thanks for the review. Merging.

PR #1584, opened 2021-05-14
Add "What's new?" section to docs

reszelaz:
The need for this section was identified in https://github.com/sardana-org/sardana-followup/issues/32.
The format was not decided at that time. The last release provided it in the [GitHub release description](https://github.com/sardana-org/sardana/releases/tag/3.0.3). This time I tried another approach and add it to the docs directly.
IMHO it is better cause it unifies the news with the docs making them easier to discover.

Here is how it looks:

sardana_news


teresanunez (approved):

teresanunez:
Hi @reszelaz, many thanks for implementing this. I think it is really good to have it.

reszelaz:
Thanks to you for the review! Merging.

PR #1588, opened 2021-05-18
Error handling for missing libs

### LINKED ISSUE #1587: Handle missing controller libraries.
Right now, if the controller defined in Sardana is missing the definition class/library, it does not get initialized and is not properly handled for feedback to the users. Reading the `Status` for such controllers or respective axes give ugly messages like `NoneType object has no attribute PreStateAll`. This is very confusing for the end users and for support staff.
The Status of the controller comes like this:
image
And of some axis of this controller it outputs empty Status.
These messages get propagated to spock as well.

13bscsaamjad:
Trying to fix #1587. 
Fixes:
1. `State`, `Status` handling of controllers/axes when missing libraries.
2. `_verify_controllers` only configures controllers that are initialized properly and thus avoiding the `DevFailed` exception when starting the server.
3.  Before this fix, the MeasurementGroup would report `On` without any information about the elements it is composed of. `_calculate_states` for basegroupelement gets called with {} as well other than None, this is probably an overlook and checking for this makes the measurement group to reflect the status of its components. This still does not fix below issue which needs more discussion.
* The `State`, `Status` resolution of experimental channels is passive, i.e., it is not resolved until these methods are called from a client. This is a problem for `State`, `Status` resolution of the MeasurementGroup which is composed of these channels. If one reads the `State`, `Status` of MeasurementGroup it reports `UNKNOWN` with `None` status for all elements.


src/sardana/pool/poolmeasurementgroup.py
```diff
@@ -1384,3 +1386,18 @@ def stop(self):
...
         self._pending_starts = 0
         PoolGroupElement.abort(self)
+
+    # -------------------------------------------------------------------------
+    # utils
+    # -------------------------------------------------------------------------
+
+    def _verify_controllers(self, cfg):
+        pool = self.pool
+        for ctrl_name, _ in list(cfg['controllers'].items()):
```
johanfforsberg, line 1396: No real point in using `.items()` here, and since you're not changing `cfg["controllers"]` you can just:
```python
for ctrl_name in ctrl["controllers"]:
```

src/sardana/pool/poolcontroller.py
```diff
@@ -341,8 +343,9 @@ def _create_controller(self):
 
     def _init(self):
         if self._ctrl_info is None:
-            if self._lib_info is not None:
-                self._ctrl_error = self._lib_info.get_error()
+            self._ctrl_error = f"{self.get_class_name()} controller class NOT found in library {self.get_library_name()}."
+            if self._lib_info is not None and self._lib_info.get_error() is not None:
+                self._ctrl_error += "\n" +self._lib_info.get_error()
```
johanfforsberg, line 348: Missing space after "+"

src/sardana/pool/poolcontroller.py
```diff
@@ -84,6 +84,8 @@ def get_ctrl_error_str(self):
         err = self._ctrl_error
         if err is None:
             return ""
+        if type(err) == str:
```
johanfforsberg, line 87: You should probably use `isinstance(err, str)` here, it's recommended by PEP8 and supposedly also faster :)

johanfforsberg (commented):
This looks good to me, I had a few comments about style.

src/sardana/pool/poolbasegroup.py
```diff
@@ -100,7 +100,7 @@ def _calculate_states(self, state_info=None):
         none, unknown = set(), set()
         fault, alarm, on, moving = set(), set(), set(), set()
         status = []
-        if state_info is None:
+        if state_info is None or not state_info:
```
johanfforsberg, line 103: None evaluates as false, so `if not state_info:` should be eqiuvalent.

reszelaz (commented):
Hi @13bscsaamjad and @johanfforsberg,

I partially reviewed it and corrected some minor things in 1249515 and 11fe276
For the moment I tested with a motor when its controller library is missing. Thanks to your PR, we get this nice information:
```
Door/zreszela/1 [1]: motctrl01.state
-------------------> motctrl01.state()
        Result [1]: tango._tango.DevState.FAULT

Door/zreszela/1 [2]: motctrl01.status()
        Result [2]: 'DummyMotorController controller class NOT found in library DummyMotorController.py.'

Door/zreszela/1 [3]: mot01.state
-------------------> mot01.state()
        Result [3]: tango._tango.DevState.FAULT

Door/zreszela/1 [4]: mot01.status()
        Result [4]: 'mot01 is in Fault\nController is not initialized!'
```
Here an explanation to you regarding the f-string formatting. Sardana still needs to support Python 3.5 so we can not use this feature yet. At ALBA we still run on Debian 9  which comes with Python 3.5. 

Tomorrow I will continue with the measurement group part.
Thanks a lot!

reszelaz (commented):
I've tested the measurement group part.

The `set_configuration_from_user()` is prevented only when the measurement group object is being constructed. This is fine for the measurement groups whose _configuration_ was never changed. But whenever one write the `Configuration` Tango attribute e.g. creating measurement group from the `expconf`, since it is a memorized attribute, then it will be also reestablished on the server startup, consequently calling `set_configuration_from_user()` again. If we follow the strategy of preventing the `set_configuration_from_user()` calls, then we would also need to call it here (so maybe I would move the check into this method).

I also observed that to reestablish the `IntegrationTime` Tango memorized attribute we also need information about the controllers. There is certainly some problem with this implementation.

I see that this PR fixes problems with the measurement group on startup. Even so, I was not able to list the measurement groups with `lsmeas`, should this already work?

Finally, in PR description you mention that there are some aspects to be discussed, more precisely about the state evaluation. Just let me know when you have time and we could make a short meeting.

13bscsaamjad:
Thanks for testing.

I could not understand fully the memorized issue that you pointed; we should talk about it.

`lsmeas` should work, but it should not list the bad measurement groups. I think this is alright, or we would have to extend the MacroServer/spock to point-out to the bad measurement groups somehow.

Yes, the state evaluation of the measurement group; but that has been fixed with #1591

reszelaz (approved):
Today with @13bscsaamjad we addressed the issues reported in https://github.com/sardana-org/sardana/pull/1588#pullrequestreview-678580483.
In summary:
- verification whether controllers are _online_/_offline_ was moved inside of `MeasurementGroup.set_configuration_from_user`. This correctly treats applying default configuration at the initialization time and applying memorized configuration value.
-  measurement group's configuration can be read even if there are _offline_ controllers and therefore it is possible to list the measurement group with the `lsmeas`
- Problems with restoring the memorized value of IntegrationTime should be solved whenever we agree on making this attribute **not** memorized.

With this it is ready to be merged. @13bscsaamjad thanks again for your contribution and @johanfforsberg for the review.

Issue #1316, opened 2020-03-25
MeasurementGroup state is not evaluated if not acquiring


reszelaz:
Measurement group State attribute does not read state from the involved controllers when measurement group is not acquiring. Note that similar architecture of a pseudo motor reads state from the involved controllers.

This is because the MeasurementGroup action - `PoolAcquisition` is composed from multiple acquisition actions. The top level acquisition action never fills its `_pool_ctrl_dict` and this dict is used when reading the state in the `PoolAction._raw_read_state_info_serial`.

I think this is not a regression but it never worked (I looked at the code from 2011 and the _pool_ctrl_dict was neither filled) but I have not tested it with such an old release.

13bscsaamjad:
moreover, i think scan framework should atleast verify once before starting the measurement group if all the channels are in the right state (standby or on), and if they are not it should not continue with the scan and throw an exception. i have seen that if it tries to start the scan anyway it messes up with the events completely and then it breaks Sardana with `transient corba exceptions` and the only way to recover is to restart the Sardana.

reszelaz:
@13bscsaamjad, now I read again your above comment and I think that #1591 only partially fixes it.

Now we will prevent MeasurementGroup start if the MeasurementGroup state != Moving. Which could be On, Alarm or Fault:
https://github.com/sardana-org/sardana/blob/256aca955871e3e64bd5d9507dfada570815b3c3/src/sardana/taurus/core/tango/sardana/pool.py#L481
I think we should allow start in case of On and Alarm. What do you think?

13bscsaamjad:
I think so too `On` and `Alarm` are good states to start from.

reszelaz:
> I think so too `On` and `Alarm` are good states to start from.

Proposed in https://github.com/sardana-org/sardana/pull/1594.

PR #1594, opened 2021-05-25
Allow to start PoolElement only when it is in On or Alarm states

reszelaz:
PoolElement can can try to start when its state is different than
Moving. This tries to start elements in Fault state as well
what can hang the server, for example, in the case of
the measurement group. See discussion in #1316. Allow to start
the PoolElement only if it is in On or Alarm state.


teresanunez (approved):

reszelaz:
@teresanunez, many thanks for the review. Merging.

PR #1595, opened 2021-05-26
Fix stop/abort element in ct macro when used with channels

reszelaz:
Independent acquisition of experimental channels done with `ct`
macro can not be stopped/aborted. This is because the channel is
not reserved by the macro. Reserver it by using `Macro.getObj()`
call.


teresanunez (approved):

teresanunez:
Just checked and it nicely works. Many thanks.

reszelaz:
@teresanunez, many thanks for the review. Merging.

Issue #1528, opened 2021-03-18
Cannot recreate measurement group with same name


tiagocoutinho:
Hi,
Steps:
1. create a measurement group `mg1` with a channel `c1`: `defmeas mg1 c1`
2. delete measurement group `mg`: `udefmeas mg1`
3. create a measurement group `mg1` with a channel `c2`: `defmeas mg1 c2` (this channel `c2` should belong to a different controller than `c1`)

now we cannot operate the measurement group because the configuration still sees the old c1:
```
Door_bl04_1 [10]: ct
Thu Mar 18 12:07:51 2021

Elements ended acquisition with:
d1d_1:
       State: On (12:07:42.859062)
      Status: c2 is On
              Stopped (12:07:51.041025)
An error occurred while running ct:
PyDs_PythonError: Exception: The synchronization description group has not been initialized

Hint: in Spock execute `www`to get more details

Door_bl04_1 [11]: get_meas_conf all
ActiveMntGrp = mg1
An error occurred while running get_meas_conf:
DB_DeviceNotDefined: device controller/dummyctctrl/dummy2_ctrl not defined in the database !
Hint: in Spock execute `www`to get more details
```

tiagocoutinho:
suggestion for fix:
When removing a measurement group, the pool should also clean the tango properties of that group and also the attribute properties since apparently tango does do it.

reszelaz:
Hi @tiagocoutinho,
Thanks for the report! This is an old issue. It was reported in https://github.com/sardana-org/sardana/issues/145.
At some point, we were trying to fix it with @amilan but at that time we were blocked by a bug in Tango 8. I even have a local branch from that time with the fix and I will try to update it with the develop branch and try it again. I think it should work now, or be very close to the working state because we don't need to care about Tango 8 anymore.
Meanwhile I propose to close this issue as a duplicate.

PR #1607, opened 2021-06-10
Fix: completely clean client side after removing a measurement group

### LINKED ISSUE #145: Changes in MG not see by MacroServer (SF#255)
Hello,
       I am not sure if this bug was already reported, but I have not found any ticket related to it. I have a problem when I delete a MG in the Pool and I create
a new one with the same name. The macroserver see the MG with the elements it has
before I delete it, it does not the see its new elements but the old ones. I have to stop the MacroServer and start it again in order to see the MG (using lsmeas) with the right elements. The problem does not appear if I create a MG with a new name, in this case the MacroServer see that a new MG has been created.
Do you have the same problem?.

                         Regards,
                                     Teresa 

Original Ticket: [sardana/tickets/255](https://sourceforge.net/p/sardana/tickets/255)

reszelaz:
As said on the last follow-up meeting I recovered on old work (still done together with @amilan many years ago) and tried it again. Indeed with Tango 9 it works like a charm and looks like fixes the said bug. Details below:

MeasurementGroup _proxy_ object resides on the client side after removing it on the server. This leads to inconsistency when one recreates the measurement group using the same name e.g. its list of the channels corresponds to the list of the channels of the previously deleted measurement group.

Explicitly remove the Taurus objects (device and its attributes) from the factory so they do not interfere with the future attempts to recreate objects with the same name.

These are the circular references that I tracked and are explained in the inline comment:

measurement_group_circular_refs

Fixes #145, #1528.

### DIFF
--- a/src/sardana/taurus/core/tango/sardana/pool.py
+++ b/src/sardana/taurus/core/tango/sardana/pool.py
@@ -3519,6 +3519,13 @@ class Pool(TangoDevice, MoveableSource):
                                   _pool_data=data)
         return obj
 
+    def _delObject(self, element):
+        elem_type = element.getType()
+        if elem_type in ('ControllerClass', 'ControllerLibrary', 'Instrument'):
+            return
+        obj = element.getObj()
+        obj.factory().removeExistingDevice(obj)
+
     def on_elements_changed(self, evt_src, evt_type, evt_value):
         if evt_type == TaurusEventType.Error:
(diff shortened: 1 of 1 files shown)


13bscsaamjad (approved):

reszelaz:
@13bscsaamjad many thanks for the review. Merging.

PR #1559, opened 2021-04-14
(fix) Avoid already finished macro being aborted at MS exit

reszelaz:
Macro which was already finished is considered as _running
macro_ and at the MacroServer (MS) exit the running macro gets
aborted. This abort is unnecessary and causes the MS to crash
on Windows.

- Introduce the concept of _last macro_.
- When the macro finishes, update the _last macro_ information
and at the same time clear the _running macro_ - this avoids
the macro aborting.
- Maintain the backwards compatibility when reading the macro
data from the door: if it is read during the macro execution
it reports the data of the macro currently being run and as
soon as the macro finished it returns the data of the macro
which just finished execution.

PR #1608, opened 2021-06-10
Hotfix 3.1.1: allow to stop/abort macro executing another hooked macro

### LINKED ISSUE #1603: pre-scan general hooks block ctrl-c
If a general hook is placed in the pre-scan position, then it's not possible to stop for example an ascan with ctrl-c. It seems to go completely ignored.

To reproduce, define anything as a pre-scan hook:
```
Door_B304A [17]: defgh lsenv pre-scan
Defining general hook
lsenv

Door_B304A [18]: lsgh
   Hook place   Hook(s)
 ------------ ---------
     pre-scan     lsenv
```
Then run some scan, an ascan or dscan with dummy motors is enough, and try stoppning it with ctrl-c.
Removing the hook makes it work normally again.
Edit: this seems to affect sardana 3.1.0 only, in 3.0.3 it works.

reszelaz:
#1559 introduced a bug which does not allow to abort macros executing hooked macros. This is because a dedicated `_macro_pointer` is being wrongly invalidated at the end of the `runMacro()` method.

Refactor the concept of _macro_pointer_ - use a property which points to the macro at the `_macro_stack`. This avoids the problem of having to update the `_macro_pointer`.

Deprecate `clearRunningMacro()` in favor of `clearMacroStack()`.

Fixes #1603.

@HEnquist, @13bscsaamjad could you please review and test it? I put a release date for tomorrow. If you merge on another day, before the merge, please update the date of the release in the "What's new?" and in the CHANGELOG.md. Finally, please also continue the [hotfix release process](https://github.com/sardana-org/sardana/wiki/How-to-Hotfix-Sardana) by completing point 6. Thanks!

@teresanunez, also it would be nice if you crosscheck if this is not the same issue you mentioned today on the follow-up. Thanks!

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 3.1.0
+current_version = 3.1.1-alpha
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
 serialize = 
 	{major}.{minor}.{patch}-{release}
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 3.1.1-alpha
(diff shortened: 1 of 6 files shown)


teresanunez:
Hi Zibi, thanks for this. I can not reproduce the issues here, it was indeed related to the GUI.

PR #1610, opened 2021-06-14
Update MeasurementGroup's Elements property when Configuration attr is written

reszelaz:
MeasurementGroup's persistent information about elements is stored in two places:
1. Elements property
2. Configuration memorized attribute

The first one is set only once, at the MeasurementGroup creation time. The subsequent
changes of the configuration does not update the Elements property. Update it on the
successful write of the Configurtion attribute in order to maintain them in sync.


13bscsaamjad (approved):

reszelaz:
I just realized that this PR introduced a small bug. I have fixed it in ef07a263.

PR #1611, opened 2021-06-14
not memorize measurement group attrs like integration time etc.

13bscsaamjad:
Changed some attributes on measurement group to not be memorized.

### DIFF
--- a/src/sardana/tango/pool/MeasurementGroup.py
+++ b/src/sardana/tango/pool/MeasurementGroup.py
@@ -331,10 +331,10 @@ class MeasurementGroupClass(PoolGroupDeviceClass):
     #    Attribute definitions
     attr_list = {
         'IntegrationTime': [[DevDouble, SCALAR, READ_WRITE],
-                            {'Memorized': "true",
+                            {'Memorized': "false",
                              'Display level': DispLevel.OPERATOR}],
         'MonitorCount': [[DevLong, SCALAR, READ_WRITE],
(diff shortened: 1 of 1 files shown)


reszelaz (approved):
When working on #1588 we realized that restoring the IntegrationTime's memorized value when some of the measurement group's controllers is _offline_ is problematic.

After quick reasoning, we think that IntegrationTime, MonitorCount, NbStarts, and SynchDescription should actually not be memorized. This is somehow _in conflict_ with the original design of Device Pool still available for reference in this [old document](https://github.com/sardana-org/sardana/blob/develop/doc/source/devel/api/tango_device_pool.rst):

> Integration_time, Integration_count, timer and monitor are memorized attributes. This means that the configuration values of these attributes are stored in the database. The next time the Pool starts the values are restored. This is done in order to reduce Pool configuration at startup to the minimum.

However, we believe this is outdated e.g. Timer and Monitor attributes were already removed and form part of the Configuration attribute which is memorized. We also believe that IntegrationTime and these other attributes are just related to a given acquisition and should not be stored persistently (we even doubt about the Moveable attribute).

Before merging it, it would be nice to hear other's opinions. Many thanks!

reszelaz:
@sardana-org/integrators any comment on this one?

teresanunez:
For me it is ok not having them memorized. I think it is not needed.

reszelaz:
Thanks @teresanunez for your quick answer. I'm merging.

PR #1612, opened 2021-06-14
Added state print to list macros (lsm, lsmeas etc.)

13bscsaamjad:
- Added state print to list macros (lsm, lsmeas etc.)

### DIFF
--- a/src/sardana/macroserver/macros/lists.py
+++ b/src/sardana/macroserver/macros/lists.py
@@ -110,9 +110,9 @@ class _lsobj(_ls):
 
     subtype = Macro.All
 
-    cols = 'Name', 'Type', 'Controller', 'Axis'  # , 'State'
-    width = -1,     -1,           -1,     -1  # ,      -1
-    align = Right,  Right,        Right,  Right  # ,   Right
+    cols = 'Name', 'State', 'Type', 'Controller', 'Axis'
+    width = -1, -1, -1, -1, -1
+    align = Right, Right, Right, Right, Right
 
     show_overwritten_msg = False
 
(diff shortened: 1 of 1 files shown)


reszelaz (approved):
@13bscsaamjad, thanks for this improvement!

I tried it and works fine.
I also realized that it was already foreseen to add the state column to the list macros (it was commented out).
Here is a sample of how it works:

```
Door/zreszela/1 [10]: lsmeas
  Active        Name   State   Timer Experim. channels       
 -------- ---------- ------- ------- ------------------------
    *             kk      ON    ct01 ct01, ct02, ct03, ct04  
            mntgrp01      ON    ct01 ct01, ct02, ct03, ct04  

Door/zreszela/1 [11]: lsm
       Name   State          Type   Controller   Axis
 ---------- ------- ------------- ------------ ------
      gap01      ON   PseudoMotor   slitctrl01      1
     mot-05      ON         Motor    motctrl01      5
      mot01      ON         Motor    motctrl01      1
      mot02      ON         Motor    motctrl01      2
      mot03      ON         Motor    motctrl01      3
      mot04      ON         Motor    motctrl01      4
   offset01      ON   PseudoMotor   slitctrl01      2
```

I also like the idea of having this column right after the name column, but if there are other opinions, it would be nice to hear them before merging. Many thanks!

reszelaz:
Maybe before eventual merging, a good idea would be to try it in a real system, with a big number of elements. I remember that, in the past, after complaints about the slowness of the `wa` macro we optimized and use Tango asynchronous readout of the attributes. Could someone try it first?

teresanunez:
Hi, I find the idea nice. About speed ... it will depend on the motors, there are some slow one where it will take
time, even probably cases where a motor is not responding ... but I do not think that this will be causing problems
in a macro that is only for listing.

PR #1614, opened 2021-06-16
fix: hang of IPython when macro input gives timeout

### LINKED ISSUE #1613: Interactive question in macro freezes spock when timed out
The macro defined below ask for input from user. If user answers before time out then it works just fine. But if you allow the question to time out then spock gets freezed when prompt is returned.

The problem can be reproduced with python3, sardana 3.0.4 in debian9.

```python
#!/usr/bin/env python


from sardana.macroserver.macro import Macro


class question(Macro):

    interactive = True

        answer = ''
        while not answer.lower() in ('y', 'n'):
            answer = self.input('Answer (y/n)', timeout=10, default_value='y')
        self.output('User answered: %s' % answer)

```

reszelaz:
When macro input gives timeout then the Spock (IPython) gets hung completely.
It seems to be a bug in IPython - see https://github.com/ipython/ipython/issues/12078.
Use [`prompt_toolkit.prompt()`](https://python-prompt-toolkit.readthedocs.io/en/master/pages/reference.html#prompt_toolkit.shortcuts.prompt) (a dependency of IPython) instead of `input()` and raise an exception in the thread waiting for input when the timeout was reached. This apparently works while `input()` does not react on exceptions sent in this way.

Also change the message in Spock when the timeout is reached so it appears this way:

```
Door/zreszela/1 [3]: question
Answer (y/n) [y] Input timeout reached!
User answered: y
```

instead of:

```
Door/zreszela/1 [4]: question
Answer (y/n) [y] SpockInputHandler input timeout
User answered: y
```
Fixes #1613.

@jairomoldes, could you please test it? Many thanks!

### DIFF
--- a/src/sardana/spock/inputhandler.py
+++ b/src/sardana/spock/inputhandler.py
@@ -29,6 +29,9 @@ __all__ = ['SpockInputHandler']
 
 __docformat__ = 'restructuredtext'
 
+import threading
+
+from sardana.util.thread import raise_in_thread
 from sardana.taurus.core.tango.sardana.macroserver import BaseInputHandler
 
 from sardana.spock import genutils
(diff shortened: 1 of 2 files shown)


reszelaz:
@jairomoldes, in ALBA's internal ticket, confirms that this PR fixes the issue. @sardana-org/integrators if there are no comments, I will proceed to merge. Thanks to all!

reszelaz:
Thanks @jairomoldes for the tests. Since there are no comments, I proceed to merge.

PR #1615, opened 2021-06-17
Udef checks dependencies

### LINKED ISSUE #1586: udef checks dependencies
Hi,

we just somehow crashed our Sardana configuration and it was somehow due to the deletion of a motor that was required for several pseudo motors to work and was included in `motor groups`

For the later issue I already opened #1059 

For the first problem I suppose to add a dependency check to the `udefelem` and `udefctrl` macros in order to check if the according element has any dependencies within the connected Pools.

Do you think this is possible?

Best

Daniel

catunlock:
This pull request resolves #1586. Now a check for dependent elements is performed when an element is going to be undefined. 

So far we tested it for Pseudocounters, Motorgroups, Pseudomotors and, Measurement Groups.

The remaining tests could be testing for Pseudomotors that have as dependent elements another Pseudomotors.

----

IMPORTANT: This PR introduces usage of `gc.collect()` in Sardana. cycle-reference may exist between an element and a traceback stored in `SardanaValue` or `SardanaAttribute` as a consequence of getting `sys.exc_info()` and storing a reference to it. Since _dependent elements_ are determined by introspecting the listeners we need to attempt garbage collection of these cycled-referenced objects.

### DIFF
--- a/src/sardana/pool/pool.py
+++ b/src/sardana/pool/pool.py
@@ -557,10 +557,15 @@ class Pool(PoolContainer, PoolObject, SardanaElementManager, SardanaIDManager):
             except:
                 raise Exception("There is no element with name '%s'" % name)
 
-        elem_type = elem.get_type()
+        # TODO: most probably due to some cycle-reference the psuedo counter
+        #  objects are not being deleted when undefining them, as a workaround
+        #  try to delete them with gc.collect()
         dependent_elements = elem.get_dependent_elements()
-        if len(dependent_elements) > 0:
-            gc.collect()
+        for dependent_element in dependent_elements:
+            if dependent_element.get_type() == ElementType.PseudoCounter:
+                gc.collect()
+                break
+
         dependent_elements = elem.get_dependent_elements()
         if len(dependent_elements) > 0:
             names = [elem.name for elem in dependent_elements]
@@ -571,7 +576,8 @@ class Pool(PoolContainer, PoolObject, SardanaElementManager, SardanaIDManager):
                 "DeleteElement(<motor_group_name>) command on the Pool e.g. "
                 "Pool_demo1_1.DeleteElement('_mg_ms_20671_1') in Spock."
                 .format(name, ", ".join(names)))
-            
+        
+        elem_type = elem.get_type()    
         if elem_type == ElementType.Controller:
             if len(elem.get_elements()) > 0:
                 raise Exception("Cannot delete controller with elements. "
--- a/src/sardana/pool/pool.py
+++ b/src/sardana/pool/pool.py
@@ -32,6 +32,7 @@ __all__ = ["Pool"]
 
 __docformat__ = 'restructuredtext'
 
+import gc
 import os.path
 import logging.handlers
 
@@ -557,7 +558,9 @@ class Pool(PoolContainer, PoolObject, SardanaElementManager, SardanaIDManager):
                 raise Exception("There is no element with name '%s'" % name)
 
         elem_type = elem.get_type()
-
+        dependent_elements = elem.get_dependent_elements()
+        if len(dependent_elements) > 0:
+            gc.collect()
         dependent_elements = elem.get_dependent_elements()
         if len(dependent_elements) > 0:
             raise Exception(
--- a/src/sardana/pool/pool.py
+++ b/src/sardana/pool/pool.py
@@ -562,9 +562,10 @@ class Pool(PoolContainer, PoolObject, SardanaElementManager, SardanaIDManager):
         if len(dependent_elements) > 0:
             raise Exception(
                 "The element {} can't be deleted because {} depend on it."
-                "\n\nIf the name of the dependent element starts with" 
-                "'_mg_ms_*' it means that are motor groups, use any tango " 
-                "client (e.g. Jive) to delete them and restart the server."
+                "\n\nIf the name of the dependent element starts with " 
+                "'_mg_ms_*' it means that are motor groups, execute "
+                "DeleteElement(<motor_group_name>) command on the Pool e.g. "
+                "Pool_demo1_1.DeleteElement('_mg_ms_20671_1') in Spock."
                 .format(name, ", ".join(dependent_elements)))
             
         if elem_type == ElementType.Controller:
--- a/src/sardana/pool/pool.py
+++ b/src/sardana/pool/pool.py
@@ -558,11 +558,13 @@ class Pool(PoolContainer, PoolObject, SardanaElementManager, SardanaIDManager):
                 raise Exception("There is no element with name '%s'" % name)
 
         elem_type = elem.get_type()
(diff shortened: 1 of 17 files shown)


reszelaz:
@dschick I think this is ready to be tested. 
All the implementation looks ok, but the 97a96ee. But I think it is better to go for it with `gc.collect()` in this special case of pseudo counters - we already spent a lot of time trying to break the cycle reference and there are a lot of other issues waiting in the backlog. I hope it won't cause any problem.
Also, we have found that even if we delete Tango device it somehow resides in the device server precess. This explains 
2b9359d. @catunlock will try to reproduce it with a _dummy_ device server.
Thanks a lot @catunlock and @dschick ! I think this enhancement will add a great value to Sardana.

yimli:
I have tested this PR for #1658 . It works as expected and prevents the user from removing a measurement group dependent channel with udefelem. Thanks for the implementation!

reszelaz:
Many thanks @yimli for your test!
@dschick do you also want to perform some tests?

dschick:
Its fine for me. Many thanks

reszelaz ***@***.***> schrieb am Mi., 14. Juli 2021, 11:29:

teresanunez (approved):

src/sardana/pool/pool.py
```diff
@@ -557,6 +558,19 @@ def delete_element(self, name):
                 raise Exception("There is no element with name '%s'" % name)
 
         elem_type = elem.get_type()
+        dependent_elements = elem.get_dependent_elements()
+        if len(dependent_elements) > 0:
+            gc.collect()
```
13bscsaamjad, line 563: perhaps a comment regarding why this is required here in the PR description?

13bscsaamjad (approved):

13bscsaamjad:
I have tested Pseudomotors that have as dependent elements another Pseudomotors and it works!

reszelaz:
I have performed some more work on this and introduced the following changes.

### change return type of get_dependent_elements()
I change the return value of `get_dependent_elements()` to return dependent elements objects instead of names in order to extend use cases of this method e.g. allow to check the type of the returned dependent elements - see 146edfa

### Fix case with meas group with undefined disabled channel
Disabled channels in the measurement group are considered as not members of the group. They neither appear in "element list" attribute nor the measurement group listens to their events. Hence it is possible to undefine an experimental channel that is disabled. This may lead to a situation when we try to enable again an undefined channel. Consider this case by not accepting a configuration that contains an undefined element. Also, allow to acquire with a measurement group that contains an undefined and disabled channel - see 6ac7aa6

**Example:**

```
Door/zreszela/1 [11]: get_meas_conf
ActiveMntGrp = mntgrp01
   Channel   Enabled   PlotType   PlotAxes
 --------- --------- ---------- ----------
      ct01      True         No        n/a
      ct02     False         No        n/a

Door/zreszela/1 [12]: udefelem ct02

Door/zreszela/1 [13]: get_meas_conf
ActiveMntGrp = mntgrp01
   Channel   Enabled   PlotType   PlotAxes
 --------- --------- ---------- ----------
      ct01      True         No        n/a
      ct02     False         No        n/a

Door/zreszela/1 [14]: ct
Wed Jul 21 11:32:31 2021

  ct01  =         1.0

Door/zreszela/1 [15]: set_meas_conf enabled True ct02
ActiveMntGrp = mntgrp01
An error occurred while running set_meas_conf:
PyDs_PythonError: ValueError: ct02 is not defined

Door/zreszela/1 [16]: ct
Wed Jul 21 11:32:55 2021

  ct01  =         1.0

Door/zreszela/1 [17]: defelem ct02 ctctrl01 2
Created ct02

Door/zreszela/1 [18]: set_meas_conf enabled True ct02
ActiveMntGrp = mntgrp01

Door/zreszela/1 [19]: ct
Wed Jul 21 11:33:09 2021

  ct01  =         1.0
  ct02  =         2.0
```
### Fix `gc.collect()` of dependent elements

cycle-reference may exist between an element and a `traceback` stored in `SardanaValue` or `SardanaAttribute` as a consequence of getting `sys.exc_info()`. Keeping a list of dependent elements in the scope of `Pool.delete_element()` makes it impossible to `gc.collect()` to collect the cycled objects. Use `has_dependent_elements()` in order to determine if there are dependent elements and then call the `gc.collect()`. Also, do not call it conditionally only for the pseudo counters but for any element type - the cycle may be created between any element and a traceback - see 563e299

**IMPORTANT: this last point should trigger a discussion if we should keep in Sardana references to the `sys.exc_info()` (traceback) objects.**

I added a comment about the `gc.collect()` usage to the PR description @13bscsaamjad. Now it is well understood why we need to call it.

Thanks to all for your work and patience in this PR: @catunlock, @13bscsaamjad, @yimli and @dschick. I proceed with the merge!

yimli:
Many thanks! @reszelaz I just have a follow-up question about undefining disabled channel. It is reasonable that we should be able to undefine a disabled experimental channel through 'udefelem', but that still cause the expconf to report channel missing error if we open expconf after undefining a disabled channel. Maybe even if the channel is disabled, it is better to first manually remove it from expconf? I am not sure about this.

reszelaz:
Thanks for the tests with the expconf @yimli. I confirm what you observe. We get the following exception following the example from https://github.com/sardana-org/sardana/pull/1615#issuecomment-885015887

```
Door/zreszela/1 [6]: expconf
MainThread     WARNING  2021-07-23 13:38:05,385 TaurusRootLogger: Problem with taurus.qt.editor (hint: is spyder >=3 installed?)

Door/zreszela/1 [7]: MainThread     INFO     2021-07-23 13:38:05,856 TaurusRootLogger: Using PyQt5 (v5.12.3 with Qt 5.12.9 and Python 3.9.1)
MainThread     INFO     2021-07-23 13:38:06,028 TaurusRootLogger: Plugin "taurus_pyqtgraph" lazy-loaded as "taurus.qt.qtgui.tpg"
MainThread     WARNING  2021-07-23 13:38:06,040 TaurusRootLogger: <frozen importlib._bootstrap>:228: DeprecationWarning: taurus.core.util.argparse is deprecated since 4.5.4. Use argparse or (better) click instead

MainThread     WARNING  2021-07-23 13:38:06,147 TaurusRootLogger: epics scheme not available: ModuleNotFoundError("No module named 'epics'")
MainThread     WARNING  2021-07-23 13:38:06,167 TaurusRootLogger: Problem with taurus.qt.editor (hint: is spyder >=3 installed?)
  File "/homelocal/zreszela/workspace/sardana/src/sardana/taurus/qt/qtgui/extra_sardana/measurementgroup.py", line 474, in flags
    ch_info = self.getAvailableChannels()[ch_name]
KeyError: 'tango://pt222.cells.es:10000/expchan/ctctrl01/2'

```

I will check if there is an easy way to prevent this exception in the expconf.

Issue #1117, opened 2019-05-27
Basic tutorial for debugging MacroServer with rconsole


daneos:
Is there any tutorial on using rconsole for debugging?
In the documentation there is only a simple connection instruction and not much more.

reszelaz:
With this [example](https://sardana-controls.org/users/configuration/macroserver.html) you get the the [MacroServer](https://sardana-controls.org/devel/api/sardana/macroserver/macroserver.html#sardana.macroserver.macroserver.MacroServer) object and then the idea is to explore it according to its API.
Do you need something specific? What type of problem are you debugging?

daneos:
I have a problem with some macro sequences hanging the MacroServer. I'd like to see if there are some stuck operations, not finished threads or something like this.
Since I'm not getting any useful information in logs, I thought that probably exploring MacroServer internals will help to debug this problem.

reszelaz:
You could also attach to a running process with gdb and check the backtrace of all threads. Maybe this will give you more details.

reszelaz:
Could it be that you also hit #1080?

reszelaz:
I think there is no `rfoo` module for python3. I'll propose a PR eliminating rfoo usage from Sardana code and docs.

PR #1622, opened 2021-06-29
fix: remove rfoo (rconsole) usage in sardana and in docs

reszelaz:
rfoo is not available for Python 3. Remove its usage from sardana.

Relates to #1117


reszelaz:
Thanks to the investigation of @cpascual we know that the `rfoo` was dropped from Debian starting with the Bullseye release which does not support Python 3. IMO there is no need to bother @sardana-org/integrators with the review. I will auto-merge.

PR #1624, opened 2021-06-30
fix: provide back compat for external ctrls meas conf

reszelaz:
#867 accidentally introduced backward-incompatible changes
which break the measurement groups which external controllers e.g. Tango
have synchronizer/timer/monitor configured. Relax validation and print
a deprecation warning instead of raising an exception.

This was discovered at ALBA by @jairomoldes and @tiagocoutinho. @jordiandreu could you test if it avoids the problem?
I would expect that with this change, if you read the configuration back, the synchronizer/timer/monitor parameters should not be there anymore, could you check this as well, please?

### DIFF
--- a/src/sardana/pool/poolmeasurementgroup.py
+++ b/src/sardana/pool/poolmeasurementgroup.py
@@ -719,15 +719,13 @@ class MeasurementConfiguration(object):
             # The external controllers should not have synchronizer
 
             if external:
-                if 'synchronizer' in ctrl_data:
-                    raise ValueError('External controller does not allow '
-                                     'to have synchronizer')
-                if 'monitor' in ctrl_data:
-                    raise ValueError('External controller does not allow '
-                                     'to have monitor')
-                if 'timer' in ctrl_data:
-                    raise ValueError('External controller does not allow '
-                                     'to have timer')
+                for parameter in ['synchronizer', 'timer', 'monitor']:
+                    if parameter in ctrl_data:
+                        msg = (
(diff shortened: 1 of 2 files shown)


src/sardana/pool/poolmeasurementgroup.py
```diff
@@ -719,15 +719,13 @@ def set_configuration_from_user(self, cfg):
...
-                if 'timer' in ctrl_data:
-                    raise ValueError('External controller does not allow '
-                                     'to have timer')
+                for parameter in ['synchronizer', 'timer', 'monitor']:
+                    if parameter in ctrl_data:
+                        msg = (
+                            '{} is deprecated for external controllers '
+                            'e.g. Tango, since 3.0.3. Re-apply configuration '
+                            'in order to upgrade.').format(parameter)
+                    self._parent.warning(msg)
```
tiagocoutinho, line 728: I think line 728 should have one more level of indentation

reszelaz: Thanks for pointing this out. I already fixed it.

jordiandreu:
I have applied this patch and it worked fine. Many thanks to both of you @reszelaz and @tiagocoutinho !

reszelaz:
I have make it a little bit more _sophisticated_. The backwards compatibility works only for the moment of restoring the memorized value. This was necessary because there were unit tests expecting the exception.

Thanks for the tests @jordiandreu! @sardana-org/integrators, this is a trivial change, I will auto-merge if there are no comments. Thanks!

PR #1627, opened 2021-07-01
Allow to run Sardana scripts as Python modules

catunlock:
This addition allows for the user to run the Sardana scripts as
modules, using `python -m` i.e: `python -m sardana.tango.macroserver`.

This allows running debuggers passing the module name.
 (Especially helpful for VS Code).


reszelaz:
Many thanks @catunlock. Works as expected. For PyCharm it is also very usefull!

PR #1629, opened 2021-07-06
Improve devcontainer

catunlock:
The improvements of this PR are:

- Adds the Git Graph and Git Lens extensions to the VS Code. This allows a better workflow using Git, and moving between branches.
- Adds autopep8 and flake8 linter and autoformat to the environment, these packages are used by VS Code to format and lint the code.
- Adds ssh package allowing to push to repositories using ssh+git protocol.


reszelaz (approved):
Thanks for this improvement @catunlock. I have personally not tested them, but I think this should basically work.

PR #1630, opened 2021-07-07
feat: add history log message for motor attributes

### LINKED ISSUE #1508: Log changes of position related attributes
In older Tango versions we were relying on the history of the attribute properties used by memorized attributes, like offset, sign and step_per_unit to keep track of the changes. This however was romoved after this [poll](https://lists.tango-controls.org/wws/arc/info/2015-07/msg00008.html). At ALBA, we plan to use the HDB++ (extended with archiving on change events) for this purposes, but this won't come shortly.

Meanwhile,  with @jordiandreu we were thinking that simply logging the changes in the usual server log files would be of great help in tracking the changes. Hence we propose to add logs like:

- sign set to -1
- offset set to -2.34
- step_per_unit set to 567

For the moment I would start with the debug log level.

With this, at least, we will gain access to the history of changes limited by the numer and size of log files which are configurable with the [`sardanacustomsettings`](https://sardana-controls.org/users/configuration/sardanacustomsettings.html#sardana-custom-settings) module. Also, the ones using elastic, or similar tools, could easily gain unlimited history.

Note that on the server startup, when restoring the memorized values, this may repeat the last setting in the log but it won't be harmfull and after implementing https://github.com/sardana-org/sardana/issues/1458 this repeated log could be removed.

Thanks in advace for your feedback!

catunlock:
Since Tango version 8 the memorized attributes lost the history of
changes and is not possible to easily track changes.

Add log messages to offset, sign and step_per_unit
attributes to facilitate tracking over changes.

fixes #1508

### DIFF
--- a/src/sardana/pool/poolmotor.py
+++ b/src/sardana/pool/poolmotor.py
@@ -403,6 +403,7 @@ class PoolMotor(PoolElement):
 
     def set_offset(self, offset, propagate=1):
         self._offset.set_value(offset, propagate=propagate)
+        self.debug("Set attribute offset = {}".format(offset))
 
     offset = property(get_offset, set_offset, doc="motor offset")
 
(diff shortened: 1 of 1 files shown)


reszelaz (approved):
Nice demo today on the SBPS! Many thanks @catunlock!

PR #1631, opened 2021-07-07
put in limit checks

### LINKED ISSUE #267: Validate limits in set_lim & set_lm (SF#422)
This is a request from Daniel Fulla (internal Alba's ticket: #42912):

Recently I did a mistake. I set the limits to a motor in the wrong order resulting that the maximum was the minimum and vice versa. The system accepted the values and afterwards I could not act the motor neither get the position. This mistake could happen again to me or other people. The error message that you get is not clear and it makes the troubleshooting difficult. It would be an improvement if the system would not accept a maximum lower than the minimum for example. (set_lim & set_lm)

Original Ticket: [sardana/tickets/422](https://sourceforge.net/p/sardana/tickets/422)

HassaanTS:
limit checking for making sure that limits don't accept maximum lower than the minimum and adding try except blocks 

fixes #267

### DIFF
--- a/src/sardana/macroserver/macros/standard.py
+++ b/src/sardana/macroserver/macros/standard.py
@@ -263,9 +263,17 @@ class set_lim(Macro):
     def run(self, motor, low, high):
         name = motor.getName()
         self.debug("Setting user limits for %s" % name)
-        motor.getPositionObj().setLimits(low, high)
-        self.output("%s limits set to %.4f %.4f (user units)" %
-                    (name, low, high))
+        if low > high:
+            raise Exception('set_lim exception: lower limit can not be a higher value than the high limit')
+        else:
+            try:
+                motor.getPositionObj().setLimits(low, high)
+                self.output("%s limits set to %.4f %.4f (user units)" %
+                (name, low, high))
+            except Exception as e:
+                self.debug("exception occured during setting user limits %s" % e)
+
+            
 
 
 class set_lm(Macro):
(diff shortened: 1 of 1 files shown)


src/sardana/macroserver/macros/standard.py
```diff
@@ -263,9 +263,15 @@ class set_lim(Macro):
...
-                    (name, low, high))
+        if low > high:
+            raise Exception('set_lim exception: lower limit can not be a higher value than the high limit')
+        else:
+            try:
+                motor.getPositionObj().setLimits(low, high)
+                self.output("%s limits set to %.4f %.4f (user units)" %
+                (name, low, high))
+            except Exception as e:
+                self.debug("exception occured during setting user limits %s" % e)
```
reszelaz, line 274:
```suggestion
        if low > high:
            raise ValueError('lower limit must be lower than the higher limit')
        motor.getPositionObj().setLimits(low, high)
        self.output(
            "%s limits set to %.4f %.4f (user units)" % (name, low, high)
        )
```
- `else` is not necessary, an exception already controls the flow and one indentation level less is easier to read:)
- macros which due to some exceptional condition can not do what they are supposed to do should not catch the exception and return successfully. Let's imagine that we execute the macro in a sequence and the following macros rely on the previous macro result.
If you agree, could you please apply similar changes to the `set_lm` macro. Many thanks!

jangarrevoet:
Would it make sense to also include a check that the current position is actually within the limit range?

HassaanTS:
> Would it make sense to also include a check that the current position is actually within the limit range?

what if there's a case where the user decides to change the limits for a newer scan while being outside of those limits because the previous scan required them? wouldn't this always fail in that case?

aureocarneiro (approved):

aureocarneiro:
If the actual position is higher, the device will go into ALARM state. If the user called this macro he knows that this is the limit he wants. I think this PR is fine.

reszelaz (approved):

reszelaz:
Merging! Thanks to all for your contribution to this enhancement.

Regarding https://github.com/sardana-org/sardana/pull/1631#issuecomment-876122898, maybe a good compromise, in this case, would be to add a warning message and not raise an exception? But this can be treated as a separate PR. I will merge it as it is. Thanks again!

PR #1632, opened 2021-07-07
Take scan starttime before snapshots

### LINKED ISSUE #1321: Scan start time taken too late
When a scan starts the following actions are done in sequence:

- construct GScan
  - _setupEnvironment()
    - prepare header (serialno, user, title, etc)
    - prepare columns
    - **takeSnapshot**
    - prepare directories (ScanDir & ScanFile)
- Gscan.scan()
  - start()
    - set header *start time*
    - scan loop
    - end

Take note that some actions, like takeSnapshot, may take some time. This not only because it is done sequentially but also because there might be some timeouts.

I would expect that it reflects the time when the scan macro was started. This is the first thing that appears in a scan header in a spec file for example so I would say it what the users would expect also.

But since the start time is taken only when the scan loop starts, it might be several seconds off the actual scan operation started.
(the method name `start()`  misleads us as it is intended to state the start of the scan loop and not the start of the scan)

Therefore, I propose that *start_time* be moved to the initial region when the serialno, user and title are set.

HEnquist:
This pull request fixes #1321 
The fix is to take the start time already when creating the scan object, and not when starting the scan loop (after potentially slow things like taking snapshots).

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -677,6 +677,9 @@ class GScan(Logger):
              'user': user,
              'title': self.macro.getCommand()})
 
+        env['startts'] = ts = time.time()
+        env['starttime'] = datetime.datetime.fromtimestamp(ts)
+
         # Initialize the data_desc list (and add the point number column)
         data_desc = [
             ColumnDesc(name='point_nb', label='#Pt No', dtype='int64')
@@ -993,8 +996,6 @@ class GScan(Logger):
     def start(self):
         self.do_backup()
         env = self._env
-        env['startts'] = ts = time.time()
-        env['starttime'] = datetime.datetime.fromtimestamp(ts)
         env['acqtime'] = 0
         env['motiontime'] = 0
         env['deadtime'] = 0


HEnquist:
Alternatively, one can also keep track of both macro start time and scan start time. This makes it possible to report the deadtime from both motion and scan setup:
```
Door_dummy_1 [5]: ascan mot01 0 1 4 0.1
Operation will be saved in /tmp/dummmy2.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #24 started at Thu Jul  8 08:38:42 2021. It will take at least 0:00:01.348528
 #Pt No    mot01      ct01      ct02       dt   
   0         0        0.1       0.2     0.34582 
   1        0.25      0.1       0.2     0.660809
   2        0.5       0.1       0.2     0.969109
   3        0.75      0.1       0.2     1.28313 
   4         1        0.1       0.2      1.5818 
Operation saved in /tmp/dummmy2.h5 (HDF5::NXscan)
Scan #24 ended at Thu Jul  8 08:38:43 2021, taking 0:00:01.726501. Dead time 69.8% (setup time 1.2%, motion dead time 56.1%)
```

PR #1633, opened 2021-07-07
raise exception in relmaclib and relmac methods if any is detected

aureocarneiro:
This suggestion comes from Manu Taurel from ESRF.

Tango commands: ReloadMacro and ReloadMacroLib does not report problems with reloading macros or macro libraries. If the module being reloaded contains bugs it gets silently "reloaded". Maybe raising an exception could a good way of communicating it?

Original Ticket: sardana/tickets/118

Fix: #75 

-----------------------------

Description of the argout says: "[OK] if successfull or a traceback if there was a error (one string with complete traceback of each error". However even if the error occurs the [OK] string is returned.

Original Ticket: sardana/tickets/119

Fix: #76

### DIFF
--- a/src/sardana/macroserver/macroserver.py
+++ b/src/sardana/macroserver/macroserver.py
@@ -447,8 +447,10 @@ class MacroServer(MSContainer, MSObject, SardanaElementManager, SardanaIDManager
         new_elements, changed_elements, deleted_elements = [], [], []
 
         new_lib = manager.reloadMacroLib(lib_name)
+        
         if new_lib.has_errors():
-            return new_lib
+            exc_type, exc_value, exc_traceback = new_lib.get_error()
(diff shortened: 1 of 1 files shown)


aureocarneiro:
ping @CarlaCTaka and @vanesil

aureocarneiro:
Tested in the startup, using `ReloadMacro` and `ReloadMacroLib` tango commands and from spock using `relmac`, `relmaclib` and `addmaclib`

teresanunez:
Hi, I have checked this, but I do see an error without the change here and the message is clearer than after the change.
I am now doubting if I am not checking the right thing ...

aureocarneiro:
probably you are not looking at the tango commands `ReloadMacro` and `ReloadMacroLib`.

teresanunez:
I saw now the behaviour and it is nicely implemented. Thanks a lot. I merge it.

PR #1634, opened 2021-07-07
Issue 1303: change the timescan first parameter to nb_points

### LINKED ISSUE #1303: one-point scan
this is a rather stupid question, but I was wondering, if there is a simple way to do a step-scan with only one point?
I was not able to realize that with a simpe `ascan`, I always got two point.
I guess, I could use the `ascan `macro as a template and change it accordingly, but maybe you have a better advice?

The reason for my request is, that especially when worling with 2D detectors, I would like to write an alternative to a simple `ct` in order save also all meta data for an acquired image.
So instead of doing a `ct` and writing down the filename and all motor positons for that image in my labbook, I would like to do a `onepointscan`, or name it `capture`, which creates a full scan entry in the data file, including the snapshot, but with only one scan point, ideally with a dummy or the time as the movable.

jkotan:
It resolves #1303 by 
- changing the timescan first parameter to nb_points
-  allowing perform ascan with nr_intervals = 0

### DIFF
--- a/src/sardana/macroserver/macros/scan.py
+++ b/src/sardana/macroserver/macros/scan.py
@@ -127,7 +127,10 @@ class aNscan(Hookable):
 
         self.motors = motorlist
         self.starts = numpy.array(startlist, dtype='d')
-        self.finals = numpy.array(endlist, dtype='d')
+        if scan_length:
+            self.finals = numpy.array(endlist, dtype='d')
+        else:
+            self.finals = numpy.array(startlist, dtype='d')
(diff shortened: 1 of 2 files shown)


jkotan:
Hi Zibi, I've just added implement of the not `do_last_point` flag with negative interval values so know you can run
```
ascanct 0 3 -3 0.1 0.1
``` 
to get points from 0 to 2
and 
```
ascanct 0 3 3 0.1 0.1
``` 
to get points from 0 to 3.

PR #1635, opened 2021-07-07
Avoid errors in edctrlcls and edctrllib macros

teresanunez:
This PR avoid errors coming up when running edctrlcls and edctrllib macros. The macros are still not
functional, since they have to be moved to magic commands (like edmac and edmaclib is done), but
this PR solves bugs in the code of the macros and of the sardana core. Now the code is ready for
further implementation (not done now since it has very low priority and it has to be discussed if it
is worth to implement it). Issue #317 is closed with this PR.


johanfforsberg:
This looks good to me, I managed to reproduce the issue in develop, and check that it no longer happens with this branch.

reszelaz:
Thanks @teresanunez and @johanfforsberg for your work. Merging.

PR #1637, opened 2021-07-07
Fix #87  (error with duplicated pre-scan snapshots)

### LINKED ISSUE #87: nexus recorder fails when an element is present twice in the snapshot group (SF#138)
If you add an element twice in the snapshot group (expconf allows you to do that), the scan execution fails due to problem with nexus recorder:

  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/msmacromanager.py", line 1132, in runMacro
    for step in macro_obj.exec_():
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/macro.py", line 2150, in exec_
    for i in it:
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/macros/scan.py", line 250, in run
    for step in self._gScan.step_scan():
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/gscan.py", line 789, in step_scan
    self.start()
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/gscan.py", line 745, in start
    self.data.start()
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/scandata.py", line 259, in start
    self.datahandler.startRecordList(self)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/recorder/datarecorder.py", line 58, in startRecordList
    recorder.startRecordList(recordlist)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/recorder/datarecorder.py", line 116, in startRecordList
    self._startRecordList(recordlist)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/recorder/storage.py", line 794, in _startRecordList
    self._createPreScanSnapshot(env)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/recorder/storage.py", line 812, in _createPreScanSnapshot
    nid = self._writeData(label, pre_scan_value, dtype, shape=dd.shape or (1,)) #@todo: fallback shape is hardcoded!
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/recorder/storage.py", line 603, in _writeData
    self._makedata(name, dtype=dtype, shape=shape, chunks=chunks)
  File "/home/zreszela/workspace/sardana/src/sardana/macroserver/scan/recorder/storage.py", line 588, in _makedata
    self.fd.compmakedata(name, dtype=dtype, shape=shape, mode=mode, chunks=chunks)
  File "/homelocal/sicilia/lib/python/site-packages/nxs/napi.py", line 909, in compmakedata
    "Could not create compressed data %s: %s"%(name,self._loc())
ValueError: Could not create compressed data mot36: /home/zreszela/tmp/test.h5(/entry144/measurement/pre_scan_snapshot)

Original Ticket: [sardana/tickets/138](https://sourceforge.net/p/sardana/tickets/138)

cpascual:
- Nexus recorder: handle duplicated snapshot labels 
- Do not allow setting duplicated snapshots in Exponf 

Fixes #87 

(worked with @emiliojmorales )

### DIFF
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -301,6 +301,13 @@ class NXscanH5_FileRecorder(BaseFileRecorder):
 
         for dd in self.preScanSnapShot:
             label = self.sanitizeName(dd.label)
+            if label in _snap:
+                self.warning(
+                    "PreScanSnapShot: skipping duplicated label'{}'".format(
+                        label
+                    )
+                )
+                continue
             dtype = dd.dtype
             pre_scan_value = dd.pre_scan_value
             if dd.dtype == 'bool':
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
@@ -427,7 +427,8 @@ class ExpDescriptionEditor(Qt.QWidget, TaurusBaseWidget):
         conf = door.getExperimentConfiguration()
         self._originalConfiguration = copy.deepcopy(conf)
         self.setLocalConfig(conf)
-        self._setDirty(False)
+        # Flag as "dirty" if some config was changed during the set-up
+        self._setDirty(self._localConfig != self._originalConfiguration)
         self._dirtyMntGrps = set()
         # set a list of available channels
         avail_channels = {}
--- a/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
+++ b/src/sardana/taurus/qt/qtgui/extra_sardana/expdescription.py
@@ -671,7 +671,18 @@ class ExpDescriptionEditor(Qt.QWidget, TaurusBaseWidget):
             else:
                 full_name = nfo.full_name
                 display = nfo.name
+            if full_name in [fn for fn, _ in preScanList]:
+                msg = ("'{}' defined more than once in snapshot.\n"
+                       + "Only one entry will be kept").format(display)
+                Qt.QMessageBox.warning(
+                    self, "Duplicated entry", msg, Qt.QMessageBox.Ok
+                )
+                continue
             preScanList.append((full_name, display))
+        if len(preScanList) != len(items):
+            # refresh the preScanList removing duplicated entries (if any)
+            self.ui.preScanList.clear()
+            self.ui.preScanList.addModels(preScanList)
         self._localConfig['PreScanSnapshot'] = preScanList
         self._setDirty(True)
 


reszelaz (changes requested):
I agree that validating it on setting the `PreScanSnapshot` would be rather difficult. BTW we had a similar request for validating the `ScanDir` - see https://github.com/sardana-org/sardana/issues/925. Maybe we need a generic way to implement it? On the other hand, the _Return of Investment_ may not justify starting this enhancement...

But coming to this particular case, what about moving the validation to the: https://github.com/sardana-org/sardana/blob/0307007258d34055606d6e0fffccda42e7f1f33b/src/sardana/macroserver/scan/gscan.py#L835
then I think it will be recorder independent?

cpascual:
> what about moving the validation to the gscan.takeSnapshot()?

We considered it, but then it would be gscan-dependent and, in principle, you could have scans not using the gscan (or heavily customizing it), right?.

And in any case, making the nexus recorder robust against this would still be interesting in case it is call from with an unsanitized environment eventually.

IMHO the ideal solution would be implementing an specific API for the snapshot in the Door and validating/sanitizing there. But this would also be true for many other features configured via setEnv... and the API would become huge, so it may not be a good idea

reszelaz:
> We considered it, but then it would be gscan-dependent and, in principle, you could have scans not using the gscan (or heavily customizing it), right?.

It is hard for me to think about this kind of _Sardana scans_. I think if you don't use the GSF then you simply program the whole scan as a custom macro, which is of course possible. In this case, you can of course instantiate a recorder, see this [example](https://github.com/reszelaz/storage-kpi/blob/py3/sardana_scan.py), and use it. So, I agree that in the end, the Recorder class needs its own validation. I revoke my _request for change_.

> IMHO the ideal solution would be implementing an specific API for the snapshot in the Door and validating/sanitizing there. But this would also be true for many other features configured via setEnv... and the API would become huge, so it may not be a good idea

This may sound like a crazy idea, but what about providing an _entry point_ for environment variables validators. A validator could be assigned to an environment variable name and be a simple callable that receives one argument - an environment variable value. Then validators would be executed on
https://github.com/sardana-org/sardana/blob/0307007258d34055606d6e0fffccda42e7f1f33b/src/sardana/macroserver/msenvmanager.py#L479
Validators should then raise an exception when the environment variable value is not valid.

If you like this idea, as the plugins expert:), could you estimate how much effort this would be?

reszelaz:
I made manual tests. I found the following:
1. If you open an expconf when the environment variable has a duplicated name already, then it shows a pop-up with the warning, and in the list the duplicate does not appear anymore and everything looks fine. Here if you close the widget and open it again then you get the pop-up again. Is it intended to be like this?
2. The scan proceeds correctly even if there are duplicated name, and we get this warning logs:
    ```
    SardanaTP.W003 WARNING  2021-07-07 22:55:14,237 NXscanH5_FileRecorder: PreScanSnapShot: skipping duplicated label'mot01'
    SardanaTP.W003 WARNING  2021-07-07 22:55:14,238 NXscanH5_FileRecorder: Could not create link to 'mot01' in '/slit'. Reason: OSError('Unable to create link (name already exists)')
    SardanaTP.W003 WARNING  2021-07-07 22:55:14,238 NXscanH5_FileRecorder: Could not create link to 'mot01' in '/slit'. Reason: OSError('Unable to create link (name already exists)')
    ```
    I'm concerned about the one 'Unable to create link (name already exists)', is this normal?

cpascual:
> what about providing an entry point for environment variables validators?

I think that technically it is certainly possible and simple to implement, but I personally would not go that direction. I was already worried about cluttering the API with specific setters and getters for the different features... and this can potentially take it from an internal project API clutter to an inter-projects level (pluguins being a "whole project ecosystem API" of sorts...). Also one would need to deal with potentially conflicting validators introduced by different macros/recorder developers which would affect to *all* attempts (i.e. one recorder may accept only a variable value starting with a lower-case and another may require it to be a number repr).

Furthermore: the whole idea of the recorders is based in the fact that, while the data manager provides a lot of info, the recorders decide how or whether to act on it. IMHO this leads to a situation in which strictly there are only "partial loose  contracts" between specific macros and specific recorders regarding which "interfaces" they both comply.  Therefore I think that the burden of validating/sanitizing should be in the recorder, not in the macro

cpascual:
> If you open an expconf when the environment variable has a duplicated name already, then it shows a pop-up with the warning, and in the list the duplicate does not appear anymore and everything looks fine. Here if you close the widget and open it again then you get the pop-up again. Is it intended to be like this?

Yes, this was already known. The popup says that it will "deduplicate" but it  does not actually apply any change (just marks the conf as "dirty". So, unless you apply, the changes will not persist (just like with any other changes). Since the nexus recorder won't choke on it and you get the warning again, IMHO this behaviour is acceptable.

cpascual:
> I'm concerned about the one 'Unable to create link (name already exists)', is this normal?

Good catch! This is coming from the `NXscanH5_FileRecorder._populateInstrumentInfo()`. 

IMHO, I would leave it as is: it is harmless (it is an already well-handled warning) and it happens only if the pre-scan snapshot info is not sanitized (so it should not occur if you defined/edited your measurement group with defmeas or expconf)

reszelaz:
> Yes, this was already known. The popup says that it will "deduplicate" but it does not actually apply any change (just marks the conf as "dirty". So, unless you apply, the changes will not persist (just like with any other changes)

The thing is that this "dirty" marking is not enabling the "Apply" button. So to actually apply this change, you need fo make some fake changes in the expconf. So, to me, it is somehow counterintuitive. I see the pop-up, then I don't see the duplicates in the list, everything looks ok, and then I close and open again, and I see the pop-up again.

cpascual:
> The thing is that this "dirty" marking is not enabling the "Apply" button.

You are right!  I'll try to fix that.

cpascual:
I've been looking at it and didnt find an elegant solution for activating the apply button:

Setting the "dirty" flag to True does indeed activate the button, but the problem is that the flag is forced to False just after the widget has finished initialization of contents (which makes sense because all the changes done during initialization are **not** supposed to have changed configs).

The only (hackish) way I can think of is to do an extra call to the validation of the pre-scan snapshot (doing it in the init, just after the local config has been loaded).

cpascual:
I actually came out with a (IMHO) not hackish and generic solution in 7cda33d to the "apply" enabling problem

cpascual:
I now get [an error in the testsuite](https://github.com/sardana-org/sardana/pull/1637/checks?check_run_id=3028021318#step:6:1431) . I suspect it is actually the test that needs to be updated, but I am not sure. Any hints? @reszelaz

reszelaz:
> I actually came out with a (IMHO) not hackish and generic solution in [7cda33d](https://github.com/sardana-org/sardana/commit/7cda33d8117016088b390938d7e25183623494a3) to the "apply" enabling problem

Many thanks! I tested it now and it works perfectly! Good to know this generic solution, I think we could need it in the future.



> I now get [an error in the testsuite](https://github.com/sardana-org/sardana/pull/1637/checks?check_run_id=3028021318#step:6:1431) . I suspect it is actually the test that needs to be updated, but I am not sure. Any hints?

Now the CI problems are fixed. I re-run it and there were no failures. It may be some spurious error with the test tearDown().

Merging! Many thanks!

reszelaz (approved):

PR #1640, opened 2021-07-07
fix: correctly extract taurus version

reszelaz:
taurus for many years was allowing to retrieve its version from the
taurus.core.release module directly, even if this was not a recommended way.
taurus!1199 does not allow it anymore.

Change the way of obtaining the version number. In the future it will be
possible to use `taurus.__version__` directly.


cpascual:
LGTM. merging. 

Note: as a general rule,  I would recommend using `taurus.__version__` instead of `taurus.Release.version`, but in this case I think it is better as is in order to transparently support taurus<5

cpascual (commented):
LGTM

PR #1643, opened 2021-07-08
Sequencer action buttons

### LINKED ISSUE #305: sequencer action buttons (new & save) state is inconsistent (SF#460)
When a macro is added to the sequence, the "New sequence" button remains disabled but should allow to create a new sequence from the scratch. It gets enabled only after loading a sequence using the "Open sequence..." button.

When no macros remain in the sequence tree after deleting them, the "Save sequence" button remain enabled and should not - it shouldn't be possible to save an empty sequence.

Original Ticket: [sardana/tickets/460](https://sourceforge.net/p/sardana/tickets/460)

cpt-majkel:
Fixes #305 

Couple action buttons bug fixes:

- When a macro is added to the sequence, now the "New sequence"  button remains enabled.
- When no macros remain in the sequence tree after deleting them, the "Save sequence" and "Play sequence" buttons now remain disabled,
- When new sequence is created "new sequence" remain enabled and "play" button remain disabled

### DIFF
--- a/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py
+++ b/src/sardana/taurus/qt/qtgui/extra_macroexecutor/sequenceeditor/sequenceeditor.py
@@ -580,7 +580,7 @@ class TaurusSequencerWidget(TaurusWidget):
                                    Qt.QMessageBox.No) == Qt.QMessageBox.Yes:
             self.onSaveSequence()
         self.tree.clearTree()
-        self.newSequenceAction.setEnabled(False)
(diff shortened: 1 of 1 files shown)


cpascual:
As discussed with @cpt-majkel : we thought it better to change the `macroListEmpty` signal by `macroTreeChanged`, and emit it in all the places where the tree changes. Then just check the condition of emptyness in the slot

cpt-majkel:
@cpascual I've introduced discussed changes

cpascual:
I reviewed the resulting code an tested the widget with `taurus gui macrogui` and it all LGTM.

Merging.

Thanks @cpt-majkel !

Issue #1535, opened 2021-03-22
Docstring style


reszelaz:
This issue is to poll if there is any interest in the community to use a different docstring style than the reStructuredText (reST) that we currently use.

We were discussing this few days ago with @cpascual while reviewing the legacy custom sphinx extension [`sardanaextension`](https://github.com/sardana-org/sardana/blob/develop/doc/source/sphinxext/sardanaextension.py) (similar one is used in taurus).

Also, recently, in #1530 we received a contribution using the numpy style. While #1533 makes it possible to use the numpy style docstings, I think it would be better to agree on one format and use it consistenly along the project.

In [this stack overflow thread](https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format/24385103#24385103) you can find a review of the available styles (AFAIK, google and numpy are currently the strongest alternatives to the pure reST and they significantly improve the docstring's human readability).

#1534 is an example on how to convet the whole project from one format to the other using [pyment](https://github.com/dadadel/pyment/blob/master/doc/sphinx/source/pyment.rst).

cpascual:
Regarding the style, I prefer the google one over the numpy one (it is a little more concise and I prefer not using underscores).

Just one note: we should check if [python3's type annotations](https://docs.python.org/3/library/typing.html) add something to this discussion

cpascual:
> we should check if python3's type annotations add something to this discussion

... and actually, they do: https://pypi.org/project/sphinx-autodoc-annotation/

cpascual:
...and both the google and numpy styles also support type annotations out of the box. So basically type annotations can be used instead of defining the parameter and return types, regardless of choosing the basic sphinx style, the google style or the numpy style.

So one could think that this makes it irrelevant for the discussion (since they all support it)... but not really: **if** we decided to start using type annotations, then the main source of clutter in plain sphinx style (the `:type:` and `:rtype:` tags) is gone and there would be less reasons for moving to a non-default style.

dschick:
I also go for google over numpy, same arguments as @cpascual.

I also see the advantages of type annotation, but I guess this will be alot of work to add all of them to the code and do all the testing...

jangarrevoet:
I do not have a strong preference besides that the google and numpy versions are way more readable in comparison to rest.

reszelaz:
With @catunlock we were looking at how to use type annotations and how it could help in writing neat docstrings.
[...]

PR #1648, opened 2021-07-08
add support for running without an X-session

### LINKED ISSUE #1106: spock does not work without X
```> spock 
MainThread     INFO     2019-05-14 16:45:43,514 TaurusRootLogger: Using "PyQt4" for Qt
MainThread     WARNING  2019-05-14 16:45:43,617 TaurusRootLogger: Qt: : cannot connect to X server 
Spock 2.5.1-alpha -- An interactive laboratory application.

help      -> Spock's help system.
object?   -> Details about 'object'. ?object also works, ?? prints more.

IPython profile: spockdoor

Connected to Door_BL01_1
MainThread     WARNING  2019-05-14 16:45:44,061 TaurusRootLogger: Qt:  : cannot connect to X server 
```

jangarrevoet:
Fixes #1106 

Enables to run spock without an X-session

### DIFF
--- a/src/sardana/spock/magic.py
+++ b/src/sardana/spock/magic.py
@@ -31,10 +31,12 @@ __all__ = ['expconf', 'showscan', 'spsplot', 'debug_completer',
            'post_mortem', 'macrodata', 'edmac', 'spock_late_startup_hook',
            'spock_pre_prompt_hook']
 
+from sardana.util.graphics import xsession_available
 from sardana.util.whichpython import which_python_executable
 from .genutils import MSG_DONE, MSG_FAILED
 from .genutils import get_ipapi
 from .genutils import page, get_door, get_macro_server, ask_yes_no, arg_split
+
 
 
 def expconf(self, parameter_s=''):
@@ -47,6 +49,13 @@ def expconf(self, parameter_s=''):
         print("Qt binding is not available. ExpConf cannot work without it."
               "(hint: maybe you want to use experiment configuration macros? "
               "https://sardana-controls.org/users/standard_macro_catalog.html#experiment-configuration-macros)")
+        return
+
+    try:
+        if not xsession_available():
+            raise RuntimeError
+    except RuntimeError:
+        print("No X-session available. ExpConf cannot work without it.")
         return
 
     try:
@@ -97,6 +106,13 @@ def showscan(self, parameter_s=''):
         print("Qt binding is not available. Showscan cannot work without it.")
         return
 
+    try:
+        if not xsession_available():
+            raise RuntimeError
+    except RuntimeError:
+        print("No X-session available. Showscan cannot work without it.")
+        return
+
     params = parameter_s.split()
     door = get_door()
     scan_nb = None
@@ -116,6 +132,13 @@ def spsplot(self, parameter_s=''):
         from taurus.external.qt import Qt
     except ImportError:
         print("Qt binding is not available. SPSplot cannot work without it.")
+        return
+
+    try:
+        if not xsession_available():
+            raise RuntimeError
+    except RuntimeError:
+        print("No X-session available. SPSplot cannot work without it.")
         return
 
     get_door().plot()
--- a/src/sardana/spock/magic.py
+++ b/src/sardana/spock/magic.py
@@ -53,8 +53,10 @@ def expconf(self, parameter_s=''):
 
     if not display_available():
         print(
-            "Running without graphical user interface support."
-            " ExpConf cannot work without it."
+            "Running without graphical user interface support. "
+            "ExpConf cannot work without it. "
+            "(hint: maybe you want to use experiment configuration macros? "
+            "https://sardana-controls.org/users/standard_macro_catalog.html#experiment-configuration-macros)"
         )
         return
 
--- a/src/sardana/spock/magic.py
+++ b/src/sardana/spock/magic.py
@@ -51,10 +51,7 @@ def expconf(self, parameter_s=''):
               "https://sardana-controls.org/users/standard_macro_catalog.html#experiment-configuration-macros)")
         return
 
-    try:
-        if not display_available():
-            raise RuntimeError
-    except RuntimeError:
+    if not display_available():
         print(
             "Running without graphical user interface support."
             " ExpConf cannot work without it."
@@ -109,10 +106,7 @@ def showscan(self, parameter_s=''):
         print("Qt binding is not available. Showscan cannot work without it.")
         return
 
-    try:
-        if not display_available():
-            raise RuntimeError
-    except RuntimeError:
+    if not display_available():
         print(
             "Running without graphical user interface support."
             " Showscan cannot work without it."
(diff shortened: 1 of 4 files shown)


teresanunez:
Hi Jan, I have tested it and it works. My only doubt is if it could have some unexpected effect running spock on
windows (since you are checking specific things. But as I said ... it is only  a doubt, I do not have any experience
on windows with sardana.

HEnquist:
This will fail on windows unfortunately, there is no "DISPLAY" variable. But I think there will always be a display.
Maybe skip the checks if `os.name == "nt"`?

reszelaz:
Thanks @jangarrevoet for the fix! Good point @teresanunez and @HEnquist about the Windows issue. We need to make it not fail on Windows.

jangarrevoet:
I updated the PR so that the checking of the Xsession is only done on linux platforms. I have not a single windows machine (call me a lucky man) so I am unable to fix it for windows. But I assume that not having a graphical session is not a thing on windows and if so that one does not come across that very often in this scenario.

reszelaz:
Thanks for making it Windows compatible. I just checked it and it works.
I think the only missing part is to make the graphical tools give an informative message when started without X session.
Please take a look at these changes that were recently added for running Spock without PyQt: https://github.com/sardana-org/sardana/pull/1463/files.

Maybe to not duplicate the code you could encapsulate the X session presence check into a separate function (I would put it in the `sardana.util.graphic` or some better name:) I'm not good in inventing names...) and use it in `get_gui_mode()` as well as in all the other places e.g. `expconf`, `showscan`, etc (from #1463).

jangarrevoet:
ok will have a look at that.

jangarrevoet:
btw @reszelaz what was now the consensus regarding the doc string formatting?

reszelaz:
> btw @reszelaz what was now the consensus regarding the docstring formatting?

No consensus yet:) - see https://github.com/sardana-org/sardana/issues/1535 (I just add a comment there).
If you have a strong preference, to use numpy style (as the last time you contributed) feel free. We have enabled napoleon extension so it will work. numpy style docstring is better than no docstring:) If you do not have a preference, using ReST would be more consistent with the rest of sardana.

jangarrevoet:
Please check the docstring in the graphics file. I am not at all familiar with the format.
Besides that the requested changes should be implemented.

src/sardana/util/graphics.py
```diff
@@ -0,0 +1,51 @@
...
+##
+##############################################################################
+
+"""This package provides graphics related helper functions"""
+
+import os
+import sys
+
+
+def xsession_available():
```
HEnquist, line 33: I would rename this to display_available.

src/sardana/util/graphics.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Checks if an X-session is available.
```
HEnquist, line 35: This should say that it assumes that display is always available if the system isn't linux.

src/sardana/spock/magic.py
```diff
@@ -97,6 +106,13 @@ def showscan(self, parameter_s=''):
         print("Qt binding is not available. Showscan cannot work without it.")
         return
 
+    try:
+        if not xsession_available():
+            raise RuntimeError
+    except RuntimeError:
+        print("No X-session available. Showscan cannot work without it.")
```
HEnquist, line 113: Does everyone know what X-session means? Better to say "No display available."?

HEnquist (changes requested):
Looks good, I would just rename things a little.

src/sardana/util/graphics.py
jangarrevoet: I have no idea how this works on windows so there I need someone to tell me how this is done on windows.
I made the assumption that on windows you can not do any remote starting of sardana as it would be running locally. I assume you will always have an RDP or a VNC session, thus always having access to a window manager. Or are you running some headless windows machines that run sardana?

I excluded OSX since that is not in the scope of the project.

src/sardana/spock/magic.py
jangarrevoet: You are correct. I think 99.99% of the users have no idea what that means. But stating "No display available" might also confuse them since they are staring at one :-)

src/sardana/util/graphics.py
jangarrevoet: I will add a wrapper (display available) that could in the future handle also windows and make the xsession_available a private function.
Would you agree to that?

src/sardana/util/graphics.py
HEnquist: I don't think windows without a graphical display is a thing, so there is no need for any check. Just the docstring should mention that so nobody gets confused.

src/sardana/spock/magic.py
HEnquist: Yes you are right, "No display" isn't much better.. 
How about something like "Running in a text-only terminal. Showscan cannot work without a graphical display."?

guifrecuni: I think this is the kind of message the user would appreciate: A warning about "missing features"

jkotan: At DESY we don't use `Showscan` so the above message would be confusing. I would prefer general one about graphical display

jangarrevoet: I will change it to something along these lines: "Running without graphical user interface support". Fine?

teresanunez: > At DESY we don't use `Showscan` so the above message would be confusing. I would prefer general one about graphical display

I agree that instead of showscan should be written something more general, but only clarify that in DESY showscan
is indeed used in some beamlines.

teresanunez: > I will change it to something along these lines: "Running without graphical user interface support". Fine?

In my opinion, yes

jangarrevoet:
@HEnquist, changes are implemented.

src/sardana/spock/magic.py
HEnquist: I'm also fine with that.

HEnquist:
I'm happy with this now. Should we perhaps give it a quick test run on windows before merging?

jangarrevoet:
> I'm happy with this now. Should we perhaps give it a quick test run on windows before merging?

Would be good if you could.

jangarrevoet:
@teresanunez @reszelaz can someone restart CI since it seems that it has timed out on some events.

teresanunez (approved):

HEnquist:
> > I'm happy with this now. Should we perhaps give it a quick test run on windows before merging?
> 
> Would be good if you could.

I don't have a windows machine where I could test, but I think @reszelaz maybe does?

reszelaz:
I just updated this branch with the latest develop which should also fix the CI problems. 
I will try to test it on Windows tomorrow. Thanks to all!

reszelaz:
I just made some minor addition/changes. I tested it on Linux without Xsession and on Windows in normal environment. Everything works as expected. Many thanks to all! Merging.

PR #1656, opened 2021-07-12
WIP: Correcting max_vel calculation in backwards movements

teresanunez:
We get zero division errors (about once everz one hundred times and onlz with some motors) when running scans.
The problem comes from the calculation of the maximum velocity, being set to 0 if the motor moves from higher
to lower positions and by chance two terms added for computing the maximum velocity have the same absolute
value but opposite sign. One of the terms (a power) is always positive and IMO the second one has to be added
in absolute value. This PR changes the code accordingly.

### DIFF
--- a/src/sardana/util/motion/motion.py
+++ b/src/sardana/util/motion/motion.py
@@ -193,7 +193,7 @@ class MotionPath(object):
 
                 # maximum velocity possible
                 cnst = 2 * accel * decel * displacement / (decel - accel)
-                max_vel_2 = pow(motor.min_vel, 2) + abs(cnst)
+                max_vel_2 = pow(motor.min_vel, 2) + cnst
 
                 max_vel = sqrt(abs(max_vel_2))
 
--- a/src/sardana/util/motion/motion.py
+++ b/src/sardana/util/motion/motion.py
@@ -193,7 +193,7 @@ class MotionPath(object):
 
                 # maximum velocity possible
                 cnst = 2 * accel * decel * displacement / (decel - accel)
-                max_vel_2 = pow(motor.min_vel, 2) + cnst
+                max_vel_2 = pow(motor.min_vel, 2) + abs(cnst)
 
                 max_vel = sqrt(abs(max_vel_2))
 
--- a/src/sardana/util/test/test_motion.py
+++ b/src/sardana/util/test/test_motion.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+##############################################################################
+##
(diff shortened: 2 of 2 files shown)


reszelaz (changes requested):
Hi @teresanunez,
First, many thanks for your contribution and second very sorry you had to wait so long for the review. 

I still don't understand why this issue is not 100 % reproducible. In order to understand it better I have prepared a unit test to check the `MotionPath` class in 329b0eb. 

In this test I have just tried some simple examples and of course it worked. From the place where you apply the patch, I understand that it happens in _small movement_ i.e. where the motor's max velocity is not possible to reach. And I also suspect that your motor min velocity (base rate) != 0, could you confirm?

Could you please try to find which motor's parameters (accel time, decel time, min velocity, max velocity) or motion's start and end trigger the `ZeroDivisionError` in the place you try to protect? If you can not find them manually, then maybe you could leave a log traces at the beamline with these parameters and report back whenever this issue gets reproduced. 

Note, I'm able to reproduce `ZeroDivisionError` when I modify `motor01` min_vel to 1 (min_vel == max_vel), but the exception happens in another place:

```python
            else:  # Small movement
                # position where maximum velocity will be reached
                max_vel_pos = initial_pos * accel - final_pos * decel
>               max_vel_pos /= accel - decel
E               ZeroDivisionError: float division by zero

src/sardana/util/motion/motion.py:186: ZeroDivisionError
```

Also, that we don't forget, the patch would need to be applied also in the:
https://github.com/sardana-org/sardana/blob/84b6bf899e4831d90255b66c655b38fe0ef136a3/src/sardana/pool/poolcontrollers/DummyMotorController.py#L305

teresanunez:
Hi Zibi,
               no problem, thanks for looking at this now.
I am afraid I can not give much more information. We have not had this issue any more, as I said it was very
seldom and only with a couple of motors. 
What I know is that the acceleration was not equal to the decceleration.
The problem came because this number:

2 * accel * decel * displacement / (decel - accel)

was completely equal to this one:

pow(motor.min_vel, 2)

but with opposite sign, so this pure accident ... and this
implies that max_vel is 0 and we get in exception by:

at_max_vel_time = abs(at_max_vel_displacement / max_vel)

But as I said, it is completely accidentaly and seldom.

            Best Regards,
                                Teresa

reszelaz:
With @teresanunez we agreed to not integrate this solution yet and wait until it gets reproduced again. Meanwhile I integrate the automatic tests.

reszelaz (approved):

PR #1652, opened 2021-07-09
Add ascanct parameter output

johanfforsberg:
Adds a table with some helpful output of the motor settings that are about to be used in a continuous scan. 

In order to implement this, we had to change how the header of the scan table is output; it is now printed when the first point arrives, and not immediately. This also prevents some issues with info/debug messages appearing after the header.

The original issue #692 also suggests showing the restored settings after the scan but for similar reasons (see issue #1651) it is currently commented out. It can be added in whenever that issue is solved.

Example:
```
Door_2_1 [5]: ascanct mot10  0 10 10 0.1 0.1
ScanDir is not defined. This operation will not be stored persistently. Use "expconf" or "newfile" to configure data storage (or eventually "senv ScanDir <abs directory>")
Scan #31 started at Thu Jul  8 13:05:35 2021. It will take at least 0:00:00
Motor positions and relative timestamp (dt) columns contains theoretical values
 Motor    Velocity   Acceleration   Deceleration  Start   End  
 mot10       5            2              2          -5     16  

 #Pt No    mot10      ct09      ct10      ct11      ct12       dt   
   0         0        0.1       0.2       0.3       0.4        2    
...
```
Opinions on how the output should be formatted are very welcome.

Fixes #692

PR #1657, opened 2021-07-12
revert control_exceptions arg to waitForEvent

13bscsaamjad:
Different types of exceptions are not possible here so there is no need for the control_exceptions argument to Taurus' waitforEvent. Ref to [PR](https://gitlab.com/taurus-org/taurus/-/merge_requests/1204) in Taurus. Reverting here as well.


reszelaz (approved):
Thank your @13bscsaamjad !
I merge even if the CI is broken. There are 6 tests that failed, but I think this is due to https://github.com/sardana-org/sardana/pull/1652#issuecomment-878296177

PR #1659, opened 2021-07-15
Improve devcontainer configurations

13bscsaamjad:
- removed .env for .devcontainer as it does not work for all docker versions
- use mambaforge to easily intall packages from conda-forge using mamba
- installed system packages in the sardana image to be able to use X apps like jive, expconf
- installed jive in the sardana image which is really handy for dev when multiple clients are required
- updated the environment file for .devcontainer to install pyqt related stuff required for expconf etc.

### DIFF
--- a/.devcontainer/.env
+++ b/.devcontainer/.env
@@ -1,3 +1,0 @@
-DOCKER_REGISTRY_HOST=nexus.engageska-portugal.pt
-DOCKER_REGISTRY_USER=ska-docker
-CONTAINER_NAME_PREFIX=sardana_dev_
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -20,7 +20,7 @@ services:
     hostname: sardana
     environment:
       TANGO_HOST: databaseds:10000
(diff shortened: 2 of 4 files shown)


catunlock:
Great improvement @13bscsaamjad. I tested it in my Linux, everything works. LGTM.

catunlock (approved):

reszelaz (approved):
Thanks @13bscsaamjad for your contribution. I just tried it and it works.
I just have two observations:
1. I still need to enable _broad_ access to my host's X server e.g. `xhost +local:` otherwise i get:
    ```
    (base) root@sardana:/workspace# xeyes
    No protocol specified
    Error: Can't open display: unix:0
    ```
    I think it has to do with the fact that within the container we run apps as root. Do you have any other solution to that?
2. It is not related to this PR, but I just realized that the tangodb container mounts a volume for the mysql databases files. This makes the tango DB persistent even if we recreate the tangodb container. Was it intentional?

I proceed with the merge anyway, but it would be nice to discuss on those two points at some time. Thanks again!

PR #1663, opened 2021-07-15
fix: lsmeas use name instead of channel label

### LINKED ISSUE #1053: renameelem does not change the MG configuration
The macro does not change the label and name of the channel on the MG configuration. 
To reproduce: After to rename a experimental channel run lsmeas.

catunlock:
When an element is renamed using renameelem the macro lsmeas still shows the previous name (label) #1053.

We propose to acquire and show the name of the elements of the measurement group from the full name (i.e: ct1, ct2, sys/tg_test/1/ampli).

closes #1053.

### DIFF
--- a/src/sardana/macroserver/macros/lists.py
+++ b/src/sardana/macroserver/macros/lists.py
@@ -26,6 +26,7 @@
 """This is the lists macro module"""
 
 
+from taurus.core.tango.tangovalidator import TangoAttributeNameValidator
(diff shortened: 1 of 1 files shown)


13bscsaamjad:
the call `getTimerName()` is still returning the old timer name!

catunlock:
Yes, but it's a deprecated method and with @reszelaz we decided that probably the best we can do is remove the timer column from the macro. So it shouldn't be a problem for now.

I will make another pull request removing the timer column as soon as this one gets accepted.

13bscsaamjad (approved):

Issue #1095, opened 2019-04-11
Memory leak in scan macros


jkotan:
Hi, 

I've just tested MacroServer with 1D channel detector (MCA) and after a less than one day running MacroServer uses 4.6g memory (It started to be proportional to number of point in my scan I've performed). 

To look at this problem in more details I create macros:
```python
@macro()
def getobjgraph(self):

    import gc
    import objgraph
    mystdout = StringIO()
    objgraph.show_most_common_types(file=mystdout)
    self.output(str(mystdout.getvalue()))


@macro()
def gcdump(self):

    import gc
    gc.set_debug(gc.DEBUG_LEAK)
    gc.collect()
    gres = gc.garbage
    self.output("GARBAGE %s" % len(gres))
    res = str(gres)
    import time
    fname = "/tmp/gcdump-%s.gc" % str(time.time())
    self.output(fname)
    fl = open(fname, "w")
    fl.write(res)
    fl.close()
```  
from `gcdump` where I see that a number of garbages grows proportionally to a number of scan points I've performed (with proportional factor: 3)

On the other hand from `getobjgraph`  I see that .e.g. 
```
p09/door/haso228jk.01 [36]: %getobjgraph
dict               86524
CaselessDict       71883
Record             71828
function           27811
list               15452
tuple              12674
weakref            7533
wrapper_descriptor 5856
getset_descriptor  4476
type               3116
```
where `Record`, `CaselessDict` and `dict` are amost equal (up to a constant factor) to number of scan points I've performed with `ascan` macro. 

Any idea what it comes from? Something is not deleted? (I use sardana 2.5.1)

reszelaz:
Thanks for the report @jkotan !

I have not yet tested your macros - I will certainly try them! But I have already some questions.
In your test do you run a single long scan? During the scan all the records are kept in the memory. Furthermore, by default, they are even kept in the memory after the scan finish in order to allow the scan data analysis in the subsequent macro. This second feature can be however disabled with [PreserveMacroData](https://sardana-controls.org/users/environment_variable_catalog.html#preservemacrodata) environment variable.

If this still does explain this issue, for example you run a sequence of short scans and the memory sill grows, this may be related to some cyclic references that makes the life harder to the garbage collector. #816 tries to fix them. This PR was already approved but before merging it I wanted to carefully review the public API that this PR touches to not break anything. It should be available soon.

jkotan:
Thanks @reszelaz ,
[...]

PR #1664, opened 2021-07-15
Hotfix 3.1.2 - fix memory leak in continuous scans

reszelaz:
#1095 documents MacroServer memory growing problems when performing numerous long scans. The last investigation [demonstrated a significant improvement when using jemalloc](https://github.com/sardana-org/sardana/issues/1095#issuecomment-596144102). Those tests were done on timescan, and I assumed that the continuous scans should behave the same since these scans reuse the same implementation of the data handling. But this assumption turned out to be wrong, as demonstrated in this trends:

(All examples in this description use measurement group with 20 software synchronized dummy counter/timers)

- `repeat 10 "ascanct mot01 0 100 10000 0.001"`  - increas of 260 MB!
  10ascanct_10kpoints_20channels_develop
- `repeat 10 "meshct mot01 0 50 5000 mot02 0 1 1 0.001"` - increase of 260 MB 
  10meshct_10kpoints_20channels_develop

This PR provides a hotfix release 3.1.2 which solves the above-mentioned memory leak.

**Technical explanation**:

`aNscan`, which holds a hard reference to the `CTScan`, adds its bound method as hook in the waypoint generator.
The enumeration object created from the waypoint generator is then hard referenced in the `CTScan`. Hence, a cycle reference
is created between `CTScan` and `aNscan` objects (I imagine that the last yielded waypoint, which contains the bound method, must be kept in the enumeration object). Then due to unknown reasons, the scan memory can not be released if this cycle reference exists (even if I believe that at some point the GC should collect it). This cycle reference was demonstrated using [this reduced example](https://gist.github.com/reszelaz/389fa8d262b5206879e98b0518d4be10).

**Introduced change**:

Follow the example of the `timescan` macro and `TScan` class in order to not pass  `_fill_missing_records()` as hook. Refactor the code and define the `_fill_missing_record()` in the `CAcquisition` class (base class of both `CTScan` and `TScan`). This way the memory is correctly released for `ascanct`, `a2scanct`, `meshct`, etc. macros.

This PR solves the memory leak in the Sardana as demonstrate on this examples:

- `repeat 10 "ascanct mot01 0 100 10000 0.001"`
  10ascanct_10kpoints_20channels_hotifx-3 1 2
- `repeat 10 "meshct mot01 0 50 5000 mot02 0 1 1 0.001"`
  10meshct_10kpoints_20channels_hotifx-3 1 2

...but the following way of developing a macro could still trigger it:

```python
from sardana.macroserver.macro import Macro


class MyAscanct(Macro):

    def hook(self):
        self.print("In MyAscanct.hook()")

    def run(self):
        self.ascanct, _ = self.prepareMacro("ascanct", "mot01", 0, 100, 10000, 0.001)
        self.ascanct.appendHook((self.hook, ["post-move"]))
        self.runMacro(self.ascanct)
```

In order to avoid it I see the following two options:
1. make [`GScan.steps`](https://github.com/sardana-org/sardana/blob/f0ac950f1dddf14b625c410e515da1fbb6155248/src/sardana/macroserver/scan/gscan.py#L992) property do not keep a reference to the enumeration object 
2. make `Hookable` class keep weak references to the hooks

@tiagocoutinho, @rhomspuron, @jairomoldes I know you suffered these memory leaks, it would be nice if you could review the changes and make tests.
@13bscsaamjad, have you observed this at MAXIV?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 3.1.1
+current_version = 3.1.2-alpha
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
 serialize = 
 	{major}.{minor}.{patch}-{release}
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 3.1.2-alpha
+current_version = 3.1.2
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
 serialize = 
 	{major}.{minor}.{patch}-{release}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the formats and conventions from [keepachangelog.com]
 
-## [3.1.2] 2021-07-16
+## [3.1.2] 2021-08-02
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 This file follows the formats and conventions from [keepachangelog.com]
 
+## [3.1.2] 2021-07-16
+
+### Fixed
+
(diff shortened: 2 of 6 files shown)


guifrecuni:
Thanks @reszelaz for such a detailed explanation, I really appreciate it!

13bscsaamjad:
It has not came to our attention here yet.

dschick:
Hi, I just tested the hotfix (merged into the latest develop branch) in a docker-container and the memory is still increasing when doing `ascan`s with `DummyTwoD` images.
I am not sure if this PR should also fix this issue for `ascan`s?

reszelaz:
Thanks for the tests @dschick , this PR only fixes continues scans. We are still working on the issue with 2D as you reported.

reszelaz:
@sardana-org/integrators could you please take a look on this one. We won't get feedback from ALBA's beam lines soon and IMO it urgent to get merged.

teresanunez (approved):

reszelaz:
Thanks, @teresanunez for the review! Merging!

tiagocoutinho:
Hi,
Upgraded today on a beamline and ran some continuous scans.
I don't see the leak anymore.
Nice work! Thanks

PR #1666, opened 2021-07-23
Step scan generator has now the possibility to not move motors. Fix #1159

### LINKED ISSUE #1159: Step scan generator with the possibility to not move motors

jairomoldes:
Note that this also fixes a bug in returnObj() of macro.py

Please find attached 2 useless test macros to check the fix works fine.
[useless.txt](https://github.com/sardana-org/sardana/files/6870004/useless.txt)

It worked as expected in a test environment running:
ascan mot01 0 100 10 0.1
u2scan mot03 0 100 gap01 30 40 8 0.1
uscan mot01 0 10 9 0.1
uscan gap01 0 10 9 0.1

I also suggest that maybe it will be nice to adapt the aNscan family in order to allow a new "repeat" parameter

### DIFF
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -1173,24 +1173,63 @@ class SScan(GScan):
         self._env['acqtime'] = self._sum_acq_time
 
     def stepUp(self, n, step, lstep):
-        motion, mg = self.motion, self.measurement_group
+        mg = self.measurement_group
         startts = self._env['startts']
 
-        # pre-move hooks
-        for hook in step.get('pre-move-hooks', ()):
-            hook()
-            try:
-                step['extrainfo'].update(hook.getStepExtraInfo())
-            except InterruptException:
-                raise
-            except Exception:
-                pass
+        def pre_move_hooks():
+            # pre-move hooks
+            for hook in step.get('pre-move-hooks', ()):
+                hook()
+                try:
+                    step['extrainfo'].update(hook.getStepExtraInfo())
+                except InterruptException:
+                    raise
+                except Exception:
+                    pass
 
         # Move
         self.debug("[START] motion")
         move_start_time = time.time()
         try:
-            state, positions = motion.move(step['positions'])
+            nans = []
+            for idx, step_value in enumerate(step['positions']):
+                if step_value is None or np.isnan(step_value):
+                    nans.append(idx)
+            if len(nans) == 0:  # all motors will be moved
+                # if number of motors to move has changed update macro info
+                if len(self.motion.moveable_list) != len(step['positions']):
+                    self.macro.returnObj(self._motion)
+                    motion = self.macro.getMotion(
+                        [m.name for m in self.moveables])
+                    self._motion = motion
+                    self.macro.motors = self.moveables
+                pre_move_hooks()
+                state, positions = self.motion.move(step['positions'])
+            elif len(nans) == len(step['positions']):  # no motor will be moved
+                state = self.motion.readState()
+                positions = self.motion.readPosition()
+            else:  # only some motors will be moved
+                # if number of motors to move has changed update macro info
+                if len(self.motion.moveable_list) != len(step['positions']):
+                    moveables = []
+                    motors_moved = []
+                    motors_positions = []
+                    for idx, moveable in enumerate(self.moveables):
+                        if idx not in nans:
+                            moveables.append(moveable)
+                            motors_moved.append(moveable.name)
+                            motors_positions.append(step['positions'][idx])
+                    self.macro.returnObj(self._motion)
+                    motion = self.macro.getMotion(motors_moved)  # does addObj
+                    self._motion = motion
+                    self.macro.motors = moveables
+                pre_move_hooks()
+                state, positions = motion.move(motors_positions)
+                for idx in nans:  # fill the gaps of not moved motors
+                    moveable = self.moveables[idx]
+                    motor = self.macro.getMoveable(moveable.name)
+                    pos = motor.position
+                    positions.insert(idx, pos)
             self._sum_motion_time += time.time() - move_start_time
             self._env['motiontime'] = self._sum_motion_time
         except InterruptException:
--- a/src/sardana/macroserver/scan/gscan.py
+++ b/src/sardana/macroserver/scan/gscan.py
@@ -199,8 +199,9 @@ class GScan(Logger):
     The generator must be a function yielding a dictionary with the following
     content (minimum) at each step of the scan:
 
-    - 'positions' : In a step scan, the position where the moveables
-      should go
+    - 'positions' : In a step scan, a sequence with position(s) where 
+      the moveable(s) should go. ``None`` or ``float("NaN")`` in the sequence
+      means do not move a given moveable.
     - 'integ_time' : In a step scan, a number representing the integration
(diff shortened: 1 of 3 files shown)


jairomoldes:
```javascript
from sardana.macroserver.macros.scan import ascan, a2scan


class uscan(ascan):

    repeats = 3


    def hook(self):
        for name in [m.name for m in self.motors]:
            self.execMacro('wm %s' % name)

    def _stepGenerator(self):
        self.hooks = [(self.hook, ['pre-move'])]
        step = {}
        step["integ_time"] = self.integ_time
        step["pre-move-hooks"] = self.getHooks('pre-move')
        step["post-move-hooks"] = self.getHooks('post-move')
        step["pre-acq-hooks"] = self.getHooks('pre-acq')
        step["post-acq-hooks"] = self.getHooks('post-acq') + self.getHooks(
            '_NOHINTS_')
        step["post-step-hooks"] = self.getHooks('post-step')

        step["check_func"] = []
        self.nb_points = (self.nr_interv + 1) * self.repeats
        for point_no in range(self.nb_points):
            if (point_no % self.repeats) == 0:
                step["positions"] = self.starts + int(point_no/self.repeats) * self.interv_sizes
            else:
                step["positions"] = [None for i in range(len(self.starts))]
            step["point_id"] = point_no
            yield step


class u2scan(a2scan):

    repeats = 3

    def hook(self):
        for name in [m.name for m in self.motors]:
            self.execMacro('wm %s' % name)

    def _stepGenerator(self):
        self.hooks = [(self.hook, ['pre-move'])]
        step = {}
        step["integ_time"] = self.integ_time
        step["pre-move-hooks"] = self.getHooks('pre-move')
        step["post-move-hooks"] = self.getHooks('post-move')
        step["pre-acq-hooks"] = self.getHooks('pre-acq')
        step["post-acq-hooks"] = self.getHooks('post-acq') + self.getHooks(
            '_NOHINTS_')
        step["post-step-hooks"] = self.getHooks('post-step')

        step["check_func"] = []
        self.nb_points = (self.nr_interv + 1) * self.repeats
        for point_no in range(self.nb_points):
            if (point_no % self.repeats) == 0:
                step["positions"] = self.starts + int(point_no/self.repeats) * self.interv_sizes
            else:
                step["positions"] = self.starts + int(point_no/self.repeats) * self.interv_sizes
                step["positions"] = [None for i in range(len(self.starts))]
#                step["positions"][-1] = None
            step["point_id"] = point_no
            yield step

```

reszelaz:
@teresanunez could you please review this PR, we need it for #1481. Many thanks!

teresanunez:
Hi, I have tested it and the repetition of points works. The hook defined in the macro uscan is called, but
if I define a general hook (using defgh) this is not called. I guess it should.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -1173,24 +1177,63 @@ def scan_loop(self):
...
+                # if number of motors to move has changed update macro info
+                if len(self.motion.moveable_list) != len(step['positions']):
+                    self.macro.returnObj(self._motion)
+                    motion = self.macro.getMotion(
+                        [m.name for m in self.moveables])
+                    self._motion = motion
+                    self.macro.motors = self.moveables
+                pre_move_hooks()
+                state, positions = self.motion.move(step['positions'])
+            elif len(nans) == len(step['positions']):  # no motor will be moved
```
reszelaz, line 1212: This case will execute the `post-move` hooks and should not. So, I would do the same as you have done for the `pre-move` hooks and call the `post-move` hooks only in the two other cases.

src/sardana/macroserver/scan/gscan.py
```diff
@@ -1173,24 +1177,63 @@ def scan_loop(self):
...
+                    nans.append(idx)
+            if len(nans) == 0:  # all motors will be moved
+                # if number of motors to move has changed update macro info
+                if len(self.motion.moveable_list) != len(step['positions']):
+                    self.macro.returnObj(self._motion)
+                    motion = self.macro.getMotion(
+                        [m.name for m in self.moveables])
+                    self._motion = motion
+                    self.macro.motors = self.moveables
+                pre_move_hooks()
```
reszelaz, line 1210: Here, exceptions from the hooks would go through a special handling with `self.dump_information(n, step, self.motion.moveable_list)` being called before re-raising. This is not necessary and is a change with respect to the previous implementation. This special handling was foreseen only for the exceptions being raised during the motion itself.

reszelaz (changes requested):
Many thanks for your contribution @jairomoldes and for your review @teresanunez!

https://github.com/sardana-org/sardana/pull/1666#issuecomment-890864664 refers to the example macros: `uscan` and `u2scan` which you attached. And @teresanunez is right that those two macros does not respect the _general hooks_ (you override the hooks property instead of appending the hooks:

```python
self.hooks = [(self.hook, ['pre-move'])]
```

But, it is a comment about your examples, and not the feature you have implemented. I have tested the feature and it works really nice! I just left two inline comments, which IMO are things to fix. 

I let myself to fix the `GScan._estimate()` method which was not considering the `None` in the "positions" key of a step (see 9602691c). You have not found this bug because your example macros were inheriting from `ascan` which has its own estimation mechanism. However when I tried to adapt the `ascanr` example macro to this new feature (see 60fb7075b) I found this bug. 

Finally, I have added a short explanation of this new feature to the `GScan` docstring (see e1c641c).

> I also suggest that maybe it will be nice to adapt the aNscan family in order to allow a new "repeat" parameter.

It is an option. We also have the `ascanr` example macro. I don't know if we should extend the `ascan` to accept a new parameter or provide `ascanr`, etc. Let's see what other people think about it. Here is an example of the `ascanr` execution:

```
Door_zreszela_1 [7]: ascanr mot01 0 1 1 0.1 2
Scan range is not defined for mot01 and could not be verified against motor limits.
Operation will be saved in /tmp/test2d.h5 (HDF5::NXscan from NXscanH5_FileRecorder)
Scan #7 started at Fri Oct 29 09:25:26 2021. It will take at least 0:00:00.682843
 #Pt No    mot01      ct01      ct02      ct03      ct04    repetition     dt   
   0         0        0.1       0.2       0.3       0.4         0       0.0479479
   1         0        0.1       0.2       0.3       0.4         1       0.173388
   2         1        0.1       0.2       0.3       0.4         0       0.581513
   3         1        0.1       0.2       0.3       0.4         1       0.698321
Operation saved in /tmp/test2d.h5 (HDF5::NXscan)
Scan #7 ended at Fri Oct 29 09:25:26 2021, taking 0:00:00.817228. Dead time 49.8% (setup time 1.2%, motion dead time 37.8%)
```
Note the extra column "repetition".

I also realized that the "Pt No" column always increases, even if the step generator returns a fixed "point_id" for all repetitions. This is because when iterating over the generator we also enumerate and we use the enumerate counter when storing the data instead of what the contains the generated step. I don't have a strong opinion but it looks like a bug. What do you think? Anyway, if you consider it as a bug, I would treat it in a separate issue.

PR #1667, opened 2021-07-23
Fix udef elem expconf

reszelaz:
This PR fixes expconf issues introduced by #1615 and reported by @yimli in https://github.com/sardana-org/sardana/pull/1615#issuecomment-885535432 - see 3d146b6
One idea to improve it would be to somehow mark the row corresponding to a disabled and undefined channel in the expconf e.g. red background. Does someone know how to do it?

This PR also fixes one more issue with restoring the memorized value of the measurement group's configuration attribute containing disabled but undefined channels - see 244b5f2.

@yimli it would be great if you could test if it fixes the issues you observed. Many thanks!

### DIFF
--- a/src/sardana/pool/poolmeasurementgroup.py
+++ b/src/sardana/pool/poolmeasurementgroup.py
@@ -803,6 +803,9 @@ class MeasurementConfiguration(object):
                     try:
                         channel = pool.get_element_by_full_name(ch_name)
                     except KeyError:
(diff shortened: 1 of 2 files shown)


yimli:
Thanks @reszelaz ! I have tested it and no key error anymore. But the undefined channel is still shown in the channel list. I agree with you it would be a good idea to mark the undefined channel, so that users would know it from expconf that this one has been undefined.

PR #1669, opened 2021-07-28
Fix: Avoid cycle reference between h5 recorder and macro

saidaHF:
h5 recorder creates hard reference to macro object in _startRecordList()
method. This closes the cycle reference between macro, GScan and recorder
since GScan has hard references to the recorders and macro has hard
references to the GScan.

Use weakref instead of hard references between recorder and macro.

image

This trend shows how memory stays stable during 10 scans with 2d channels

@dschick @felix92 

This PR should fix the issue you reported in https://github.com/sardana-org/sardana/issues/1095#issuecomment-875706639

### DIFF
--- a/src/sardana/macroserver/recorders/h5storage.py
+++ b/src/sardana/macroserver/recorders/h5storage.py
@@ -35,6 +35,7 @@ __docformat__ = 'restructuredtext'
 import os
 import re
 import posixpath
+import weakref
 from datetime import datetime
 import numpy
 import h5py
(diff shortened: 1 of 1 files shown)


reszelaz (approved):
Many thanks, @saidaHF! Well spotted!
In the past I proposed a PR which was fixing the same issue for other recorders - #816. At that time I missed the bug in the NXscanH5_FileRecorder.

dschick:
many thanks @saidaHF.
@felix92 is currently testing it.
As far as I know, the problem does not exist with the NXscanH5_FileRecorder. So this would be consistent.

Best

Daniel

Issue #1665, opened 2021-07-20
[Doubt] Is this line necessary?


marc2332:
Hello,

I wanted to use `BaseDoor` class with a custom output, but, I found out that its `write()` function overwrites the _output_stream property for no apparent reason:

https://github.com/sardana-org/sardana/blob/3eb154ccc4b25001925c4a48ce060d7cfda7e496/src/sardana/taurus/core/tango/sardana/macroserver.py#L763

Is it ok? I tried removing that line and it worked as I wanted.

Thanks 👍🏻

reszelaz:
Thanks for the report @marc2332 ! For me it looks like we don't need this like. It would be great if you could provide a PR. Thanks!

PR #1670, opened 2021-07-29
Fix: Remove unnecessary property overwrite

marc2332:
(Related to https://github.com/sardana-org/sardana/issues/1665)

This PR remove an unnecessary property overwrite. It's unnecessary because it already assigns a value in the constructor:https://github.com/sardana-org/sardana/blob/3eb154ccc4b25001925c4a48ce060d7cfda7e496/src/sardana/taurus/core/tango/sardana/macroserver.py#L337
So, it makes it impossible to make a custom output. The solution is just to remove the line.

😄 Thanks


reszelaz (approved):
LGTM, merging! Many thanks @marc2332 !

PR #1672, opened 2021-07-30
fix: backward compatibility for meas grp conf value_ref_*

reszelaz:
#867 introduced backward compatibility layer for measurement group
configurations with non-referable channels containing value_ref_*
parameters. This compatibility layer assumes that in this case both
parameters: value_ref_enabled and value_ref_pattern are present, which is
not always true.
Fix it by considering also the case when only one of these parameters is
present.

@ByRellex could you please confirm this fixes the issue you have discovered? As soon as you confirm and the CI is green I will auto-merge. Thanks!


ByRellex:
Hi,
Yes te patch solves the issue and the beamline passes the acuisition tests with the measurement group that issued the problem.
Cheers, Alejandro Alonso

reszelaz:
Thanks for the feedback @ByRellex . Merging.

PR #1673, opened 2021-08-01
Solving amultiscan arguments error

teresanunez:
The current amultiscan macro does not work due to a problem with the input parameters.
This PR solve it, perhaps not in a very elegant way, but we needed a solution urgently.


reszelaz (approved):
Thanks, @teresanunez!
I just fixed the problem with the `nr_interv` and `integ_time` parameters (you forgot to pass them to the prepare method).
And after recovering old notes from https://github.com/sardana-org/sardana/issues/4#issuecomment-262972081, I refactored the code. Now, instead of iterating over the param repeat values, we use `zip` to extract motors, starts and ends.

PR #1675, opened 2021-08-04
fix: reduce memory leak of scans

reszelaz:
Macros are executed in worker threads of a ThreadPool of size 10.
Each macro may end up being executed by a different thread.
In the investigation of #1095 we observed that memory allocated
in small chunks in different threads is not released to the OS.
Reduce the memory consumption by always executing macros
in the same thread (per door).

With this PR the 8 repetitions of:
```
for i in range(10): 
    %ascan mot01 0 100 100 0.001
```
does not exceed (+/-) the memory used by the first scan already:

8x10_ascan_100_points_10_scalar_1_spectrum_1_image_thread_pool_1_thread

While in the develop branch (the same tests) it keeps growing:

8x10_ascan_100_points_10_scalar_1_spectrum_1_image_thread_pool_10_threads

Regarding the tests using the develop branch, I'm not sure the memory grows only until the amount of the memory of one scan multiplied by the number of worker threads (10) or until filling the available memory (there was no more free memory in the system at the end of the test).

@dschick, @felix92 could you please give it a try? Many thanks!

### DIFF
--- a/src/sardana/macroserver/msmacromanager.py
+++ b/src/sardana/macroserver/msmacromanager.py
@@ -50,6 +50,7 @@ from collections import OrderedDict
 
 from taurus.core.util.log import Logger
 from taurus.core.util.codecs import CodecFactory
+from taurus.core.util.threadpool import ThreadPool
 
 from sardana.sardanadefs import ElementType
 from sardana.sardanamodulemanager import ModuleManager
@@ -1099,6 +1100,11 @@ class MacroExecutor(Logger):
         name = "%s.%s" % (str(door), self.__class__.__name__)
         self._macro_status_codec = CodecFactory().getCodec('json')
         self.call__init__(Logger, name)
+        # ThreadPool to execute macros asynchronously
+        self._thread_pool = ThreadPool(Psize=1)
+    
+    def add_job(self, job, callback=None, *args, **kw):
+        self._thread_pool.add(job, callback, *args, **kw)
 
     def getDoor(self):
         return self._door
@@ -1594,7 +1600,7 @@ class MacroExecutor(Logger):
 
         if asynch:
             # start the job of actually running the macro
-            self.macro_server.add_job(self.__runXML, self._jobEnded)
+            self.add_job(self.__runXML, self._jobEnded)
             # return the proper xml
             return self._xml
         else:


reszelaz:
> Regarding the tests using the develop branch, I'm not sure the memory grows only until the amount of the memory of one scan multiplied by the number of worker threads (10) or until filling the available memory (there was no more free memory in the system at the end of the test).

I have made a test to verify this. Still using the develop branch I hardcoded PoolThread(Psize=5) in the MacroServer:

```
diff --git a/src/sardana/macroserver/macroserver.py b/src/sardana/macroserver/macroserver.py
index 3c4842493..da92c90ed 100644
--- a/src/sardana/macroserver/macroserver.py
+++ b/src/sardana/macroserver/macroserver.py
@@ -146,6 +146,7 @@ class MacroServer(MSContainer, MSObject, SardanaElementManager, SardanaIDManager
         self._pools = CaselessDict()
         self._max_parallel_macros = self.MaxParalellMacros
         self._path_id = None
+        self._th_pool = None
 
         MSContainer.__init__(self)
         MSObject.__init__(self, full_name=full_name, name=name, id=InvalidId,
@@ -168,8 +169,10 @@ class MacroServer(MSContainer, MSObject, SardanaElementManager, SardanaIDManager
         return kwargs
 
     def add_job(self, job, callback=None, *args, **kw):
-        th_pool = get_thread_pool()
-        th_pool.add(job, callback, *args, **kw)
+        if self._th_pool is None:
+            from taurus.core.util.threadpool import ThreadPool
+            self._th_pool = ThreadPool(Psize=5)#get_thread_pool()
+        self._th_pool.add(job, callback, *args, **kw)
 
     # --------------------------------------------------------------------------
     # Environment DB related methods
```

And the memory grows only proportionally to the number of threads, while still having available memory:

2x10_ascan_100_points_10_scalar_1_spectrum_1_image_thread_pool_5_threads.

So, it confirms all the previos findings.

felix92:
Hi,
I tested the PR and the changes are working fine. I have the same results as @reszelaz.

So I think you can merge it.

Thanks,
Felix

reszelaz:
@felix92, many thanks for your tests!
I proceed with the merge.

PR #1694, opened 2021-09-22
fix: `not DevFailed` exception catching

reszelaz:
`not DevFailed` evaluates to False and can not be used to catch
exceptions. Otherwise we get "TypeError: catching classes that do
not inherit from BaseException is not allowed"

I will auto-merge since this is a trivial change.


reszelaz:
Thanks @HEnquist for the feedback!

PR #1695, opened 2021-09-22
Allow reading Door Status when macros are stopping

HEnquist:
This fixes a deadlock that happens if someone happens to read Status of the Door exactly while a scan is stopping. 
The problem is that the Status command takes the TangoMonitor of the Door device server, and then dev_status calls methods that are decorated by mAPI. This makes it wait until the running macro has stopped. It's holding the Monitor for this entire time and that causes all kinds of trouble. For example the logging fails on `push_change_event()`, and there are change events on other attributes as well during the stopping procedure. Every such attempt will fail with:
```
PyTango.DevFailed: DevFailed[
DevError[
    desc = Not able to acquire serialization (dev, class or process) monitor
  origin = TangoMonitor::get_monitor
  reason = API_CommandTimedOut
severity = ERR]
]
```

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -2239,6 +2239,12 @@ class Macro(Logger):
         return self._parent_macro
 
     @property
+    def _command(self):
+        """**Unofficial Macro API**. Alternative to getCommand that does not
+        throw StopException in case of a Stop."""
+        return '%s %s' % (self._getName(), ' '.join([str(p) for p in self._in_pars]))
+
+    @property
     def description(self):
         """**Unofficial Macro API**. Alternative to :meth:`getDescription` that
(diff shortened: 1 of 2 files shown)


src/sardana/macroserver/macro.py
```diff
@@ -2238,6 +2238,12 @@ def parent_macro(self):
         internally by the *Executor*"""
         return self._parent_macro
 
+    @property
+    def _command(self):
```
HEnquist, line 2242: What should this be called? Just `command` feels dangerous, big risk of clashing with something in some macro. And the underscore doesn't match the rest of the unofficial api, while`get_command` should be a function rather than a property.
A few other alternatives: `command_string`, `macro_command`, `formatted_command`...

PR #1698, opened 2021-09-30
Add gitlab-ci pipeline

beenje:
Prepare migration to GitLab.

Pipeline was tested here: https://gitlab.com/beenje/sardana/-/pipelines/379922351
And here with a dummy tag: https://gitlab.com/beenje/sardana/-/pipelines/379934485
The release to PyPi of course failed because I didn't define any credentials. The variables `TWINE_USERNAME` and `TWINE_PASSWORD` shall be defined in the CI/CD settings of the project.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,66 @@
+---
+stages:
+  - build
+  - test
+  - deploy
+
+build-pypi-package:
+  stage: build
+  image: python:3.9
+  script:
+    - python setup.py check sdist bdist_wheel
+  artifacts:
+    expire_in: 1 day
+    paths:
+      - dist/
+
+run-test:
+  stage: test
+  image: reszelaz/sardana-test
+  before_script:
+    - python3 -m pip install dist/sardana-*.whl
+    - export TANGO_HOST=$(hostname):10000
+    - /usr/bin/supervisord -c /etc/supervisor/supervisord.conf
+    - sleep 10
+    - supervisorctl start Pool
+    - supervisorctl start MacroServer
+  script:
+    - xvfb-run -s '-screen 0 1920x1080x24' /bin/bash -c "pytest /usr/local/lib/python3.5/dist-packages/sardana"
+
+build-doc:
+  stage: test
+  image: reszelaz/sardana-test
+  script:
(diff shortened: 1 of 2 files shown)


cpascual:
LGTM.
I like that you added the build_pypi_package job and then used it for the tests

cpascual (approved):

beenje:
Thanks @cpascual. We should probably even set `GIT_STRATEGY: none` in the test job. I had one test failing when I tried. Probably unrelated.

I looked at taurus for building the doc. Really nice. I'm gonna use that in other projects (making a link to the artifacts)!

.gitlab-ci.yml
```diff
@@ -0,0 +1,66 @@
...
+  image: alpine
+  variables:
+    GIT_STRATEGY: none
+  script:
+    - echo "Deploying already-built docs"
+  artifacts:
+    paths:
+      - public
+  rules:
+    - if: '$CI_COMMIT_TAG =~ /^[0-9]+\.[0-9]+\.[0-9]+.*$/'
```
reszelaz, line 56: I see that this is a change with respect to what we have now. Currently, the Sardana docs are pointing to the latest develop version - see [Docs for other Sardana versions](https://sardana-controls.org/other_versions.html). Since we can now easily have access to multiple versions, I would do the change you propose, so the Sardana docs point to the latest tag. But then we would need to rewrite Docs for other Sardana versions (I would simply copy and paste [Docs for other Taurus versions](https://taurus-scada.org/other_versions.html).

reszelaz (approved):
Many thanks, @beenje!

reszelaz:
> We should probably even set `GIT_STRATEGY: none` in the test job. I had one test failing when I tried. Probably unrelated.

I agree. Could you try again, and if it fails, point me to the error?

beenje:
I changed the GIT_STRATEGY and it passed: https://gitlab.com/beenje/sardana/-/jobs/1643580111

reszelaz:
In 559d8f1 I addressed https://github.com/sardana-org/sardana/pull/1698#discussion_r720228878.

Tomorrow morning we will merge this PR and proceed with the migration. Thanks!

PR #1699, opened 2021-10-07
runMacro executed twice does not work anymore

jkotan:
At DESY we execute runMacro() twice on the same object. 
However, after the change 

  https://github.com/sardana-org/sardana/pull/1476/files

it does not work anymore because you cannot delete an object from the macro_status dictionary which does not exist anymore.
Therefore if it is allowed to runMacro twice it would be good to check before deleting 'name' and  'macro_line' if they exist in the dictionary.
Otherwise if running macro twice it is forbidden (I hope not) It should be thrown an exception informing users that running macro twice is forbidden.

### DIFF
--- a/src/sardana/macroserver/macro.py
+++ b/src/sardana/macroserver/macro.py
@@ -2346,8 +2346,10 @@ class Macro(Logger):
         # Avoid repeating same information on subsequent events. If, in the
         # future, clients that connect in the middle of macro execution need
         # this information, just simply remove the lines below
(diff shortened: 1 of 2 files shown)


jkotan:
Hi @reszelaz,  do you know why the test fails? If I run the sardana-test in my docker locally everything works fine.

jkotan:
it looks like in the first run of the PR I encountered a random event/macroexecutor error not related to the PR.

